PR #2, opened 2022-04-13
ci: use a 'constraints' file to install packages for tests

ahal:
Previously the tasks in CI would install everything that exists in
requirements/test.txt, which included a lot of unnecessary packages.

By instead passing in the file as a constraint via the `-c` option, we
only install the packages that are actually needed, while still
constraining them to the versions specified in the requirements.txt.

### DIFF
--- a/taskcluster/ci/tests/kind.yml
+++ b/taskcluster/ci/tests/kind.yml
@@ -38,8 +38,7 @@ jobs:
             symbol: unit
         run:
             command: >-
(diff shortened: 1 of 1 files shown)


ahal:
Hm, looks like you need to specify the `==`  on the command line which means we'd have our versions duplicated in two places.. so not something I think we should bother with.

Plus, even when I do specify `==` for some reason the dependency resolver takes *way* longer than simply installing everything. Just going to drop this idea.

PR #14, opened 2022-04-19
feat - bug 1764109 - allow for customized onExitStatus on generic-worker

escapewindow:


ahal (approved):
Thanks!

escapewindow:
Now that I merged, I looked back at https://docs.taskcluster.net/docs/reference/workers/generic-worker/multiuser-windows-payload and noticed there's no `purgeCaches` support =\

PR #16, opened 2022-04-20
Enable code coverage

ahal:
This will enable code coverage reporting via `pytest-cov` and codecov.io.


ahal:
Reviewing this one commit at a time might be best.

ahal:
This is currently set up to fail a check if the PR adds any new uncovered lines. I propose we let this land for now and see how annoying / useful it is and decide whether to tweak behaviour later. I will say that when I saw I added two new uncovered lines to the `fetch.py` transforms, it prompted me to go and write some tests for that file :p.. so I think it might be a good thing overall?

hneiva:
> ... so I think it might be a good thing overall?

+1 to that

ahal:
Oops, failed linting.. just a sec

ahal:
I wonder how codecov is able to compare against main given that this PR hasn't merged and we've never uploaded results from there.. Magic I guess.

ahal:
Also worth noting that coverga.py skips the files under `src/scripts`, I guess because they aren't proper modules... I'll see if there's a way to get them included in the report as a follow-up.

gabrielBusta (approved):
Failing a check if the PR adds new uncovered lines could be mildly annoying but I think it will have a positive effect on the overall test coverage.

ahal:
Yeah, I *think* Github will still display the green "Rebase and merge" button if it fails.. so it'll be easy to ignore the failure. But we can definitely adjust if it gets annoying. Thanks!

PR #23, opened 2022-04-27
ci: ensure 'run-task' scripts are included in coverage report

ahal:
The 'run-task' scripts weren't previously being picked up by the
coverage report as it isn't a proper Python module. But passing in the
path to this directory explicitly to 'pytest-cov' seems to do the trick.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,7 +11,7 @@ depends =
     report: py{37,38,39,310}
 commands =
     python --version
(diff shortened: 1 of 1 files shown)


ahal:
The reduction in coverage is expected, as this is just adding some more lowly covered files to the report.

jcristau (approved):
Makes sense, thanks!

PR #32, opened 2022-05-02
Py36 Unit Test Support

gabrielBusta:
Adds support for Py36 in the unit test. We can take out these tests as part of deprecating Py36. To install Py36 in a Debian container the `pyenv` tool builds Python 3.6 from source. I added additional dependencies to the docker image (build-essentials etc.) so the tool can build it.

To run the tests in Py36 we also have to pin sphinx (getting the latest results in a dependency conflict)


gabrielBusta:
The `codecov/patch` check is failing because of a lint error introduced from updating black. It'd take me a while to cover `stream_context_tar`

PR #34, opened 2022-05-03
fix(run-task): don't explicitly pass '--require-hashes'

ahal:
Currently when installing pip dependencies via the
`<REPO>_PIP_REQUIREMENTS` environment variable, `run-task` explicitly
passes in the `--require-hashes` option. This makes it difficult to test
against pre-release versions of Taskgraph since VCS urls do not support
hashes (https://github.com/pypa/pip/issues/6469).

Luckily when `pip` encounters a requirements with `--hash`, the
`--require-hashes` flag is implied. This means that if `run-task`
doesn't explicitly pass it in, we'll be able to test against pre-release
versions of taskgraph by adding the git URL and then simply *not* using
the `--generate-hashes` flag in `pip-compile`. But in production we'd
still get hash verification.

This does loosen restrictions a bit, since previously a repo would have
no choice other than to use hashes. But now they could opt not to.
Personally I feel like this is a trade-off that should be left to repo
owners anyway.

### DIFF
--- a/docs/howto/bootstrap-taskgraph.rst
+++ b/docs/howto/bootstrap-taskgraph.rst
@@ -142,9 +142,24 @@ accomplished using `pip's version control support`_:
 
    cd taskcluster
    echo "taskcluster-taskgraph@git+https://github.com/taskcluster/taskgraph@refs/pull/123/head" > requirements.in
-   pip-compile --generate-hashes --output-file requirements.txt requirements.in
+   pip-compile --output-file requirements.txt requirements.in
+
+Next edit your ``.taskcluster.yml`` to disable hashing since pip does not
+support `hashes with url requirements`_:
+
+.. code-block:: yaml
+
+   payload:
+      env:
+          - PIP_DISABLE_REQUIRE_HASHES: 1
+
+.. note::
+
+   Be sure to omit the ``--generate-hashes`` argument to ``pip-compile``
+   otherwise ``pip`` will implicitly turn hashing back on.
 
 This way you can push an experimental change to a PR and then install it in
 your repo's decision task.
 
 .. _pip's version control support: https://pip.pypa.io/en/stable/topics/vcs-support/
+.. _hashes with url requirements: https://github.com/pypa/pip/issues/6469
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -934,7 +934,10 @@ def install_pip_requirements(repositories):
     if not requirements:
         return
 
-    cmd = [sys.executable, '-mpip', 'install', '--require-hashes']
+    cmd = [sys.executable, '-mpip', 'install']
+    if os.environ.get("PIP_DISABLE_REQUIRE_HASHES") != "1":
+        cmd.append("--require-hashes")
+
     for path in requirements:
         cmd.extend(['-r', path])
 


ahal:
This will require new decision images to make use of.

escapewindow:
> This does loosen restrictions a bit, since previously a repo would have no choice other than to use hashes. But now they could opt not to. Personally I feel like this is a trade-off that should be left to repo owners anyway.

Is there a way we can require hashes on any production / release / level 3 branch, and not on level 2 or 1?

escapewindow:
> > This does loosen restrictions a bit, since previously a repo would have no choice other than to use hashes. But now they could opt not to. Personally I feel like this is a trade-off that should be left to repo owners anyway.
> 
> Is there a way we can require hashes on any production / release / level 3 branch, and not on level 2 or 1?

Doing so automatically might mean PRs go green and the merge breaks.
Maybe we need a flag in a PR like `NON_PRODUCTION_HACKS: True` that may be more `r-`-able.

ahal:
> > This does loosen restrictions a bit, since previously a repo would have no choice other than to use hashes. But now they could opt not to. Personally I feel like this is a trade-off that should be left to repo owners anyway.
> 
> Is there a way we can require hashes on any production / release / level 3 branch, and not on level 2 or 1?

Not really in a way that we can enforce. The `level` comes from the `.taskcluster.yml` so repo owners would need to pass it down to `run-task` via environment or something at which point they could fake it if they wanted to get around it.

> Doing so automatically might mean PRs go green and the merge breaks.
> Maybe we need a flag in a PR like NON_PRODUCTION_HACKS: True that may be more r--able.

I'm having trouble parsing this, you mean `r+` able? Would this go in the commit message or something?

There's a couple reasons I'm ok with not making hashes mandatory:
1. Prior to moving over to these requirements files we weren't doing any hash validation at all, so I think we're still in a better place relative to there.
2. Some less sensitive repos may actively choose not to care about validating hashes in exchange for not having to deal with `pip-tools`.. that seems reasonable to me as long as Releng enforces it at review time where it matters. For this point I'm mainly thinking about repos we have no involvement / stake in (e.g outside of Firefox-CI cluster).

escapewindow:
> > Doing so automatically might mean PRs go green and the merge breaks.
> > Maybe we need a flag in a PR like NON_PRODUCTION_HACKS: True that may be more r--able.
> 
> I'm having trouble parsing this, you mean `r+` able? Would this go in the commit message or something?

No, r-minus-able. Maybe a `TASKGRAPH_STRICT` or something env var we can check for in CoT, though that adds more best-practice enforcement where it's mainly supposed to be a security tool.

> There's a couple reasons I'm ok with not making hashes mandatory:
> 
>     1. Prior to moving over to these requirements files we weren't doing any hash validation at all, so I think we're still in a better place relative to there.
> 
>     2. Some less sensitive repos may actively choose not to care about validating hashes in exchange for not having to deal with `pip-tools`.. that seems reasonable to me as long as Releng enforces it at review time where it matters. For this point I'm mainly thinking about repos we have no involvement / stake in (e.g outside of Firefox-CI cluster).

This generally ends up with us finding at least one or two projects that we do have a stake in, that end up not following best practices. (It's easier to catch this at review time if there's at least one failing task.) This may or may not be an acceptable outcome, depending on perspective.

ahal:
> No, r-minus-able. Maybe a `TASKGRAPH_STRICT` or something env var we can check for in CoT, though that adds more best-practice enforcement where it's mainly supposed to be a security tool.

Oh ok, I see what you mean. We could have `--require-hashes` in by default, and only remove it if a `PIP_DISABLE_REQUIRE_HASHES` env is set. The downside is that we'd need to modify `.taskcluster.yml` anytime we want to test against a pre-release version.. but perhaps that's not so bad.

> This generally ends up with us finding at least one or two projects that we do have a stake in, that end up not following best practices. (It's easier to catch this at review time if there's at least one failing task.) This may or may not be an acceptable outcome, depending on perspective.

Yeah, that's likely.

ahal:
I'm slightly leaning towards the initial patch (being less strict about it), but it's not a sticking point.

escapewindow (approved):

PR #35, opened 2022-05-03
Fix issues specifying `--parameters` for projects using the Community instance

ahal:


ahal:
Hmm, not sure what that flake8 failure is about.. we should be skipping that file and I didn't modify it in this PR :/

matt-boris:
Nice, thanks!!

ahal:
@matt-boris if this is annoying you happy to cut a release for you.. otherwise just planning to let this roll out next time we need to do a release (seems we tend to do them weekly ish lately)

matt-boris:
@ahal no need to do a separate release! Thanks though.

PR #40, opened 2022-05-04
test: add a test for job/toolchain.py

ahal:
This is in preparation for supporting `generic-worker` in toolchains.

### DIFF
--- a/src/taskgraph/test/conftest.py
+++ b/src/taskgraph/test/conftest.py
@@ -1,4 +1,4 @@
-import os
+from pathlib import Path
 
 import pytest
 from responses import RequestsMock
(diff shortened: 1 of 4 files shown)


src/taskgraph/test/conftest.py
```diff
@@ -161,18 +162,20 @@ def make_fake_strategies():
 
 @pytest.fixture
 def transform_config(parameters):
-    graph_config = fake_load_graph_config(os.path.join("taskcluster", "ci"))
+    graph_config = fake_load_graph_config(str(here / "data" / "taskcluster" / "ci"))
```
escapewindow, line 165: Is this directory supposed to exist, or is it a bogus path?

escapewindow (commented):
Also unsure what the yamllint error is about?

src/taskgraph/test/conftest.py
ahal: It's a bogus path (for now at least). But it's needed so `GraphConfig.vcs_root` works:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/config.py#L110

I'm not very happy with that property.. but another problem for another time.

ahal:
The yamllint failure is fixed on main.. rebasing.

PR #41, opened 2022-05-05
chore(deps): upgrade index-task image

ahal:
This upgrades the index-task image:

node: 10 -> 16
taskcluster-client 12 -> 44

### DIFF
--- a/taskcluster/docker/index-task/Dockerfile
+++ b/taskcluster/docker/index-task/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10-alpine
+FROM node:16-alpine
 
 ENV       NODE_ENV        production
 RUN       mkdir /app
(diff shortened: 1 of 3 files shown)


ahal:
I'll need to figure out a good way of testing this.. maybe for now just adding a bunch of routes to an existing task..

ahal:
Ok, this appeared to work:
https://firefox-ci-tc.services.mozilla.com/tasks/ZQc3oLtTSGaTlfUyPsYQyg/runs/0

I also verified the task does appear in the index.

gabrielBusta (approved):
This looks good to me. Do the consumers have to do anything to use this newer image?

ahal:
> This looks good to me. Do the consumers have to do anything to use this newer image?

It looks like they should point to an index:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/morph.py#L81

So no, I don't think so. Though to be honest, that's probably not a good thing since it means if I make a mistake I could cause bustage across all repos as soon as this merges. We might want to tie this image to Taskgraph releases so that consumers need to update to pick it up.

So yeah, maybe I'll land this next week to avoid the risk of breaking everything on a Friday afternoon...

PR #44, opened 2022-05-05
Fix the git clean checkout regression

gabrielBusta:
This patch fixes the git clean checkout regression introduced in #22

I updated the unit test to account for this case, and I added a log warning where the remove function was failing silently.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -555,7 +555,7 @@ def _clean_git_checkout(destination_path):
         sys.exit(ret)
     data = stdout.read()
     prefix = 'Would remove '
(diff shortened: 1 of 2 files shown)


src/taskgraph/run-task/run-task
```diff
@@ -178,6 +186,7 @@ def remove(path):
         _call_windows_retry(os.chmod, (path, mode))
 
     if not os.path.lexists(path):
+        print_line(b'remove', b'WARNING: %s does not exists!\n' % path.encode('utf-8'))
```
gabrielBusta, line 189: Should we panic?

ahal (approved):
Nice find!

PR #48, opened 2022-05-20
Use pre-commit to handle linters and formatters

ahal:
We'll rely on https://pre-commit.ci to run the tasks. While using
Taskcluster for this is better ideologically, pre-commit.ci has some
compelling advantages:

1. The pre-commit cache is shared across all CI users so tasks will
almost never need to clone repos. Due to the low volume of commits to
taskgraph, if we used Taskcluster we'd rarely get a cache hit and tasks
would be *much* slower.
2. The pre-commit Github integration can automatically fix PRs.
3. The pre-commit Github integration can automatically submit PRs to
update dependencies.


ahal:
Neat.. Guess I either need to configure it to ignore `run-task` or else we start enabling checks there.

hneiva:
I always feel like 99 chars line length limit is a bit low :/

ahal:
> I always feel like 99 chars line length limit is a bit low :/

Interesting.. usually the feedback I get from people is that it's too high :p.

But either way, I'm purposefully trying to make this PR not change any of the linting rules. Should just be a straight migration of how the linters are run.

ahal:
> Neat.. Guess I either need to configure it to ignore `run-task` or else we start enabling checks there.

Since this comment doesn't make sense anymore, initially `pre-commit` wasn't configured to ignore the `run-task` dir and a bot added a commit with all the fixes there. I since got `pre-commit` to ignore this directory (like we were previously).

I do think we should turn on linting there, just wanted to keep that change out of this PR.

hneiva:
> > Neat.. Guess I either need to configure it to ignore `run-task` or else we start enabling checks there.
> 
> Since this comment doesn't make sense anymore, initially `pre-commit` wasn't configured to ignore the `run-task` dir and a bot added a commit with all the fixes there. I since got `pre-commit` to ignore this directory (like we were previously).
> 
> I do think we should turn on linting there, just wanted to keep that change out of this PR.

Yeah sounds like a good strategy

gabrielBusta (approved):

Issue #26, opened 2022-04-28
Drop support for Python 3.6


ahal:
This version is officially unsupported. The only blocker here is making sure that consumers are using the latest decision docker image as older versions still had 3.6 baked in. Luckily since consumers are pinning Taskgraph, this isn't a hard blocker.

jcristau:
Fixed by #207

PR #52, opened 2022-05-25
chore!: drop support for Python 3.6

ahal:
Issue: #26

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,6 @@ setup(
         "Development Status :: 5 - Production/Stable",
         "Environment :: Console",
         "License :: OSI Approved :: Mozilla Public License 2.0 (MPL 2.0)",
-        "Programming Language :: Python :: 3.6",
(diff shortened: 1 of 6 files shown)


gabrielBusta:
We could deprecate. I don't have to move the consumers to the new Taskgraph version right away. When we bump the Taskgraph version we can also bump the Taskgraph decision image

ahal:
Chatted with glandium, while building Firefox itself doesn't directly depend on Taskgraph, there are places where we import it at the top level. Which means we do risk breaking builds. This would only happen at vendor update time..  but still feels safer to me to just leave 3.6 alone for now. We could also refactor the build system to only import Taskgraph when it's used, but that doesn't feel worth the effort to me.

gabrielBusta:
can we do this when this bug is closed @ahal? https://bugzilla.mozilla.org/show_bug.cgi?id=1734402

ahal:
Superseded by #207

PR #54, opened 2022-05-27
Sync various changes from gecko_taskgraph that are reachable from `transforms.base`

ahal:
This is a pre-requisite to merging `transforms/base.py` between the taskgraphs.


ahal:
I think there was a blip in the coverage data caused by me migrating the tests in the previous commit, and now the coverage is returning back to the "real" level. In otherwords, this doesn't actually decrease coverage by 5%

escapewindow (approved):

PR #55, opened 2022-05-30
refactor!: rename taskgraph.util.WHITELISTED_SCHEMA_IDENTIFIERS

ahal:
In Gecko there's a lint check to prevent the use of
'whitelist/blacklist', and there's an exception for this case.

Since we're gearing up to release Taskgraph 2.0 anyway, might as well
make the switch.


gbrownmozilla (approved):
...or "allowed", if you prefer?

ahal:
I think for this case `ALLOWED` makes it sound like other keys aren't allowed? I don't know, excepted makes more sense in my mind.

PR #58, opened 2022-06-06
Draft: Task indexing

gabrielBusta:


src/taskgraph/morph.py
```diff
@@ -52,16 +52,17 @@ def amend_taskgraph(taskgraph, label_to_taskid, to_add):
 def derive_index_task(task, taskgraph, label_to_taskid, parameters, graph_config):
     """Create the shell of a task that depends on `task` and on the given docker
     image."""
+    task_index_config = graph_config["taskgraph"].get("task-index", {})
+    eager = task_index_config.get("eager", False)
```
ahal, line 56: We probably want this to be a per-task setting rather than global all or nothing setting.

src/taskgraph/morph.py
```diff
@@ -165,12 +166,22 @@ def add_index_tasks(taskgraph, label_to_taskid, parameters, graph_config):
...
     logger.debug("Morphing: adding index tasks")
 
+    task_index_config = graph_config["taskgraph"].get("task-index", {})
     added = []
     for label, task in taskgraph.tasks.items():
-        if len(task.task.get("routes", [])) <= MAX_ROUTES:
+        if (
+            not task_index_config.get("all", False)
+            and task.kind not in task_index_config.get("kinds", [])
+            and task.label not in task_index_config.get("labels", [])
```
ahal, line 175: I'm a little confused about the purpose of these filters.

src/taskgraph/morph.py
ahal: Note we have an `index_builder` mechanism in the `task.py` transforms already, e.g:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/transforms/task.py#L889

Maybe we can add this `eager` flag over there instead.

src/taskgraph/morph.py
```diff
@@ -165,12 +166,22 @@ def add_index_tasks(taskgraph, label_to_taskid, parameters, graph_config):
...
+            and len(task.task.get("routes", [])) <= MAX_ROUTES
```
ahal, line 176: This "index-task" is only needed if we go over the maximum routes allowed for a task. Most tasks don't go over this maximum. So the "eager" solution will need to work for the case we don't have an index task as well (or does it already work for that case?)

ahal (commented):
I know this is still draft, but thought I'd leave a few drive-by comments.

ahal:
After chatting with @gabrielBusta I don't believe this PR is necessary anymore.

PR #60, opened 2022-06-09
fix: add 'if-dependencies' key to job_description_schema

ahal:
This key should have been forwarded from the task_description_schema,
but was accidentally ommitted.


ahal:
Yeah this type of ommission happens all the time.. Probably we should just use voluptuous's `Extra` feature in the `job_description_schema` so that any unrecognized keys get passed through. Then we can leave it up to the `task_description_schema` to do the validation there.

PR #64, opened 2022-06-20
feat(run-task): set 'TASK_WORKDIR' environment variable

ahal:
It's often useful to know what the working directory of a task is in
order to find things like the artifacts or fetches dir.

Using the cwd doesn't work if either `--task-cwd` is specified or some
earlier step in a command changed to a new directory. Storing this
information in the env can help simplify things.


hneiva (approved):
lgtm

PR #73, opened 2022-07-11
fix: use 'zipfile' module to extract fetches on Windows

ahal:
Zipfiles generated natively on Windows use backslash path separators for
their internal path structure. The `unzip` utility in Linux is unable to
interpret this and creates files with `\` in their name rather than
directories.

Even if a task is running on Windows, the `unzip` utility is bundled
with `mozilla-build` and runs in the shell there. So it is still unable
to extract the archive properly.

This patch switches to extracting with the `zipfile` stdlib rather than
the `unzip` utility when running from Windows.

Note that the bug can still exist if a task running on Linux attempts to
consume a zipfile created from a task on Windows.

### DIFF
--- a/src/taskgraph/run-task/fetch-content
+++ b/src/taskgraph/run-task/fetch-content
@@ -329,15 +329,22 @@ def extract_archive(path, dest_dir, typ):
         if sys.platform == 'win32':
             tar = tarfile.open(fileobj=ifh, mode='r|')
             tar.extractall(str(dest_dir))
-            args = []
+            args = None
         else:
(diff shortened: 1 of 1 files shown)


ahal:
I found:
https://github.com/PowerShell/Microsoft.PowerShell.Archive/issues/48

So this issue was actually fixed in PowerShell, but I guess the VPN workers have a very old version of it installed. For this reason, I don't think it's worth solving in Taskgraph. Instead I'll see `extract: false` in the VPN repo and the task will handle extraction itself.

PR #77, opened 2022-07-13
docs: add howto page for using docker images

ahal:
Jira: RELENG-709

### DIFF
--- a/docs/howto/docker.rst
+++ b/docs/howto/docker.rst
@@ -1,0 +1,251 @@
+Use Docker Images
+=================
+
+Containerization is a key component of any CI system. Taskcluster provides a
+built-in worker implementation called `docker-worker`_ which makes running
+tasks with docker easy!
+
+.. note::
+
+   Taskcluster's generic-worker implementation can also run docker images and
+   will eventually replace the older docker-worker. But for now docker-worker
+   provides a better experience with docker.
+
+The `docker-worker payload`_ supports specifying images as a string (to be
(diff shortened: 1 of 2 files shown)


docs/howto/docker.rst
```diff
@@ -0,0 +1,237 @@
...
+
+.. code-block:: yaml
+
+   my-task:
+     worker:
+        docker-image: "ubuntu:latest"
+
+This method is easy and convenient, but has a few drawbacks. Namely if you need
+a custom image, you'll have to maintain a separate release process for it that
+involves versioning and pushing to DockerHub.
```
escapewindow, line 35: Best practice, which should be mandated for release graphs, is pinning the hash digest.
Chain of Trust also tries to limit [which task types can use prebuilt docker images](https://github.com/mozilla-releng/scriptworker/blob/a79eb8d6da984ef3254be99349132f55d5d676fb/src/scriptworker/constants.py#L509-L524) since we have limited/no auditing capability for prebuilt images (as opposed to when we can review the Dockerfile).

Not sure if we want to link to https://docs.mozilla-releng.net/en/latest/taskcluster/uploading_an_image.html or similar?

escapewindow (approved):

docs/howto/docker.rst
ahal: Good call, I'll add a note to this effect.

I don't really want to link back to the releng specific docs as that information won't be useful to anyone but us. Trying to make these docs target a more "generic taskgraph" audience.

PR #56, opened 2022-05-30
refactor!: rename 'job*' keys in transform loader to 'task*'

ahal:
This is a straight rename from job -> task in keys in the kind.yml
files.

Issue: #25

PR #85, opened 2022-07-26
Bug 1781248 - Import release_notifications transforms from gecko_task…

JohanLorenzo:
…graph

Tested in https://github.com/mozilla-mobile/fenix/pull/26207 in staging releases. I left example emails there.


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Change Log
 
+## [2.0.0] - 2022-08-01
+
+### Changed
+- Breaking change: you now need a `version.txt` file at the root of your repository. This file only contains a version number which is used by taskgraph for release-related operations
```
jcristau, line 6: I would prefer if we didn't make this mandatory.  Not all repos want/need release-related operations.

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+
+### Added
+- `release_notifications` transforms. It enables a task to send e-mail notifications. It usually used in the context of releases when you want to inform a group of people about the completion of a phase of a release. Notifications can contain data about what project/version reached a given phase.
+- New parameters: `version`, `next_version`, `build_number`. As of now, they are used by `release_notification` to send informative e-mails
```
jcristau, line 10: Shouldn't this be left up to consumers to define what they want in their notifications?

version.txt
```diff
@@ -1 +1 @@
-1.7.1
+1.8.0
```
ahal, line 1: There's already backwards incompatible changes on `main` so the next bump needs to be 2.0.0.

I'd appreciate if we could get https://github.com/taskcluster/taskgraph/pull/56 into the release as well :).

I already have a patch that does the bump and adds a migration guide so probably best to just drop this commit from your PR

ahal: Er, unless this is still a ways from being ready and we release `2.0` first, in which case you can keep this around bumping to `2.1.0`

src/taskgraph/util/scriptworker.py
```diff
@@ -0,0 +1,46 @@
...
+    if partial_updates != "" and config.kind in (
+        "release-bouncer-sub",
+        "release-bouncer-check",
+        "release-update-verify-config",
+        "release-secondary-update-verify-config",
+        "release-balrog-submit-toplevel",
+        "release-secondary-balrog-submit-toplevel",
+    ):
+        partial_updates = json.loads(partial_updates)
+        release_config["partial_versions"] = ", ".join(
```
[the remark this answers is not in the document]
JohanLorenzo: Yeah, I was wondering if we should remove that too. Done!

CHANGELOG.md
JohanLorenzo: Discussed offline. Breaking change removed. We don't want to force all projects to define a `version.txt` file. Some projects are meant to be continuous delivered, making version numbers useless.

CHANGELOG.md
JohanLorenzo: That's some bad wording of mine. Thanks for spotting that! I changed the sentence to be less ambiguous. Let me know if that works better now.

version.txt
JohanLorenzo: Oh right! Thanks for reminding me that! I updated the changelog with your recent changes. Let's see which PR lands last to cut the 2.0.0 release 🙂

ahal: Oh I meant I have a local stack that does it which I haven't submitted to a PR yet, so was planning to just do it as a standalone PR once ready :p, but we can do it as part of this too :)

src/taskgraph/transforms/release_notifications.py
```diff
@@ -6,11 +6,36 @@
...
+                    "on-exception",
+                    "on-failed",
+                    "on-pending",
+                    "on-resolved",
+                    "on-running",
+                )
+            ],
+        },
+    },
+    extra=ALLOW_EXTRA,
```
ahal, line 33: Oh neat, I didn't know about this. I always used the `Extra: object` trick (not sure if there's any difference between the two?)

But yeah, big +1 on only validating the keys that are actually used in the transform!

ahal (approved):
Thanks, looks great!

PR #91, opened 2022-08-09
feat(morph): ability for consumers to register custom morphs

ahal:
This will allow Gecko to use `taskgraph.morph.morph` while still registering its own custom morphs.


gbrownmozilla (approved):

JohanLorenzo:
Just a shout out: this new features unlocks [bug 1642977 comment 3](https://bugzilla.mozilla.org/show_bug.cgi?id=1642977#c3). Thanks for making that possible @ahal!

PR #93, opened 2022-08-11
Bug 1784232 -  Add `base_rev` parameter in taskgraph

JohanLorenzo:


JohanLorenzo:
Most of the testing was done through unit tests and this repo. I'd love to get this tested on hg before the actual merge. That said, I think it's ready for a first round of review, @ahal 🙂

src/taskgraph/util/vcs.py
```diff
@@ -97,6 +101,16 @@ def branch(self):
 
         return None
 
+    @property
+    def remote_name(self):
+        remotes = self.run("paths", "--quiet").splitlines()
+        if len(remotes) == 1:
+            return remotes[0]
+
+        raise RuntimeError(
```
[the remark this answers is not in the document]
ahal: If this is only used in a Decision task, or we're careful to handle the exception, I could be fine with it as a workaround. Though I think we should just select `default` if there are more than one and it exists (can still raise if it doesn't exist though).

src/taskgraph/util/vcs.py
```diff
@@ -139,6 +153,27 @@ def base_rev(self):
...
+            return remote_branch_name.split("/")[0]
+        except subprocess.CalledProcessError as e:
+            # Error code 128 comes with the message:
+            # "fatal: no upstream configured for branch $BRANCH"
+            if e.returncode != 128:
+                raise
+
+        remotes = self.run("remote").splitlines()
+        if len(remotes) == 1:
+            return remotes[0]
```
ahal, line 215: Similarly I wonder if we should just use `origin` if it exists?

src/taskgraph/decision.py
```diff
@@ -166,6 +167,9 @@ def get_decision_parameters(graph_config, options):
     except UnicodeDecodeError:
         commit_message = ""
 
+    if not options.get("base_ref"):
+        parameters["base_ref"] = repo.default_branch
```
ahal, line 172: Why is `default_branch` needed if we have `base_rev`? Wouldn't the latter be a better default anyway?

src/taskgraph/util/vcs.py
```diff
@@ -174,6 +187,58 @@ def remote_name(self):
...
+
+        try:
+            # This call works if you have (network) access to the repo
+            return self._get_default_branch_from_remote_query()
+        except (subprocess.CalledProcessError, RuntimeError):
+            pass
+
+        # this one is the last resort in case the remote is not accessible and
+        # the local repo is where `git init` was made
+        return self._guess_default_branch()
```
ahal, line 238: Nice :)

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ### Added
 - `vcs.remote_name` that tracks the name of the remote repository (e.g.: `default` on `hg` or `origin` on `git`)
 - `base_ref` parameter that points to the reference (e.g.: a branch) on the base repository
+- `vcs.find_latest_common_revision()` which find the common ancestor between a provided `base_ref` and the current `head_rev`
```
ahal, line 15: nit: finds

src/taskgraph/util/vcs.py
```diff
@@ -143,6 +148,15 @@ def working_directory_clean(self, untracked=False, ignored=False):
     def update(self, ref):
         return self.run("update", "--check", ref)
 
+    def find_latest_common_revision(self, base_ref):
+        return self.run(
+            "log",
+            "-r",
+            f"last(ancestors('{base_ref}') and ancestors('{self.head_rev}'))",
```
ahal, line 162: I'm confused, isn't `base_ref` always an ancestor of `head_rev`? In what cases would this not simply return `base_ref`?

src/taskgraph/util/vcs.py
```diff
@@ -157,6 +157,16 @@ def find_latest_common_revision(self, base_ref):
             "{node}",
         ).strip()
 
+    def does_revision_exist_locally(self, revision):
```
ahal, line 167: Please add this to the abstract base class with docstring

src/taskgraph/util/vcs.py
```diff
@@ -157,6 +157,16 @@ def find_latest_common_revision(self, base_ref):
...
+        try:
+            return self.run("log", "-r", f"ancestors(.) and '{revision}'").strip() != ""
```
ahal, line 169: So looks like on `hg` this only returns True if it exists AND is an ancestor of the current revision. But on Git that's not the case? Why the discrepency? Let's make sure to document the differences here.

src/taskgraph/util/vcs.py
```diff
@@ -16,6 +16,13 @@
 
 PUSHLOG_TMPL = "{}/json-pushes?version=2&changeset={}&tipsonly=1&full=1"
 
+# Both mercurial and git use sha1 as revision idenfiers. Luckily, both define
+# the same value as the null revision.
+#
+# https://github.com/git/git/blob/dc04167d378fb29d30e1647ff6ff51dd182bc9a3/t/oid-info/hash-info#L7
+# https://www.mercurial-scm.org/repo/hg-stable/file/82efc31bd152/mercurial/node.py#l30
+NULL_REVISION = "0000000000000000000000000000000000000000"
```
ahal, line 24: Would this make sense as an attribute on the Repository class?

ahal (changes requested):
Thanks, this looks pretty cool!

src/taskgraph/util/vcs.py
JohanLorenzo: I haven't tried the `hg` case yet. Maybe it's much simpler to deal with than git. I left an example of such a corner case in this comment.

src/taskgraph/util/vcs.py
JohanLorenzo: Nice catch! Done.

src/taskgraph/util/vcs.py
JohanLorenzo: Oh thank you very much for spotting this 👍 I just fixed the discrepancy by making this function compliant to the git one.

src/taskgraph/util/vcs.py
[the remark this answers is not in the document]
JohanLorenzo: Good thinking! It now returns `default` if it exists 👍

src/taskgraph/util/vcs.py
JohanLorenzo: Added.

JohanLorenzo (commented):
Thank you very much for the thorough review, @ahal! I just addressed your comments. I'll try this out end-to-end on git then hg. I don't think it's ready to merge just yet. I'll get back to this PR.

src/taskgraph/decision.py
ahal: Thanks for the explanation!

I'm not sure if this belongs in documentation, but I might be inclined to include it in the commit message itself!

JohanLorenzo: Sounds good! I just reformated my previous comment and put it in https://github.com/taskcluster/taskgraph/pull/93/commits/877c1b8ac499196badd129ae43d79c15aab2c9fd

src/taskgraph/run-task/run-task
```diff
@@ -606,6 +606,7 @@ def git_checkout(
         'git',
         'checkout',
         '-f',
+        '-B', ref,
```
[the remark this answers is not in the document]
JohanLorenzo: Good catch, fixed!

src/taskgraph/decision.py
JohanLorenzo: End-to-end testing showed there were a few things wrong in my previous explanation: we can't always rely on the `base_ref`. 

## Same head branch name, but force-pushed commits

Let's imagine we force-push several commits commit. The `base_ref` remains the same as the `head_ref` one. If we rely on the `base_ref`, then the `merge-base` between `base_ref` and the `head_rev` yields a single commit, the `head_rev`. This is a problem since we force pushed several commits.


## New head branch, but same code

This case needs to be split into 2. If the new branch is 100% the same as the old one, then `base_ref` is still the right way to go. 
If the head branch is slightly different (say a commit was amended), Github doesn't provides a wrong `base_ref` and gives `NULL_REVISION` as the `base_rev`. In this case, we're overriding the `base_ref` to match the default branch. It a saner default than keeping the provided `base_ref` which behaves as a single commit was pushed (like in the force-push` scenario.

## Changes

That's why I changed `run-task` to fetch the `base_ref` and the `base_rev` if it's the `NULL_REVISION`, if they're provided. The heuristic to determine more accurate `base_ref`/`base_rev` we changed too.

JohanLorenzo:
PR has been tested in many cases for git and hg. It's ready for review and we should be close to releasing 3.0.0 🙂

src/taskgraph/decision.py
```diff
@@ -236,6 +244,26 @@ def get_decision_parameters(graph_config, options):
...
 
+def _determine_more_accurate_base_ref(repo, candidate_base_ref, head_ref, base_rev):
+    base_ref = candidate_base_ref
+
+    if not candidate_base_ref:
+        base_ref = repo.default_branch
+    elif candidate_base_ref == head_ref and base_rev == Repository.NULL_REVISION:
+        logger.warning(
+            "base_ref and head_ref are identical but base_rev equals the null revision. "
+            "This is a new branch but Github didn't identify its actual base."
```
ahal, line 263: aside: I wonder if we should stop assuming Github in Taskgraph :). This is fine for now though, there are plenty bigger Github assumptions in Taskgraph.

.taskcluster.yml
```diff
@@ -49,6 +49,23 @@ tasks:
                   else:
                       $if: 'tasks_for in ["cron", "action"]'
                       then: '${repository.project}'
+          base_branch:
```
ahal, line 52: Shouldn't this also be called `base_ref` as it's not necessarily going to be a branch?

CHANGELOG.md
ahal: Looks like you may have lost the fix somewhere :)

src/taskgraph/decision.py
```diff
@@ -264,6 +272,38 @@ def _determine_more_accurate_base_ref(repo, candidate_base_ref, head_ref, base_r
     return base_ref
 
 
+def _determine_more_accurate_base_rev(repo, base_ref, candidate_base_rev, head_rev):
+    base_ref_or_rev = candidate_base_rev
```
ahal, line 276: nit: I think the logic would be a bit easier to follow if this goes into an `else:` clause. I was wondering why you do this assignment up here, only to compare using `candidate_base_rev` below, but then realized this was acting as the else condition.

src/taskgraph/decision.py
```diff
@@ -264,6 +272,38 @@ def _determine_more_accurate_base_ref(repo, candidate_base_ref, head_ref, base_r
...
+    if not candidate_base_rev:
+        logger.warning("base_rev is not set.")
+        base_ref_or_rev = base_ref
+    elif candidate_base_rev == Repository.NULL_REVISION:
+        logger.warning("base_rev equals the null revision. This branch is a new one.")
```
ahal, line 281: I'm a bit worried that logging all these at the warning level is going to lead people to believe something went wrong. These feel more like `debug` messages to me (i.e, we'd only really want to see them when something *has* gone wrong and we've enabled extra logging to determine what. Though I could probably be convinced to leave them at the info level.

src/taskgraph/decision.py
```diff
@@ -264,6 +272,38 @@ def _determine_more_accurate_base_ref(repo, candidate_base_ref, head_ref, base_r
...
+        base_ref_or_rev = base_ref
+    elif not repo.does_revision_exist_locally(candidate_base_rev):
+        logger.warning(
+            "base_rev does not exist locally. It is likely because the branch was force-pushed. "
+            "Please set the _BASE_REV environment variable to let run-task fetch it."
```
ahal, line 286: This message seems to imply an error, but looks like it will work just fine (which may cause some confusion). I also don't think the average developer will understand that they need to set `<REPO>_BASE_REV` in `.taskcluster.yml`. Can you make the instructions a bit more explicit?

.taskcluster.yml
```diff
@@ -75,6 +75,15 @@ tasks:
...
                       then: '${push.branch}'
+          base_sha:
+              $if: 'tasks_for == "github-push"'
+              then: '${event.before}'
+              else:
+                  $if: 'tasks_for == "github-pull-request"'
+                  then: '${event.pull_request.base.sha}'
+                  else:
+                      $if: 'tasks_for in ["cron", "action"]'
+                      then: '${push.revision}'
```
ahal, line 86: aside: we really need to get these definitions into some shared place 😅

src/taskgraph/run-task/run-task
```diff
@@ -612,6 +620,22 @@ def git_checkout(
 
         retry_required_command(b'vcs', args, cwd=destination_path, extra_env=env)
 
+    # When commits are force-pushed (like on a testing branch), base_rev doesn't
+    # exist on base_ref. Fetching it allow taskgraph to compute differences
```
ahal, line 624: nit: s/allow/allows

ahal (changes requested):
Thanks, looks great! I think all my comments are fairly trivial at this point, but I'll request changes in the name of being explicit around expectations :)

CHANGELOG.md
JohanLorenzo: I think there was some confusion because I didn't click on the "Resolve conversation" button. I confirm I made the fix which is visible in this comment. Closing conversation to prevent further confusion 🙂 

[1] https://github.com/taskcluster/taskgraph/pull/93/commits/d46b95744be12516be4c15678083cb23994f99d8#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR15

.taskcluster.yml
JohanLorenzo: Good call, fixed! I also renamed `head_branch` into `head_ref`.

src/taskgraph/decision.py
JohanLorenzo: Sounds good. `else` condition added!

.taskcluster.yml
JohanLorenzo: Yeah, I totally agree! All along that PR, I was like: "we need a new schema for `.taskcluster.yml` so that we stop having that much boilerplate". I'm having some ideas!

src/taskgraph/run-task/run-task
JohanLorenzo: Fixed!

src/taskgraph/decision.py
JohanLorenzo: You're right, maybe I was too easygoing with warning logs here. That said, I believe they have some value being output as INFO.

Let's say a developer doesn't get the expected tasks in a graph: either they or we have a way to immediately look into the logs and see what happened in this case. That doesn't require anyone to try to reproduce locally. Especially given the fact that the generated `parameters.yml` won't let you reproduce the bug since it contains the new `base_rev`/`base_ref` values. 

In addition to that, neither the developer nor us will likely remember to `git fetch` the `base_rev` Github provides. This means, the behavior will be different by default, locally. 

Finally, pushing things (to turn on debug logs, for instance) may lead to a different state which makes the bug not reproducible remotely either. 

Long story short: there are many footguns which will take time to figure out. Let's save ourselves some time by always being explicit about these heuristics.

What do you think?

src/taskgraph/decision.py
JohanLorenzo: Great catch! I just changed the message. Let me know if looks clearer 🙂

JohanLorenzo:
Minor changes made! I'm ready to get taskcluster 3.0.0 released if that looks good to you @ahal 🙂

src/taskgraph/decision.py
ahal: My only counter arguments are:
1. We log all the detailed optimization stuff at the DEBUG level already, so people kind of need to use DEBUG anyway to figure out why their task isn't running (though these logs are massive so there's good reason to hide them by default).
2. Gecko decision tasks use DEBUG verbosity by default, other repos could too.

But yeah, logging at INFO sounds fine to me.. no compelling reason not to.

ahal (approved):
Thanks, 🚢it!

src/taskgraph/decision.py
JohanLorenzo: I agree, we should switch to DEBUG level anyway to get into the optimization stuff. That said, that part is stateless unlike the `base_rev`/`base_ref` resets. I'd better be safe than sorry in this case. I really how confused people (myself included) can be because of the state of the repo.

PR #94, opened 2022-08-12
Fixes #81 - Support Git repositories in `files_changed.py` and `skip-unless-changed` optimization strategy 

### LINKED ISSUE #81: Support Git repositories in `files_changed.py` and `skip-unless-changed` optimization strategy
This strategy raises a RuntimeError if used on a Git repo:
https://github.com/taskcluster/taskgraph/blob/b4081f459831cc24fafb9687d8fca77c260f8c6b/src/taskgraph/optimize.py#L455

This is because `files_changed.py` queries `hg.m.o` to retrieve the files. Though there's no reason it couldn't query e.g [Github](https://docs.github.com/en/rest/pulls/pulls#list-pull-requests-files) for this information as well. Though I'm unsure whether there's an API for pushes...

Alternatively maybe this information could be found locally by comparing the head REF to the origin branch and `git diff-tree` or something similar.

A bit of thought is needed here, but it should be do-able for most cases.

JohanLorenzo:
Depends on https://github.com/taskcluster/taskgraph/pull/93 and https://github.com/taskcluster/taskgraph/pull/96


JohanLorenzo:
I recommend reviewing the last 6 commits only. Previous ones are from https://github.com/taskcluster/taskgraph/pull/93

src/taskgraph/util/vcs.py
```diff
@@ -210,14 +216,19 @@ def _files_template(self, diff_filter):
...
         if rev is None:
+            if base_rev is not None:
+                raise ValueError("Cannot specify `base_rev` without `rev`")
             # Use --no-status to print just the filename.
             df = self._format_diff_filter(diff_filter, for_status=True)
             return self.run("status", "--no-status", f"-{df}").splitlines()
         else:
             template = self._files_template(diff_filter)
-            return self.run("log", "-r", rev, "-T", template).splitlines()
+            revision_argument = rev if base_rev is None else f"{base_rev}~-1::{rev}"
```
ahal, line 230: Huh, TIL you can do `~-1`. I think just `~1` is equivalanet fwiw.

src/taskgraph/files_changed.py
```diff
@@ -8,23 +8,45 @@
...
+
+    if repository.tool == "hg":
+        # TODO Use VCS version once tested enough
+        return _get_changed_files_json_automationrelevance(
+            head_repository_url, head_rev
+        )
+    elif repository.tool == "git":
+        return repository.get_changed_files(rev=head_rev, base_rev=base_rev)
+
+    raise ValueError(f"Unknown repo tool: {repository.tool}")
```
ahal, line 40: Might as well change the `elif` to and `else` and get rid of this `raise ValueError`. I believe `get_repository` will already raise if the repo is not either `hg` or `git`.

ahal (approved):
Looks great, thanks!

src/taskgraph/files_changed.py
JohanLorenzo: 👌 done!

src/taskgraph/util/vcs.py
JohanLorenzo: Same here, TIL when I stumbled upon that problem 🙂 They're not totally equivalent. Here's the relation between between all 3:

```
tip
...
base_rev~-1
base_rev
base_rev~1
...
older_revisions
```

I confess I'm not 100% sure `~-1` is the right way to go. It works in the unit tests but I wonder how it behaves when a `base_rev` has more than a single child revision - which is quite common on `try`. 

The good news is: this code won't impact our `hg` repos yet. `hg` still uses `json-automationrelevance` while `git` gets tested in the real world. If that looks good enough, I'll perform another round of tests on hg before stopping to rely on `json-automationrelevance`

PR #96, opened 2022-08-16
Bug 1783690 - Let `run-task` clone all git-submodules

JohanLorenzo:


src/taskgraph/run-task/run-task
```diff
@@ -611,6 +611,23 @@ def git_checkout(
 
     run_required_command(b'vcs', args, cwd=destination_path)
 
+    if os.path.exists(os.path.join(destination_path, '.gitmodules')):
```
JohanLorenzo, line 614: While running these commands has no effect if there are no submodules defined, there is no need to run them and more importantly, log them out. This may cause confusion if someone sees `git submodule init` but there's no submodules at all.

For the record, `.gitmodules` has no other purpose than tracking submobules[1].

[1] https://www.git-scm.com/docs/gitmodules

gbrownmozilla (approved):

PR #100, opened 2022-08-18
actions: fix create_tasks after #82

### LINKED ISSUE #99: taskgraph.actions.util.create_tasks broken in 2.0.0
As far as I can tell https://github.com/taskcluster/taskgraph/commit/df9eb09e1c0cfed9d14eee17c1d71cc777b9efff added a new param to optimize_task_graph but one of the callers (https://github.com/taskcluster/taskgraph/blob/2.0.0/src/taskgraph/actions/util.py#L161) didn't get updated to pass it.  @rvandermeulen reported this breaking the add-new-jobs action in fenix with `TypeError: optimize_task_graph() missing 1 required positional argument: 'decision_task_id'`

jcristau:
Update another optimize_task_graph caller for the new function signature.

Fixes #99

### DIFF
--- a/src/taskgraph/actions/util.py
+++ b/src/taskgraph/actions/util.py
@@ -160,6 +160,7 @@ def create_tasks(
         target_task_graph.for_each_task(update_dependencies)
     optimized_task_graph, label_to_taskid = optimize_task_graph(
         target_task_graph,
+        to_run,
         params,
         to_run,
         decision_task_id,


JohanLorenzo:
I confirm this is the right fix! I tested it against staging fenix. Before this patch, the `add-new-jobs` action task failed[1]. Later on, it turned green[2]. Let's get this patch in taskgraph 3.0.0 👍 

[1] https://firefox-ci-tc.services.mozilla.com/tasks/MvGpE3VKQemhhqqfTwEAWg/runs/0/logs/public/logs/live.log#L2914
[2] https://firefox-ci-tc.services.mozilla.com/tasks/VYHOCLAPSR68F3sVMZSO3Q/runs/0/logs/live/public/logs/live.log

JohanLorenzo (approved):

ahal:
I guess normally stuff like this should go out in a patch release, but I kind of messed things up by landing a breaking change on main and not worth creating a branch imo.. So oh well.

JohanLorenzo:
Yeah, I agree. We still have a way to cut a branch for a dot release but I don't think it's worth the effort. I actually handled the migration to 3.0 on Fenix during the test and it was fairly straightforward[1].

[1] https://github.com/mozilla-releng/staging-fenix/commit/c43141920d137eaba03806b5e0758beb5b14dbc8

PR #102, opened 2022-08-23
Bug 1785439 - enforce milliseconds value in timestamp even if value i…

jcristau:
…s 0 to match parser expectation in graph builder.

Bug 1690947 started to parse timestamp strings provided for task expiration and
deadline which lacked the milliseconds part if the microseconds including
milliseconds had a value of 0 (because Python's `isoformat()` method got used
without enforcing the precision).

This caused the graph builder in the Gecko decision task to fail.

[jcristau: porting the fix from gecko_taskgraph to standalone]


ahal (approved):
Thanks!

ahal:
Looks like you missed updating a test expectation

PR #104, opened 2022-08-25
Bug 1786535 - Trim ".git" from tc-treeherder route's project

gbrownmozilla:
firefox-android is a private github repo accessed via ssh, so the base repository looks like "git@github.com:mozilla-mobile/firefox-android.git". Existing code for constructing the tc-treeherder route's base_project on pull request extracts "firefox-android.git" from the base repo, resulting in a route like "queue:route:tc-treeherder.v2.firefox-android.git-pr..." instead of the desired "queue:route:tc-treeherder.v2.firefox-android-pr..."


ahal (approved):
Thanks for adding the scaffolding for testing the task transforms! Probably the most important ones, so wasn't great they were untested.

PR #107, opened 2022-09-01
Bug 1786780 - Support SSH repositories in action tasks

JohanLorenzo:
While the upcoming monorepo is still private, we need to support SSH repositories in action tasks. Here's the error[1] I hit when this kind of value was provided: `[redacted]_HEAD_REPOSITORY: "git@github.com:[redacted]/[redacted].git"`: 

```
[pip-install 2022-09-01T14:04:19.153Z] You should consider upgrading via the '/usr/bin/python3 -m pip install --upgrade pip' command.
[task 2022-09-01T14:04:19.253Z] executing ['bash', '-cx', 'cd /builds/worker/checkouts/vcs && ln -s /builds/worker/artifacts artifacts && ~/.local/bin/taskgraph action-callback\n']
[task 2022-09-01T14:04:19.256Z] + cd /builds/worker/checkouts/vcs
[task 2022-09-01T14:04:19.256Z] + ln -s /builds/worker/artifacts artifacts
[task 2022-09-01T14:04:19.257Z] + /builds/worker/.local/bin/taskgraph action-callback
[task 2022-09-01T14:04:19.794Z] Traceback (most recent call last):
[task 2022-09-01T14:04:19.794Z]   File "/builds/worker/.local/lib/python3.8/site-packages/taskgraph/main.py", line 649, in action_callback
[task 2022-09-01T14:04:19.794Z]     return trigger_action_callback(
[task 2022-09-01T14:04:19.794Z]   File "/builds/worker/.local/lib/python3.8/site-packages/taskgraph/actions/registry.py", line 337, in trigger_action_callback
[task 2022-09-01T14:04:19.794Z]     sanity_check_task_scope(callback, parameters, graph_config)
[task 2022-09-01T14:04:19.795Z]   File "/builds/worker/.local/lib/python3.8/site-packages/taskgraph/actions/registry.py", line 301, in sanity_check_task_scope
[task 2022-09-01T14:04:19.795Z]     raise Exception(
[task 2022-09-01T14:04:19.795Z] Exception: {} is not either https://hg.mozilla.org or https://github.com !
```

[1] https://firefox-ci-tc.services.mozilla.com/tasks/Qc9hV3fNTril6r1B4baooQ/runs/0/logs/public/logs/live.log

PR #109, opened 2022-09-08
Trim .git from project in taskgraph.parameters._get_defaults

gabrielBusta:
The untrimmed project name from this function is breaking `taskgraph test-action-callback`

### DIFF
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -14,6 +14,7 @@ from subprocess import CalledProcessError
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
+from mozilla_repo_urls import parse
 from voluptuous import ALLOW_EXTRA, Any, Optional, Required, Schema
 
 from taskgraph.util import yaml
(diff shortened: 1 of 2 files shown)


jcristau (approved):

src/taskgraph/parameters.py
```diff
@@ -80,6 +80,8 @@ def _get_defaults(repo_root=None):
     try:
         repo_url = repo.get_url()
         project = repo_url.rsplit("/", 1)[1]
+        if project.endswith(".git"):
+            project = project[:-4]
```
JohanLorenzo, line 84: I'm glad this problem comes at the same time as `mozilla-repo-urls` got out. https://github.com/taskcluster/taskgraph/pull/107 introduced it. Let's leverage it here. Once that PR is done, we can bump `mozilla-repo-urls` to 0.0.3 and use this piece of code:

```suggestion
        repo_url = repo.get_url()
        parsed_url = parse(repo_url)
        project = parsed_url.repo_name
```

What do you think, @gabrielBusta?

gabrielBusta: I think it's great :) I can bump it in this PR and get rid of this repetitive logic while Im at it

JohanLorenzo: Awesome! mozilla-repo-urls 0.0.3 has been published! 🚢 

https://pypi.org/project/mozilla-repo-urls/0.0.3/#files

ahal (changes requested):
Sounds like we're planning to use `mozilla-repo-urls` here, so requsting changes for that.

ahal (approved):

JohanLorenzo:
🚀 \o/

PR #110, opened 2022-09-09
Fix: Handle git-cinnabar use cases

JohanLorenzo:
Context in: https://bugzilla.mozilla.org/show_bug.cgi?id=1789814#c6


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## [3.1.1] - TBD
```
ahal, line 3: Since this release will include a `feat`, the next version should be `3.2.0`

src/taskgraph/util/vcs.py
```diff
@@ -172,15 +198,10 @@ def branch(self):
...
-        if len(remotes) == 1:
-            return remotes[0]
-
-        if "default" in remotes:
-            return "default"
-
-        raise RuntimeError(
-            f"Cannot determine remote repository name. Candidate remotes: {remotes}"
+        return self._get_most_suitable_remote(
+            ("paths", "--quiet"),
```
ahal, line 202: What about if instead of passing in a `remote_command`, each vcs has:
```
@property
def remotes():
```
?

ahal: Similarly I wonder if `default_remote_name` should A) be a list of remote names, and B) be an attribute on the class.

ahal (changes requested):
Thanks, looks good! Just a few issues / suggestions.

CHANGELOG.md
JohanLorenzo: You're right! Done.

src/taskgraph/util/vcs.py
JohanLorenzo: These are excellent points! I didn't like what I made with `remote_command`. I find your suggestion much more elegant! 💯 

Done

ahal (approved):
Thanks, looks great!

PR #113, opened 2022-09-20
ci: switch to the 'public_restricted' pull request policy

ahal:
This means pull requests opened by non-collaborators will assume a different role and have a different set of scopes associated to them.

For taskgraph, the only task we'd want to limit is the `codecov-upload` one to avoid leaking the codecov API token. So we add a `run-on-tasks-for` key here which will filter out tasks_for == "github-pull-request-untrusted".

### DIFF
--- a/taskcluster/ci/codecov/kind.yml
+++ b/taskcluster/ci/codecov/kind.yml
@@ -35,6 +35,7 @@ tasks:
                 - codecov-uploader
             tests:
                 - coverage.xml
+        run-on-tasks-for: ["github-push", "github-pull-request"]
         run:
             using: run-task
             cwd: '{checkout}'
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -7,7 +7,7 @@
 version: 1
 reporting: checks-v1
 policy:
-    pullRequests: collaborators
+    pullRequests: public_restricted
 tasks:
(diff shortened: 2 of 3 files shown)


taskcluster/ci/codecov/kind.yml
```diff
@@ -35,6 +35,7 @@ tasks:
                 - codecov-uploader
             tests:
                 - coverage.xml
+        run-on-tasks-for: ["github-push", "github-pull-request"]
```
escapewindow, line 38: Will this work for `github-pull-request:*` or do we need to `rework run-on-tasks-for`?

ahal: The `:*` construct is only internal to ci-configuration. Taskcluster Github will either set `tasks_for = github-pull-request` or `tasks_for = github-pull-request-untrusted`. So this will prevent the task from running on the latter without re-working anything.

escapewindow (approved):
I don't love that our terminology switches from .tc.yml to ci-config, but this otherwise lgtm.

ahal:
> I don't love that our terminology switches from .tc.yml to ci-config, but this otherwise lgtm.

Yeah, in hindsight I should have made Taskcluster Github specify roles like `pull-request:trusted` and `pull-request:untrusted` as well, but I didn't to avoid breaking backwards compatibility. It's not too late to change this at some point in the future either though.

PR #119, opened 2022-09-21
Catch InvalidRepoUrlError and fail gracefully.

kdashg:
This fixes `mach bootstrap`'s InvalidRepoUrlError issues with a remote like "git@example.com:gecko".

Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1791852

### DIFF
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -14,7 +14,7 @@ from subprocess import CalledProcessError
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
-from mozilla_repo_urls import parse
+import mozilla_repo_urls
 from voluptuous import ALLOW_EXTRA, Any, Optional, Required, Schema
 
 from taskgraph.util import yaml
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Thank you very much for detailing the context and for providing this patch, @kdashg! At a first glance, it looks good to me. I'd like to add unit tests to ensure we don't regress that in the future. I see Taskcluster didn't run because of the policy on this repo (cc @ahal). 

I was running the test suite when it actually enabled me to reproduce issues that other have had in the last comments of [bug 1784232](https://bugzilla.mozilla.org/show_bug.cgi?id=1784232). That is such an awesome side effect of your PR so I would like to thank you again for helping me out 😃 . For the record, I documented how I got to repro in [bug 1784232 comment 44](https://bugzilla.mozilla.org/show_bug.cgi?id=1784232#c44).

If that's okay with you, I'll first debug what I repro then get back in this PR to add tests. Let me know what you think!

ahal:
> I see Taskcluster didn't run because of the policy on this repo (cc @ahal).

Heh, this will be fixed when #113 lands (which is blocked on review for https://phabricator.services.mozilla.com/D157493)

JohanLorenzo:
> If that's okay with you, I'll first debug what I repro then get back in this PR to add tests. Let me know what you think!

Status update. I managed to get my local clone in a better shape which allowed me to write tests. I opened https://github.com/taskcluster/taskgraph/pull/120 to get CI to run. If all goes well, I'll merge it 🙂

JohanLorenzo (approved):
LGTM! All checks are green per https://github.com/taskcluster/taskgraph/runs/8495563748. Let's get this landed. Thanks again for your patch @kdashg! 👍

PR #121, opened 2022-09-22
Bug 1784232 - part 17 - Provide empty base_ref and base_rev in default parameters

JohanLorenzo:
Context in [bug 1784232 comment 45](https://bugzilla.mozilla.org/show_bug.cgi?id=1784232#c45). The decision task still gets its values set per:

https://github.com/taskcluster/taskgraph/blob/49d407c4dd2b2bb5c335c9e24f4179d675ce5bde/src/taskgraph/decision.py#L171-L184


jcristau:
Thanks for doing this, that sounds like it'll solve a bunch of headaches.

ahal (approved):
That was easier than I thought. Do we have to update any logic in e.g `files_changed.py` to account for the fact it may be an empty string? E.g woud `./mach taskgraph optimized` still work after this?

JohanLorenzo:
> Do we have to update any logic in e.g files_changed.py to account for the fact it may be an empty string? E.g woud ./mach taskgraph optimized still work after this?

Good call! I'll try it out once I bump taskgraph in mozilla-central. I'm cutting a release in the meantime and I will create followup fixes, if needed.

JohanLorenzo:
> E.g woud `./mach taskgraph optimized` still work after this?

I tested it, it worked. I haven't looked into the details but I see 2 reasons: 
 1. gecko_taskgraph has its `files_changes` module[1], and it seems to duplicate the logic of standalone taskgraph
 2. In standalone taskgraph,`files_changes` still relies on `json-automationrelevance` for hg repos. That said, I tested `./mach taskgraph optimized` on both hg and git-cinnabar.

The rest of the testing of this PR was detailed in [bug 1784232 comment 55](https://bugzilla.mozilla.org/show_bug.cgi?id=1784232#c55).

[1] https://searchfox.org/mozilla-central/rev/a26af613a476fafe6c3eba05a81bef63dff3c9f1/taskcluster/gecko_taskgraph/files_changed.py#24-50
[2] https://github.com/taskcluster/taskgraph/pull/94/files#diff-73c1933ae321dccc0aa5f0681562b21c0fe882ab632073519d9479d1f3e537aeR32

PR #125, opened 2022-09-29
feat(optimize): support overriding optimization strategies via parameter

ahal:
The ability to override optimization strategies in 'optimize_tasks' already exists, but generator.py doesn't yet make use of this feature. This feature is being used in Gecko to run tasks that generate local graphs using experimental optimization strategies so their effectiveness can be evaluated against one another.

Adding this feature to standalone is a pre-requisite to merging `generator.py`.

This syncs a change from gecko_taskgraph.


src/taskgraph/generator.py
```diff
@@ -383,13 +383,22 @@ def _run(self):
         do_not_optimize = set(parameters.get("do_not_optimize", []))
         if not parameters.get("optimize_target_tasks", True):
             do_not_optimize = set(target_task_set.graph.nodes).union(do_not_optimize)
+
+        # this is used for testing experimental optimization strategies
+        strategies = os.environ.get(
+            "TASKGRAPH_OPTIMIZE_STRATEGIES", parameters.get("optimize_strategies")
+        )
+        if strategies:
+            strategies = find_object(strategies)
```
hneiva, line 392: nit: maybe don't overwrite variable types like this? i.e: `stategies = find_object(raw_strategies)`

hneiva (approved):
is this just an uplift from gecko code then?

LGTM

ahal:
> is this just an uplift from gecko code then?
> 
> LGTM

Yeah. I'm not 100% happy with how it's implemented, but in the interest of not blocking the merge, I'd vote we just stick with what's implemented in Gecko for now and we can always improve / change it later.

PR #126, opened 2022-09-30
fix: Catch `UnsupportedPlatformError` in order not to bust mach commands (bug 1793219)

catamorphism:
When the remote URL is rejected as unsupported, it's helpful to print a list of the platforms that are supported and explain that the default remote should be changed to point to a repository on one of these platforms.

**Note:** this patch depends on https://github.com/mozilla-releng/mozilla-repo-urls/pull/6 and shouldn't be merged until that patch is merged.

### DIFF
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -87,7 +87,19 @@ def _get_defaults(repo_root=None):
     except (CalledProcessError, mozilla_repo_urls.errors.InvalidRepoUrlError):
         repo_url = ""
         project = ""
-
+    except (mozilla_repo_urls.errors.UnsupportedPlatformError) as err:
+        repo_remote = repo.remote_name
+        """
+        Combine the error message printed by `UnsupportedPlatformError` with a hint
+        about what to do to fix the error: set the expected remote, as determined by
+        the `Repository` class, to a URL whose domain is supported. Include the list of
+        supported domains in the error message.
+        """
+        raise RuntimeError(
+           f"{str(err)}\n\n  Hint: your default remote is `{repo_remote}`, which points to"
+           f"{repo_url}"
+           f"\n  This script requires the `{repo_remote}` remote to point to a repository"
+           f"on one of these domains: {err.supported_platforms}")
     return {
         "base_repository": repo_url,
         "base_ref": "",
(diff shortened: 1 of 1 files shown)


ahal:
Going to close and re-open to see if that gets a Decision task to run (as we recently changed pull request policy to support non-collaborators).

ahal:
Ah, might need to be based on latest `main`. Oh well, not a big deal.

src/taskgraph/parameters.py
```diff
@@ -86,7 +86,20 @@ def _get_defaults(repo_root=None):
     except (CalledProcessError, mozilla_repo_urls.errors.InvalidRepoUrlError):
```
JohanLorenzo, line 86:
```suggestion
    except (
        CalledProcessError,
        mozilla_repo_urls.errors.InvalidRepoUrlError,
        mozilla_repo_urls.errors.UnsupportedPlatformError,
    ):
```

src/taskgraph/parameters.py
```diff
@@ -86,7 +86,20 @@ def _get_defaults(repo_root=None):
...
+        about what to do to fix the error: set the expected remote, as determined by
+        the `Repository` class, to a URL whose domain is supported. Include the list of
+        supported domains in the error message.
+        """
+        raise RuntimeError(
+            f"{str(err)}\n\n  Hint: your default remote is `{repo_remote}`, which points to"
+            f"{repo_url}"
+            f"\n  This script requires the `{repo_remote}` remote to point to a repository"
+            f"on one of these domains: {err.supported_platforms}"
+        )
```
JohanLorenzo, line 102:
```suggestion

```

JohanLorenzo (commented):
Thank you for this patch! After tinkering in [bug 1784232](https://bugzilla.mozilla.org/show_bug.cgi?id=1784232), I don't think `taskgraph` should find the default remote by default anymore. That's why `repo_url` gets set to an empty string if `InvalidRepoUrlError` is raised.

I think we should do the same when `UnsupportedPlatformError` comes up. An error message is too disruptive even when greatly documented like the one you wrote 🙂 

Thus, I think we can just expand the `except` statement and we should be done with it. What do you think @catamorphism?

JohanLorenzo:
> Ah, might need to be based on latest main. Oh well, not a big deal.

PR rebased and taskcluster tasks did get run, @ahal 👍

JohanLorenzo (approved):

PR #127, opened 2022-10-10
Drop support for pinning hg.mozilla.org TLS certificate fingerprint

jcristau:
Gecko is dropping this in
https://bugzilla.mozilla.org/show_bug.cgi?id=1793923, and other taskgraph users shouldn't need it either.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -42,11 +42,6 @@ import urllib.request
 from threading import Thread
 
 SECRET_BASEURL_TPL = 'http://taskcluster/secrets/v1/secret/{}'
-FINGERPRINT_URL = SECRET_BASEURL_TPL.format('project/taskcluster/gecko/hgfingerprint')
-FALLBACK_FINGERPRINT = {
-    'fingerprints':
-        "sha256:17:38:aa:92:0b:84:3e:aa:8e:52:52:e9:4c:2f:98:a9:0e:bf:6c:3e:e9"
-        ":15:ff:0a:29:80:f7:06:02:5b:e8:48"}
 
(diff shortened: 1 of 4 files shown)


src/taskgraph/run-task/run-task
```diff
@@ -1101,8 +1048,7 @@ def main(args):
         add_vcs_arguments(parser, repository, name)
 
     parser.add_argument('--fetch-hgfingerprint', action='store_true',
-                        help='Fetch the latest hgfingerprint from the secrets store, '
-                        'using the taskclsuerProxy')
+                        help="Obsolete option, ignored for backwards compatibility")
```
ahal, line 1051: Might be better to set `help=argparse.SUPPRESS`. That way the option won't even show up in the help in the first place.

test/test_transforms_job_run_task.py
```diff
@@ -47,7 +47,7 @@ def assert_docker_worker(task):
         "extra": {},
         "label": "fake-task-label",
         "routes": [],
-        "scopes": ["secrets:get:project/taskcluster/gecko/hgfingerprint"],
+        "scopes": [],
```
ahal, line 50: I guess we can't remove the scope from ci-config until the Gecko patch rides the trains everywhere..

ahal (approved):
Nice, thanks for this!

test/test_transforms_job_run_task.py
jcristau: Right.  I'm planning to let it ride 108, and uplift to ESR 102.6 or 102.7.

src/taskgraph/run-task/run-task
jcristau: Thanks for the tip :)

PR #131, opened 2022-10-11
Update README.rst

ahal-test:


ahal:
Nice, looks like `public_restricted` policy is working. The `codecov-upload` task requires secrets so is configured to not run on non-collaborator PRs (like the one my dummy account here made).

Issue #29, opened 2022-04-28
Remove 'always-target' feature


ahal:
[Always target](https://github.com/taskcluster/taskgraph/blob/2b79c1d059e409d2de550551c82e33c27733d2dd/src/taskgraph/generator.py#L363) is a feature that was added to address a narrow use case in Gecko. It's not too clear that it is useful outside of Gecko as other repos tend not to have complicated optimization logic.

There is a slight discrepency in their implementations now too, on Gecko we only use the feature for `hg-push` graphs. As the original creator of this feature. So if we want to merge the two Taskgraph's we either need to:

1. Sync the Gecko change over here.
2. Remove `always-target` and refactor Gecko to perform the same logic in the `target-tasks` phase.

As the original author of this feature, I think this feature was implemented the wrong way, and option 2 is a better approach here. So that means we should drop support for `always-target` here.

ahal:
We decided to leave always-target after all, as there's a use case in Gecko that can't be solved without it.

PR #134, opened 2022-10-12
refactor!: drop support for `always_target` attribute

ahal:
BREAKING CHANGE

This is a feature that was cargo culted from Gecko, but likely isn't useful or needed outside of Gecko. We are moving it out of `generator.py` over there, and dropping the feature here.

If consumers do end up needing something like this, they can re-implement it similar to how Gecko does. We also look into re-introducing the functionality in core Taskgraph again if needed (with nicer implementation).

Issue: #29

### DIFF
--- a/src/taskgraph/generator.py
+++ b/src/taskgraph/generator.py
@@ -359,17 +359,7 @@ class TaskGraphGenerator:
             for t in full_task_graph.tasks.values()
             if t.attributes["kind"] == "docker-image"
         }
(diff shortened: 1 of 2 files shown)


ahal:
Due to https://bugzilla.mozilla.org/show_bug.cgi?id=1795921, I think we may need to keep `always-target` around after all.. I'm tweaking the implementation in Gecko a bit to make it more generic (and not rely on the `hg-push` tasks_for value).

PR #136, opened 2022-10-18
ci: only run `github-push` tasks on main

ahal:
Previously we were treating all non-main branches as level 3, which means anyone could run level 3 tasks (luckily Taskgraph doesn't really have any sensitive ones anyway).

By simply not running the tasks on non-main branch we also prevent duplicate tasks on PRs opened within the canonical repository (as opposed to a fork).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -108,7 +108,7 @@ tasks:
       in:
           $if: >
             tasks_for in ["action", "cron"]
-            || (tasks_for == "github-push" && head_ref[:10] != "refs/tags/")
+            || (tasks_for == "github-push" && head_branch == "refs/heads/main")
             || (isPullRequest && pullRequestAction in ["opened", "reopened", "synchronize"])
           then:
               $let:


JohanLorenzo (approved):
LGTM! For the record, outside of the taskgraph repo, we've used pushes to random branches on a non-fork repos. This has been useful in order to run tasks that should run on PRs - no matter if they come from a trusted contributor or not. 

Maybe, in a glorious future, we would want to stop doing that 🙂

jcristau:
> Previously we were treating all non-main branches as level 3, which means anyone could run level 3 tasks (luckily Taskgraph doesn't really have any sensitive ones anyway).

Anyone with write access to the repo, which isn't really "anyone", right?

ahal:
> > Previously we were treating all non-main branches as level 3, which means anyone could run level 3 tasks (luckily Taskgraph doesn't really have any sensitive ones anyway).
> 
> Anyone with write access to the repo, which isn't really "anyone", right?

True, I guess in this case it was the pre-commit.ci bot, which I guess we granted write access when it was installed.

ahal:
> LGTM! For the record, outside of the taskgraph repo, we've used pushes to random branches on a non-fork repos. This has been useful in order to run tasks that should run on PRs - no matter if they come from a trusted contributor or not.

This should still work on Taskgraph, you just need to open a PR! This is how the VPN repo is setup and it seems to work well for them. I think here it'll work well too because there aren't any tasks that don't normally run on PRs. We can re-evaluate if our tasks get more complex down the road.

escapewindow:
> LGTM! For the record, outside of the taskgraph repo, we've used pushes to random branches on a non-fork repos. This has been useful in order to run tasks that should run on PRs - no matter if they come from a trusted contributor or not.
> 
> Maybe, in a glorious future, we would want to stop doing that 🙂

We can do that, as long as those tasks are run by level 1 workers. For tc.yml templates, I'm picturing config items for the main branch and relbranch_prefix: if the branch is the main branch, or the branch starts with relbranch_prefix, then we're level 3. otherwise github-push defaults to level 1.

PR #139, opened 2022-10-24
Bug 1796139 - deduplicate fetch-content with gecko

jcristau:
This brings all changes to gecko's version of fetch-content from the last couple of years to upstream taskgraph.

In addition this updates the robustcheckout hg extension to the latest version, and reformats the scripts in the `run-task` directory with black.

Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1796139

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [tool.black]
 line-length = 88
 extend-exclude = """(\
-  src/taskgraph/run-task|\
   taskcluster/scripts/external_tools)\
   """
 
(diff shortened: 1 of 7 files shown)


ahal:
Nice, good call preserving history :)

ahal (approved):
Lgtm! Hard to tell if this will cause any issues for taskgraph consumer, and harder to test. But I think it should be pretty safe.

From what I can tell, there doesn't seem to be anything backwards incompatible here.

jcristau:
Need to update this to account for [bug 1797243](https://bugzilla.mozilla.org/show_bug.cgi?id=1797243)

ahal (approved):

PR #140, opened 2022-10-27
Support docker-worker capability disableSeccomp

petemoore:
This was added in https://github.com/taskcluster/taskcluster/pull/5695.

CC @jschwartzentruber @matt-boris @lotas


ahal (approved):
Thanks, lgtm!

PR #141, opened 2022-10-27
fix: sort taskgraph label output

ahal:
When using the formatter that just outputs the labels, tasks will appear all out of order due to how dependencies are added. When the graph is large, it's very difficult to find the task you're looking for. Sorting the output makes this much better.


JohanLorenzo (approved):
Good idea! 🚢 it!

PR #142, opened 2022-10-27
feat: Add ability to enable/disable the 'always_target' attribute via…

ahal:
… parameter

This syncs a change from Gecko.

It doesn't always make sense to use the `always_target` feature. E.g, in Gecko, we only want to enable it for try and nowhere else. This allows projects to disable it in certain instances. It defaults to `True` for now to preserve backwards compatibility.


src/taskgraph/generator.py
```diff
@@ -360,11 +360,14 @@ def _run(self):
             if t.attributes["kind"] == "docker-image"
         }
         # include all tasks with `always_target` set
-        always_target_tasks = {
-            t.label
-            for t in full_task_graph.tasks.values()
-            if t.attributes.get("always_target")
-        }
+        if parameters["enable_always_target"] is True:
```
jcristau, line 363: nit: remove `is True`

jcristau (approved):

PR #145, opened 2022-10-31
docs: add better documentation on dependencies

ahal:
After needing to explain how to write a dependency, I realized the docs on this topic were sorely lacking.

### DIFF
--- a/docs/concepts/task-graphs.rst
+++ b/docs/concepts/task-graphs.rst
@@ -118,33 +118,111 @@ Dependencies
 ------------
 
 Dependencies between tasks are represented as labeled edges in the task graph.
-For example, a test task must depend on the build task creating the artifact it
-tests, and this dependency edge is named 'build'. The task graph generation
-process later resolves these dependencies to specific task ids.
+They are specified via the ``dependencies`` key which is an object of the form
+``{ "<edge>": "<label>"}``. The ``edge`` is an arbitrary name which can be used
+to refer to the dependency later on. The ``label`` is the task label of the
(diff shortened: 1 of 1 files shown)


ahal:
Rendered version can be found here:
https://taskcluster-taskgraph--145.org.readthedocs.build/en/145/concepts/task-graphs.html#dependencies

docs/concepts/task-graphs.rst
```diff
@@ -118,33 +118,111 @@ Dependencies
 ------------
 
 Dependencies between tasks are represented as labeled edges in the task graph.
-For example, a test task must depend on the build task creating the artifact it
-tests, and this dependency edge is named 'build'. The task graph generation
-process later resolves these dependencies to specific task ids.
+They are specified via the ``dependencies`` key which is an object of the form
+``{ "<edge>": "<label>"}``. The ``edge`` is an arbitrary name which can be used
+to refer to the dependency later on. The ``label`` is the task label of the
+dependency.
```
escapewindow, line 124: I may be misremembering, but the value can also be a Task object, no?

escapewindow (approved):
Thanks! This definitely needed clarification.

docs/concepts/task-graphs.rst
ahal: I'm not sure.. The `task_description_schema` allows it, but I'm failing to find where we translate from `Task` object -> label (as the latter is definitely what's expected). Maybe it used to be supported but got removed at some point?

PR #147, opened 2022-11-02
fix: make zstandard an optional dependency

ahal:
It turns out we don't want to vendor zstandard into mozilla-central because it requires compiled binaries.

A) The vendor results in 3.5Mb of unused binaries being added to the Gecko source, and
B) zstandard is listed as a "pypi" dependency in the mach environment. Also adding it as a "vendored" dependency causes some issues that need to be worked around.

Overall it is simpler if Taskgraph only lists it as an optional dependency so we don't end up pulling it into m-c.


JohanLorenzo (approved):
LGTM! Agreed to make this dependency optional for the well-being of mozilla-central!

PR #149, opened 2022-11-07
Experiment generating kinds concurrently

ahal:
This appears to work, and in all my tests results in an identical taskgraph as before (including in m-c). Though the patch isn't actually using async/await in the transforms yet, so there isn't a noticeable perf improvement.


ahal:
Early exploration for solving issue #5

PR #150, opened 2022-11-08
Bug 1798930 - Support .tgz extension in fetch-content

jcristau:
Many projects use .tgz as a shortcut to .tar.gz. Support that pattern.


JohanLorenzo (approved):
Changes make sense to me!

PR #152, opened 2022-11-09
Support all Taskcluster notifications

ahal:
This adds support for all types of notifications supported by the Taskcluster notification service:
https://docs.taskcluster.net/docs/reference/core/notify/usage


src/taskgraph/transforms/__init__.py
```diff
@@ -0,0 +1,3 @@
+from taskgraph.transforms import (  # noqa: Added for backwards compat
```
JohanLorenzo, line 1: 👍

src/taskgraph/transforms/notify.py
```diff
@@ -0,0 +1,174 @@
...
+                    Optional("content"): str,
+                    Optional("link"): {
+                        Required("text"): str,
+                        Required("href"): str,
+                    },
+                },
+                Optional("matrixBody"): str,
+                Optional("matrixFormattedBody"): str,
+                Optional("matrixFormat"): str,
+                Optional("matrixMsgType"): str,
```
JohanLorenzo, line 72: Bikeshed: I wonder if we should put the matrix-related attributes under a single dictionary, like the `email` one above. E.g.: 

```suggestion
                Optional("matrix"): {
                    Required("body"): str,
                    Optional("formatted-body"): str,
                    Optional("format"): str,
                    Optional("msg-type"): str,
                },
```

src/taskgraph/transforms/notify.py
```diff
@@ -0,0 +1,174 @@
...
+                Optional("slackText"): str,
+                Optional("slackBlocks"): dict,
+                Optional("slackAttachments"): dict,
```
JohanLorenzo, line 75: Bikeshed: same rationale as `matrix`.

src/taskgraph/util/schema.py
```diff
@@ -157,6 +157,7 @@ def resolve_keyed_by(
     # which expects interCaps
     "upstream-artifacts",
     "artifact-map",
+    lambda item: "schema['notify']['context']" in item,
```
JohanLorenzo, line 160: Bikeshed: I wonder if we can avoid the camelCase here because the generated task should be just contain some routes.

JohanLorenzo (commented):
Looks good to me overall! I bikesheded on the schema 😁

src/taskgraph/transforms/notify.py
ahal: Yes, this is a valid bikeshed and something I considered as well. However I just want to point out that these keys are set up to *exactly* match the Taskcluster's schema as per:
https://docs.taskcluster.net/docs/reference/core/notify/usage#setting-custom-messages

So the benefit of this way is that:
A) Users can lookup these keys in the Taskcluster docs easier, and
B) Taskgraph can just forward them wholesale to Taskcluster without doing any translation

I do agree that the Taskcluster schema is inconsistent and strange though (e.g why does `email` have subkeys, but not `matrix` or `slack`?).

I would be happy to make these requested changes if there is consensus this way is better. Just wanted to point out why I implemented them this way.

JohanLorenzo: Thanks for the context! I looked into the Taskcluster code and it seems nobody has ever raised this concern. The Matrix support was added in early 2020 in https://github.com/taskcluster/taskcluster/pull/2390. Slack was in late 2020 in https://github.com/taskcluster/taskcluster/pull/3610. As a reference point, email support was added in 2016[1].

People who implemented all 3 schemas are not at Mozilla anymore. So we won't be able to get any additional context.

I think it's okay to change the schema in Taskgraph for the sake of consistency. As a matter of fact, Taskgraph users are used to comply to some schema in `kind.yml` and let the transforms deal with what is really uploaded to the queue. That said, I'm happy to hear different opinions. 

[1] https://github.com/taskcluster/taskcluster-notify/commits/c557e553a703769dbe94629139865a100e4f1ca7/schemas/email-request.yml

ahal: Ok those are fair points! Doing this means there would be no documentation, so I'll add a "How To" page for this as well.

ahal:
Switched to your suggested schema format and added some docs!

docs/howto/send-notifications.rst
```diff
@@ -0,0 +1,189 @@
...
+   be configured with Slack credentials and be invited to the room.
+
+.. _status-types:
+
+Status Types
+------------
+
+A notification can be sent on the following task status changes:
+
+- ``on-defined`` - When the task is first submitted to Taskcluster.
```
JohanLorenzo, line 25: Good call on not supporting the deprecated `on-any`[1]:

> **on-any:** (DEPRECATED) This does what it sounds like and will notify you of task-completion no matter how it ends. Unless the task was deliberately canceled.

https://docs.taskcluster.net/docs/reference/core/notify/usage#filters

docs/howto/send-notifications.rst
```diff
@@ -0,0 +1,189 @@
...
+and ``room-id: <id>``, and a pulse route can be notified using ``type: pulse``
+and ``routing-key: <key>``.
+
+For each recipient, the ``status-type`` key can be any one of the
+aforementioned :ref:`status types <status-types>`. If not specified, it
+defaults to ``on-completed``.
+
+.. note::
+
+   For a full reference, see the :py:data:`notify schema <taskgraph.transforms.notify.NOTIFY_SCHEMA>`.
```
JohanLorenzo, line 69: 👍

docs/howto/send-notifications.rst
```diff
@@ -0,0 +1,189 @@
...
+Each recipient value can be interpolated using Python's `format string syntax`_
+and the ``task`` and :class:`kind config <taskgraph.generator.Kind>` objects as
+context. For example:
+
+.. code-block:: yaml
+
+   notify:
+     recipients:
+       - type: email
+         address: "{task[name]}-subscribers@example.com"
```
JohanLorenzo, line 83: Interesting! Do you know if any team has ever had the need for including the task name in the recipient?

docs/howto/send-notifications.rst
```diff
@@ -0,0 +1,189 @@
+Send Notifications
```
JohanLorenzo, line 1: Kudos for writing this extensive doc! 💯

JohanLorenzo (approved):
Looks great to me! This documentation is AWESOME! 💯 👍

docs/howto/send-notifications.rst
ahal: Not `task[name]` specifially, but Gecko does interpolate this value in a few places, e.g:
https://searchfox.org/mozilla-central/source/taskcluster/ci/release-notify-push/kind.yml#29

ahal:
Sorry for all the force pushes, I was using this PR to test a task that uses this transform in VPN. The slack blocks key was in the wrong format, but after some trial an error it's working now. Nothing substantive has changed since the review.

Issue #156, opened 2022-11-22
Upgrade Taskgraph to latest `image_builder` image


ahal:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1661637 we moved the `image_builder` image from the `taskcluster` user on DockerHub over to `mozillareleases` and made a new 5.0.0 image builder release.

This was upgraded in Gecko taskgraph, but here we're still stuck on the [older image](https://github.com/taskcluster/taskgraph/blob/c61b6a6ae0638784ac7557237990f96cb8090885/src/taskgraph/transforms/docker_image.py#L27).

PR #157, opened 2022-11-22
chore: update 'image_builder' image to 5.0.0

ahal:
Issue: #156


ahal:
I'll try testing this from some of the mobile repos.

ahal:
The test failure here is due to a problem in the Python docker-image, which ironically enough is completely unrelated to using the new image_builder image. I'm investigating over in https://github.com/taskcluster/taskgraph/pull/158. My guess is this either due to some change in `pyenv` or `xx-env`.

JohanLorenzo (approved):
I confirm the hash is the one stored on dockerhub: https://hub.docker.com/layers/mozillareleases/image_builder/5.0.0/images/sha256-e510a9a9b80385f71c112d61b2f2053da625aff2b6d430411ac42e424c58953f?context=explore

PR #158, opened 2022-11-22
fix: busted Python docker-image task

ahal:


ahal:
Looks like `pyenv` introduced a native `latest` command which interferes with the one from `xx-env`:
https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-latest

The good news is that it should be possible to remove our dependency on `xx-env` entirely now.

ahal:
Note that the `tests-unit` task is now failing due to an error installing dependencies in the `clean` tox environment. I'll look into this in another PR so we can keep this one on task.

ahal:
Note that the `tests-unit` task is now failing due to an error installing dependencies in the `clean` tox environment. This looks to be a separate issue from this PR, so I don't think we need to block landing this on it as it still improves things somewhat.

taskcluster/scripts/pyenv-setup
```diff
@@ -13,10 +15,11 @@ eval "\$(pyenv init -)"
 eval "\$(pyenv virtualenv-init -)"
 EOF
 source /builds/worker/.bashrc
-git clone https://github.com/momo-lab/xxenv-latest.git "$(pyenv root)"/plugins/xxenv-latest
+
 for i in ${1//,/ }
 do
-    pyenv latest install $i
+    version=$(pyenv latest --known $i)
+    pyenv install $version
```
JohanLorenzo, line 22: Nice, I see pyenv logs out what version (including the patch number) gets installed[1]:

```
Downloading Python-3.10.8.tar.xz...
-> https://www.python.org/ftp/python/3.10.8/Python-3.10.8.tar.xz
Installing Python-3.10.8...
Installed Python-3.10.8 to /builds/worker/.pyenv/versions/3.10.8
Downloading Python-3.9.15.tar.xz...
-> https://www.python.org/ftp/python/3.9.15/Python-3.9.15.tar.xz
Installing Python-3.9.15...
Installed Python-3.9.15 to /builds/worker/.pyenv/versions/3.9.15
Downloading Python-3.8.15.tar.xz...
-> https://www.python.org/ftp/python/3.8.15/Python-3.8.15.tar.xz
Installing Python-3.8.15...
Installed Python-3.8.15 to /builds/worker/.pyenv/versions/3.8.15
Downloading Python-3.7.15.tar.xz...
-> https://www.python.org/ftp/python/3.7.15/Python-3.7.15.tar.xz
Installing Python-3.7.15...
```

[1] https://firefox-ci-tc.services.mozilla.com/tasks/ZtcrXQW9SSmadkjrm9sqkQ/runs/0/logs/public/logs/live.log#L683

taskcluster/scripts/pyenv-setup
```diff
@@ -3,6 +3,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+set -e
```
JohanLorenzo, line 6: 👍 I see we don't have any secrets in this script. We can probably add for extra debugging too:

```suggestion
set -ex
```

taskcluster/scripts/pyenv-setup
```diff
@@ -3,6 +3,8 @@
...
+
 curl -L "https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer" -o /builds/worker/pyenv-installer
```
JohanLorenzo, line 8: Improvement unrelated to this PR: We should log out what revision we pulled from `master` for the sake of traceability.

JohanLorenzo (approved):
✅ :shipit: 

The coverage drop is interesting. I had a look at the report and it seems it doesn't evaluate function/class headers that are decorated. For sure, it's unrelated to this PR.

taskcluster/scripts/pyenv-setup
ahal: Good suggestion, I added another commit with this + `pyenv --version`.

ahal:
> The coverage drop is interesting. I had a look at the report and it seems it doesn't evaluate function/class headers that are decorated. For sure, it's unrelated to this PR.

Very strange.. I wonder if there's a bug on codecov's end or something.

ahal:
Now the unittest failure is due to an intermittent that has been happening on Python 3.6 lately (can't seem to find `appdirs` on pypi).. I'm not sure why this is happening, but it's unrelated to this PR as it's been happening on `main` as well.

PR #159, opened 2022-11-22
Py image parallel

ahal:


ahal:
This didn't speed it up at all, oh well.

PR #162, opened 2022-11-23
feat: add an action to rebuild cached tasks

ahal:
This copies the existing action in Gecko to standalone Taskgraph + adds a test.


hneiva (approved):
I'm not familiar with cached tasks, but syntax looks good

PR #172, opened 2022-12-12
fix: run-task git checkout for non-default branch

### LINKED ISSUE #170: Checkout in `run-task` fails in certain scenarios
Failure reference: https://firefox-ci-tc.services.mozilla.com/tasks/at-bWwQKQ4q_3RUfPgE45w/runs/0/logs/public/logs/live.log#L28

Example:
```shell
$ git clone https://github.com/mozilla-rally/rally-core-addon
Cloning into 'rally-core-addon'...
remote: Enumerating objects: 5038, done.
remote: Counting objects: 100% (32/32), done.
remote: Compressing objects: 100% (30/30), done.
remote: Total 5038 (delta 19), reused 4 (delta 2), pack-reused 5006
Receiving objects: 100% (5038/5038), 10.50 MiB | 7.72 MiB/s, done.
Resolving deltas: 100% (3153/3153), done.

$ cd rally-core-addon 

$ git fetch --no-tags https://github.com/mozilla-rally/rally-core-addon release
From https://github.com/mozilla-rally/rally-core-addon
 * branch            release    -> FETCH_HEAD

$ git checkout -f -B release release
fatal: 'release' is not a commit and a branch 'release' cannot be created from it
```

hneiva:
Closes #170

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -641,7 +641,7 @@ def git_checkout(
     # between the previous state before the force-push and the current state.
     #
     # Unlike base_ref just above, there is no need to checkout the revision:
-    # it's immediately avaiable after the fetch.
+    # it's immediately available after the fetch.
     if base_rev and base_rev != NULL_REVISION:
         args = ["git", "fetch", "origin", base_rev]
 
@@ -666,7 +666,7 @@ def git_checkout(
 
     if ref:
         args.extend(["-B", ref])
-    args.append(commit if commit else ref)
+    args.append(commit if commit else "FETCH_HEAD")  # Set by git fetch
 
     run_required_command(b"vcs", args, cwd=destination_path)
 
@@ -1090,7 +1090,7 @@ def main(args):
     # Sort repositories so that parent checkout paths come before children
     repositories.sort(key=lambda repo: Path(repo["checkout"] or "/").parts)
 
-    uid = gid = gids = None
+    uid = gid = gids = user = group = None
     if IS_POSIX and running_as_root:
         user, group, gids = get_posix_user_group(args.user, args.group)
         uid = user.pw_uid
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -666,7 +666,9 @@ def git_checkout(
 
     if ref:
         args.extend(["-B", ref])
-    args.append(commit if commit else "FETCH_HEAD")  # Set by git fetch
+
+    # `git fetch` set `FETCH_HEAD` reference to the last commit of the desired branch
+    args.append(commit if commit else "FETCH_HEAD")
 
     run_required_command(b"vcs", args, cwd=destination_path)
 
--- a/test/test_scripts_run_task.py
+++ b/test/test_scripts_run_task.py
@@ -47,6 +47,18 @@ def patch_run_command(monkeypatch, run_task_mod):
     return inner
 
 
+@pytest.fixture()
+def mock_stdin(monkeypatch):
+    _stdin = Mock(
+        fileno=lambda: 3,
+    )
+    monkeypatch.setattr(
+        sys,
+        "stdin",
+        _stdin,
+    )
+
+
 def test_install_pip_requirements(
     tmp_path,
     patch_run_command,
@@ -205,7 +217,7 @@ def _mark_readonly(path):
     os.chmod(path, mode & ~stat.S_IWUSR & ~stat.S_IWGRP & ~stat.S_IWOTH)
 
 
-def test_clean_git_checkout(monkeypatch, run_task_mod):
+def test_clean_git_checkout(monkeypatch, mock_stdin, run_task_mod):
     prefix = "Would remove "
     root_dir = tempfile.TemporaryDirectory()
     untracked_dir = tempfile.TemporaryDirectory(dir=root_dir.name)
@@ -240,20 +252,10 @@ def test_clean_git_checkout(monkeypatch, run_task_mod):
             wait=lambda: 0,
         )
 
-    _stdin = Mock(
-        fileno=lambda: 3,
-    )
-
     monkeypatch.setattr(
         subprocess,
         "Popen",
         _Popen,
-    )
-
-    monkeypatch.setattr(
-        sys,
-        "stdin",
-        _stdin,
     )
 
     assert os.path.isdir(root_dir.name) is True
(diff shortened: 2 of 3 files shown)


escapewindow (approved):
As mentioned in chat, we may want to add some unit tests to this since it keeps breaking things. Starting with the known failure cases (cinnabar and `ref != default branch`). This may save future headache when we change this logic again.

test/test_scripts_run_task.py
```diff
@@ -270,3 +272,56 @@ def _Popen(
...
+
+@pytest.fixture(scope="session")  # Tests shouldn't change this repo
+def mock_git_repo():
+    "Mock repository with files, commits and branches for using as source"
+    with tempfile.TemporaryDirectory() as repo:
+        repo_path = str(repo)
+        subprocess.check_call(["git", "init", "-b", "main"], cwd=repo_path)
+
+        def _commit_file(message):
+            untracked_file = tempfile.NamedTemporaryFile(dir=repo, delete=False)
```
jcristau, line 285: Should we use a non-random name here, and in the test check that the file exists (if we're on that branch)?

test/test_scripts_run_task.py
```diff
@@ -270,3 +272,56 @@ def _Popen(
...
+
+        _commit_file("Initial commit")
+        subprocess.check_call(["git", "checkout", "-b", "mybranch"], cwd=repo_path)
+        _commit_file("File in my branch")
+        subprocess.check_call(["git", "checkout", "main"], cwd=repo_path)
+        yield repo_path
+
+
+@pytest.mark.parametrize(
+    "base_ref,ref",
```
jcristau, line 299: Should we also test passing a non-None commit?

src/taskgraph/run-task/run-task
```diff
@@ -1090,7 +1090,7 @@ def main(args):
     # Sort repositories so that parent checkout paths come before children
     repositories.sort(key=lambda repo: Path(repo["checkout"] or "/").parts)
 
-    uid = gid = gids = None
+    uid = gid = gids = user = group = None
```
jcristau, line 1093: From a quick look it seems we could do without user and group and just pass uids around (the only thing that uses pw_name/gr_name is a logging message in chown_recursive, which could just as well use numeric IDs.

jcristau (approved):
LGTM, a couple of comments that could be ideas for followups.

test/test_scripts_run_task.py
JohanLorenzo: I understand this is going to be trickier to test defined commit hashes. On the bright side, commit hashes are predictable. We can set all parameters to known values and hardcode the hash we want to pass:

> The git hash is made up of the following:
>
> *    The commit message
> *    The file changes
> *    The commit author (and committer- they can be different)
> *    The date
> *    The parent commit hash

_Source: https://www.mikestreety.co.uk/blog/the-git-commit-hash/_

The parent commit hash is zero[1] when you create the first commit hash. `freezegun`[2] will help to set the date. The remaining params are already set by your test. 

With this, we should be able to test passing non-None commits that reflect how this function is called in real life 🙂 

[1] https://github.com/git/git/blob/dc04167d378fb29d30e1647ff6ff51dd182bc9a3/t/oid-info/hash-info#L7
[2] https://pypi.org/project/freezegun/

test/test_scripts_run_task.py
```diff
@@ -270,3 +272,56 @@ def _Popen(
...
+            destination_path=destination,
+            head_repo=mock_git_repo,
+            base_repo=mock_git_repo,
+            base_ref=base_ref,
+            base_rev=None,
+            ref=ref,
+            commit=None,
+            ssh_key_file=None,
+            ssh_known_hosts_file=None,
+        )
```
JohanLorenzo, line 327: Test: let's assert the repo is checked out to the actual revision we're expecting. This can be done by asserting whether a given file exists or now.

src/taskgraph/run-task/run-task
```diff
@@ -666,7 +666,7 @@ def git_checkout(
 
     if ref:
         args.extend(["-B", ref])
-    args.append(commit if commit else ref)
+    args.append(commit if commit else "FETCH_HEAD")  # Set by git fetch
```
JohanLorenzo, line 669: I'd suggest we provide a little bit more context for future readers. How about something like:

```suggestion
    # The last `git fetch` didn't set any remote repo it just reset this `FETCH_HEAD` commitish
    args.append(commit if commit else "FETCH_HEAD")
```

?

JohanLorenzo (commented):
LGTM overall. Thank you very much for adding tests to these changes. For sure, we will able to catch the kind of failures you documented in #170. The test is currently red on CI (which is a simple fix).

Either in this PR or in a follow-up, I believe we should make the test even stronger by verifying we actually checked out the expected commitish (branch, commit hash, tag, `HEAD~N`, etc.). I left some pointers down below but I'm happy to provide additional info 🙂

src/taskgraph/run-task/run-task
hneiva: I really only added these here so my IDE would stop complaining it wasn't set on line 1131

test/test_scripts_run_task.py
hneiva: I feel like this is over-engineering. For now I just pulled the hashes created by the fixture, we should be able to trust the `git` program, and the target of the test is to cover the business logic of run-task.

Maybe for a future release we could use a python library to handle git? https://gitpython.readthedocs.io/en/stable/tutorial.html and the testing wrapper https://pypi.org/project/pytest-git/

JohanLorenzo: > For now I just pulled the hashes created by the fixture, we should be able to trust the git program, and the target of the test is to cover the business logic of run-task.

I 100% agree with you 🙂 We shouldn't test `git` itself and we should limit ourselves to what `run-task` does. That said, this line: https://github.com/taskcluster/taskgraph/commit/df1b39b70b93a119e0d8dd64c3ceff180331836e#diff-e29e172140c458b95d3f5b1a5542188ffe334182ccef1c3b62e1eeb102beec1aR671 is part of `run-task` and we may regress it one day. So having an end-to-end test that ensures we are able to checkout a known revision will help us from regressing our business logic. 

In any case, I think you were right to merge this PR, yesterday! Without your PR, we didn't have any coverage there so it's definitely an improvement 👍 We can address the test in a follow-up but it can be done much later since we've lived without tests for year there. 

> Maybe for a future release we could use a python library to handle git?

Good idea! I think we should tackle this once we drop the hg support. The current logic between `hg` and `git` is very close. If we change `git` but not `hg`, it'll be harder to see patterns between the 2. 

How does that sound?

hneiva: Sounds good. Another way of testing this is mocking out the subprocess calls and check for units. Ie: Given input x,y,z we expect git to be called with commit/branch y. That way we don't have to fiddle with the git internals, we just "trust" it will work.

JohanLorenzo: Works for me too 🙂

PR #174, opened 2022-12-14
chore: version bump Taskgraph 4.1.2

hneiva:


version.txt
```diff
@@ -1 +1 @@
-4.1.0
+4.2.0
```
ahal, line 1: I think we should bump to `4.1.1` as a minor upgrade would imply there's some sort of new feature.

ahal (changes requested):

version.txt
hneiva: Works for me

ahal (approved):
Thanks! Note the commit message still says 4.2.0

ahal:
Superceded by https://github.com/taskcluster/taskgraph/pull/200

Issue #137, opened 2022-10-20
Cannot run `pip-compile-multi` anymore: Could not find a version that matches importlib-resources<3.5,>=3.0,>=5.4


JohanLorenzo:
## Steps to reproduce

Run: 

```sh
docker run -t -v "$PWD:/src" -w /src python:3.6 bash -cx "pip install pip-compile-multi && pip-compile-multi --generate-hashes base --generate-hashes dev --generate-hashes test --allow-unsafe"
```

## Expected result

All requirements get bumped

## Actual result

Only `base.txt` does. The script fails to handle `test.txt` (which also blocks `dev.txt`). It errors out this way:

```
Finding the best candidates:
  found candidate alabaster==0.7.12 (constraint was >=0.7,<0.8)
  [...]
  found candidate importlib-metadata==4.8.3 (constraint was >=0.12,>=4.8.3)
Could not find a version that matches importlib-resources<3.5,>=3.0,>=5.4 (from sphinx-book-theme==0.2.0->-r requirements/test.in (line 11))
[...]

PR #175, opened 2022-12-15
chore: pin packages + docker pin-helper script - DROPS PY36

hneiva:
Assumes sphinx will always run on python 3.9+ (as set in .readthedocs.yaml) Had to split into it's own requirements file due to conflict with upstream tox package pinning


hneiva:
It looks like we'll have to either:
a. Drop py3.6 support
b. Manage requirement files versioning (can be done with docker) ie: base.py36.in etc

Side note that I had a lot of issues with sphinx and py3.6

tox.ini
```diff
@@ -9,7 +9,7 @@ depends =
     report: py{37,38,39,310}
 commands =
     python --version
-    pytest -vv --cov=taskgraph --cov=src/taskgraph/run-task --cov-append --cov-report= {posargs}
+    pytest -n auto -vv --cov=taskgraph --cov=src/taskgraph/run-task --cov-append --cov-report= {posargs}
```
hneiva, line 12: I'm not sold on the parallel test idea, but locally at least it saves multiple minutes per run, and doesn't seem to cause any issues. Thoughts?

hneiva:
looks like `coverage` (required by `pytest-cov`) has been having issues with sql: https://github.com/nedbat/coveragepy/issues/1303 and it's most frequest in github actions. I'm unable to reproduce the error locally

tox.ini
JohanLorenzo: Fine by me! We can always revert this small change if we face too many intermittent errors.

pin-helper.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+PIN_COMMAND='
+set -e
+pip install -q -U pip pip-compile-multi
+pip-compile-multi -u \
+    -g base -g test -g dev -g docs
```
JohanLorenzo, line 10: Nit: should we pin `docs` on python 3.9 instead of 3.7?

pin-helper.sh
```diff
@@ -0,0 +1,27 @@
...
+
+set +x
+echo "###########################################################"
+echo "# Running tests on different python versions              #"
+echo "# Pinned packages already saved. You may skip with CTRL+C #"
+echo "###########################################################"
+set -x
+
+# Don't quit if a version fails
+set +e
```
JohanLorenzo, line 24: I guess the loop is to highlight failures in case the generated dependencies don't work on a later version of Python. If so, what's the use case for letting these commands silently fail?

JohanLorenzo (approved):
Fixes https://github.com/taskcluster/taskgraph/issues/137

Thanks a lot for looking into this issue. I'm in favor of dropping Python 3.6. I just checked what gecko uses and it's already on Python3.8[1]:

```bash
$ docker pull mozillareleases/gecko_decision:4.0.0@sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06
docker.io/mozillareleases/gecko_decision@sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06: Pulling from mozillareleases/gecko_decision
16ec32c2132b: Pull complete
1c6fb07fba89: Pull complete
e2066ab6a02e: Pull complete
08a664befb8a: Pull complete
Digest: sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06
Status: Downloaded newer image for mozillareleases/gecko_decision@sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06
docker.io/mozillareleases/gecko_decision:4.0.0@sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06


$ docker run -ti mozillareleases/gecko_decision:4.0.0@sha256:9f69fe08c28e3cb3cc296451f0a2735df6e25d0e3c877ea735ef1b7f0b345b06 bash
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
root@175b8e25177f:~# python --version
Python 2.7.18
root@175b8e25177f:~# python3 --version
Python 3.8.10
```

[1] https://searchfox.org/mozilla-central/rev/f40d29a11f2eb4685256b59934e637012ea6fb78/.taskcluster.yml#226

jcristau:
My understanding was we're not ready to drop 3.6 because taskgraph is used (or at least imported) in parts of the firefox build system, and we don't want to break downstream builds.  Did that change?

test/test_util_vcs.py
```diff
@@ -83,10 +82,18 @@ def _init_repo(tmpdir, repo_type):
 
 
 @pytest.fixture(params=("git", "hg"))
-def repo(request, hg_repo, git_repo):
+def repo(request, hg_repo, git_repo, monkeypatch, tmpdir):
+    """
+    The repo fixture depends on the session-scoped git and hg repo fixtures, and
+        copies the contents of the initialized repos to a temp folder
+    """
```
hneiva, line 89: This should improve the reliability on the tests as well, as it reduces runtime in a few minutes, I'm hoping it's less likely for pytest to have issues with sqlite

hneiva:
> My understanding was we're not ready to drop 3.6 because taskgraph is used (or at least imported) in parts of the firefox build system, and we don't want to break downstream builds. Did that change?

Aiui, as far as the requirements go, py36 isn't supported (even in main) because of tox, sphinx and others. That being said, we can still import and use taskgraph in py36.

test/test_util_vcs.py
```diff
@@ -29,7 +29,7 @@ def hg_repo(tmpdir, monkeypatch):
 
     # hg sometimes errors out with "nothing changed" even though the commit succeeded
     subprocess.call(
-        ["hg", "commit", "-m", "First commit", "--date", _FORCE_COMMIT_DATE_TIME],
+        ["chg", "commit", "-m", "First commit", "--date", _FORCE_COMMIT_DATE_TIME],
```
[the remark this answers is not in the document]
hneiva: Not really, but I found a problem when running chg, I'll drop it for now

pin-helper.sh
hneiva: That was my initial thought. But once I talked to ahal, we decided to go with the minimum supported version of python, as long as it causes no problem to the version that rtd is running.

pin-helper.sh
hneiva: Improved it now

kenahoo:
> looks like `coverage` (required by `pytest-cov`) has been having issues with sql: [nedbat/coveragepy#1303](https://github.com/nedbat/coveragepy/issues/1303) and it's most frequest in github actions. I'm unable to reproduce the error locally

@hneiva I whipped up a sample repo that demonstrates the problem, at [nedbat/coveragepy#1514].

test/test_util_vcs.py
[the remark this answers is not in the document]
ahal: It's a command server written in C that can bypass the startup latency caused by Python. Makes things a bit faster, though in this case I don't think it'll make a big difference vs the other improvements later on.

ahal (changes requested):
Requesting changes until CI can be fixed.

PR #178, opened 2023-01-03
perf: Optimize `util.hash.hash_paths()` when providing a dozen-plus patterns (bug 1807237)

JohanLorenzo:
I'm working on [bug 1807237](https://bugzilla.mozilla.org/show_bug.cgi?id=1807237) on [firefox-android](https://github.com/mozilla-mobile/firefox-android). There, I'm creating a hundred new tasks that all hash 10+ patterns. Interestingly, these patterns reference a single file but the decision task took more than an hour and a half to generate these 100 tasks.

The optimization below generate these 100 tasks in less than a second, instead. 

Blocks https://github.com/mozilla-mobile/firefox-android/pull/490


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## [4.2.0] - TBD
```
ahal, line 3: Generally I don't bother pre-populating the changelog as it's not possible to tell what version number is incremented, but this works, we can always adjust :). I'm also unsure if perf improvements count as a feature or not under semver, I could see it going either way.

src/taskgraph/main.py
```diff
@@ -439,6 +439,7 @@ def show_taskgraph(options):
         diffcmd = [
             "diff",
             "-U20",
+            "--color=always",
```
ahal, line 442: Hm, I think this will break piping the output into other commands which I do on occasion. It may also send control characters if redirecting into a file (like the `tgdiff` task does in Gecko).

To work around this in Gecko I created a setting that people can add to their `~/.machrc` file:
https://searchfox.org/mozilla-central/source/taskcluster/mach_commands.py#30

Though we don't have the benefit of that here.. Maybe we could look for an environment variable instead or something.

ahal (changes requested):
I'm pretty sure the `--color=always` is going to cause problems (at the least it will cause me minor annoyances :p).

I think this could be pulled into a separate PR so we can land the perf improvement faster.

src/taskgraph/main.py
JohanLorenzo: You're right. I removed this line. I'll just pipe to `colordiff` on my local machine.

CHANGELOG.md
JohanLorenzo: I just checked https://semver.org/ again. I'm actually fixing a perf bug. Let's just increment the last digit then. I also made a commit to release 4.1.1 today.

ahal (approved):
Nice find!

PR #179, opened 2023-01-09
feat: Add testing on python 3.11

hneiva:


ahal (approved):
Thanks!

taskcluster/ci/tests/kind.yml
```diff
@@ -50,5 +50,5 @@ tasks:
             command: >-
                 pip install --user -r requirements/test.txt --require-hashes &&
                 pip install --user --no-deps . &&
-                pyenv local 3.10 3.9 3.8 3.7 3.6 &&
+                pyenv local 3.11 3.10 3.9 3.8 3.7 3.6 &&
```
hneiva, line 53: and here

taskcluster/ci/docker-image/kind.yml
```diff
@@ -29,4 +29,4 @@ tasks:
     python:
         symbol: I(py)
         args:
-            PYENV_VERSIONS: "3.10, 3.9, 3.8, 3.7, 3.6"
+            PYENV_VERSIONS: "3.11, 3.10, 3.9, 3.8, 3.7, 3.6"
```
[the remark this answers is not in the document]
ahal: I was wondering that too.. I suppose we should. I don't think it hurts to leave it, but might as well remove if unused.

taskcluster/ci/tests/kind.yml
ahal: Yeah, I assume it's already removed from tox right? In that case no point in saving it. Would be nice to try and somehow share the list of supported versions better, but I don't have any clever suggestions

tox.ini
```diff
@@ -1,25 +1,25 @@
...
-    pytest -vv --cov=taskgraph --cov=src/taskgraph/run-task --cov-append --cov-report= {posargs}
+    coverage run -p --branch --source=src/taskgraph -m pytest -n auto -vv {posargs}
 
 [testenv:report]
 deps = -r{toxinidir}/requirements/test.txt
 skip_install = true
 passenv = COVERAGE_REPORT_COMMAND
 parallel_show_output = true
 commands =
-    {env:COVERAGE_REPORT_COMMAND:coverage report}
```
ahal, line 22: This env was used in `taskcluster/ci/tests/kind.yml` to set the save location to the artifacts dir. You'll need to re-implement it so that coverage artifact is in the right place when running via CI.

ahal (changes requested):

tox.ini
hneiva: Ahhh, I never intended to fully remove it.

ahal (approved):
Thanks for fixing this and improving the perf!

ahal (changes requested):
Hm, looks like the project coverage dropped 48% here, so I'm guessing something still isn't quite right.

PR #180, opened 2023-01-11
ci: prefer $switch over $if/else in .taskcluster.yml

ahal:


bhearsum (approved):
Thank you so much for doing this, it's great!

ahal:
I'm not sure what's up with the coverage thinking I'm adding lines in `hash.py`? Maybe Heitor's patch is going to fix that..

JohanLorenzo:
I wonder if it's just some red herring because your base commit was a month old. b18a7a7 was landed prior to https://github.com/taskcluster/taskgraph/pull/178 which did change `hash.py`: 

image

JohanLorenzo:
By the way, I'm so glad to see these clean `$switch` statements!

PR #181, opened 2023-01-13
vcs: make get_commit_message actually use its "revision" argument

jcristau:


jcristau:
The only test for `get_commit_message` doesn't pass an argument hence why this wasn't caught before.  I can look into adding more if you'd like.  I noticed this while working on https://github.com/mozilla-mobile/firefox-android/pull/542 where running taskgraph locally came up with HEAD's commit message instead of the one from the commit listed in the params file.

src/taskgraph/util/vcs.py
```diff
@@ -225,8 +225,8 @@ def get_url(self, remote="default"):
         return self.run("path", "-T", "{url}", remote).strip()
 
     def get_commit_message(self, revision=None):
-        revision = revision or self.head_rev
-        return self.run("log", "-r", ".", "-T", "{desc}")
+        revision = revision or "."
```
JohanLorenzo, line 228: Nit: I wonder if we should use `self.head_rev` here too for the sake of being consistent with the git implementation. Plus, `head_rev` may be different from `.` when running taskgraph with a parameter file: the checked out code is likely going to be different from the `head_rev` defined in the file. 

That said, I understand the current implementation matches what we had implemented. Thus, I wonder if we should leave a comment that we're aware of this discrepancy and we should likely fix it. 

How does that sound to you?

JohanLorenzo (approved):

src/taskgraph/util/vcs.py
jcristau: `self.head_rev` doesn't look at the params file, it's just HEAD's hash, so it's equivalent to `.`; I actually considered also changing `get_commit_message` on git to use `HEAD` directly instead of keeping the `self.head_rev` indirection, since that seems like it'd just be slower for no particular reason.

JohanLorenzo: You're right! I had forgotten this was the hg implementation:

https://github.com/taskcluster/taskgraph/blob/2f00244ef2524668845412b5f3563ce9b91cbc9f/src/taskgraph/util/vcs.py#L188-L189

You're right, using the magic string `HEAD` makes more sense! Let's use it, then 🙂

jcristau: Done!

PR #185, opened 2023-01-31
feat: support extra 'run-task' features in toolchains

ahal:
This ensures we forward unrecognized schema arguments to the `run_task.py` transforms. Validation of these arguments will then happen over there.


bhearsum (approved):

ahal:
Not sure what's up with `pre-commit`, but I'm assuming it's an infra problem in their CI and not something caused by this patch.

Issue #189, opened 2023-02-20
DONTBUILD keyword is ignored on `github-push` events


JohanLorenzo:
The `DONTBUILD` keyword is a feature we implemented in taskgraph. If a commit message contains `DONTBUILD` then taskgraph doesn't schedule any taskcluster tasks ([e.g.](https://treeherder.mozilla.org/jobs?repo=mozilla-beta&revision=165cce29e43891b30dbefdcd242815d96d470691)). It works pretty well on mercurial but last week I found it doesn't work at all on Github. See this [push of firefox-android](https://treeherder.mozilla.org/jobs?repo=firefox-android&revision=b2605e96d90af6cb5894b5877bbf10fb19e7faf0). 

## Steps to reproduce

1. Follow the alternative installation steps https://github.com/taskcluster/taskgraph#installation
3. `touch FOO && git add FOO && git commit -m 'Add foo DONTBUILD` 
4. `export TASKCLUSTER_ROOT_URL='https://firefox-ci-tc.services.mozilla.com' && export TASK_ID='dummyTask'`
5. ```sh
    taskgraph decision
      --pushlog-id='0' \
      --pushdate='0' \
      --project='taskgraph' \
      --owner='some@email.tld' \
      --level='1' \
      --base-repository='https://github.com/taskcluster/taskgraph' \
      --base-ref='main' \
      --base-rev='HEAD~1' \
      --head-repository='https://github.com/taskcluster/taskgraph' \
      --head-ref='main' \
      --head-rev='HEAD' \
      --repository-type="git" \
      --tasks-for='github-push' \
      --message="Add foo DONTBUILD" \
    ```

## Expected results

The script runs successfully and and more importantly `filter_target_tasks` should prune all tasks: 

```
2023-02-20 15:13:21,469 - INFO - Filter filter_target_tasks pruned 9 tasks (0 remain)
```

## Actual results

The script fails because it schedules tasks, but more `filter_target_tasks` doesn't filter out every task.

```
2023-02-20 15:09:24,135 - INFO - Filter filter_target_tasks pruned 6 tasks (3 remain)
```

## Side notes

I recommend searching the string `DONTBUILD`. It will highlight where the fix should happen. The fix is likely a one-liner but we can use this opportunity to add more unit tests. 

After the fix, the script will still fail for another reason. Let's handle that other reason in https://github.com/taskcluster/taskgraph/issues/190.

miladHakimi:
Hi @JohanLorenzo,
[...]

Issue #190, opened 2023-02-20
`build-docker` tasks shouldn't be silently injected in optimized graph when `DONTBUILD` is provided


JohanLorenzo:
Found while looking at https://github.com/taskcluster/taskgraph/issues/189. `DONTBUILD` makes sure no tasks are scheduled. However, it's not the case today. Today, `build-docker` images are still pulled in between the `target` phase and the `optimized`. I don't think we should keep this behavior.

Marking it as a good second bug because it looks like a good follow-up on https://github.com/taskcluster/taskgraph/issues/189.

jcristau:
I'm not convinced this is a bug FWIW...

JohanLorenzo:
@jcristau and I had a thorough conversation on Element about this. We looked back at the context when things got added. 

## Some archeology

The first step was `build-docker`. These tasks have been silently injected in the optimized graph since [a38447eb85d9](https://hg.mozilla.org/mozilla-central/rev/a38447eb85d995054d879c0b642b939615534246). The original intent is documented in [bug 1333255 comment 8](https://bugzilla.mozilla.org/show_bug.cgi?id=1333255#c8). That's when the `morph` phase was introduced in order to support tasks with more than 10 indices. 

2 years later, `DONTBUILD` was introduced [e44be7f40c6d](https://hg.mozilla.org/mozilla-central/rev/e44be7f40c6daf7a5858aee63626805527427f41). The intent at that time was to let taskgraph schedule no tasks at all. Per [bug 1534283 comment 15](https://bugzilla.mozilla.org/show_bug.cgi?id=1534283#c15), if a task needs to be targeted anyway, then the task attribute [`always-target`](https://searchfox.org/mozilla-central/rev/0e3779abe6ad44feb787a295fbb67bc9fe4e927a/taskcluster/docs/attributes.rst#263-274) is the way to proceed. It's also [documented in this comment](https://github.com/taskcluster/taskgraph/blob/f584cf13e0667b1972781dac31a8b3929f43134e/src/taskgraph/transforms/task.py#L154-L159).

## What is today's problem?

To expand on the my first comment and on #189, here were the logs I got with a `DONTBUILD` commit: 

```
2023-02-20 15:13:21,468 - INFO - Generating full task graph
2023-02-20 15:13:21,468 - INFO - Full task graph contains 9 tasks and 7 dependencies
2023-02-20 15:13:21,468 - INFO - writing artifact file `full-task-graph.json`
[...]

PR #191, opened 2023-02-23
fix: `DONTBUILD` on `github-push`.

miladHakimi:
fixes: #189

### DIFF
--- a/src/taskgraph/decision.py
+++ b/src/taskgraph/decision.py
@@ -227,7 +227,9 @@ def get_decision_parameters(graph_config, options):
     # ..but can be overridden by the commit message: if it contains the special
     # string "DONTBUILD" and this is an on-push decision task, then use the
     # special 'nothing' target task method.
-    if "DONTBUILD" in commit_message and options["tasks_for"] == "hg-push":
+    if "DONTBUILD" in commit_message and (
+        options["tasks_for"] in ("hg-push", "github-push")
+    ):
         parameters["target_tasks_method"] = "nothing"
 
     if options.get("optimize_target_tasks") is not None:
--- a/test/test_decision.py
+++ b/test/test_decision.py
@@ -84,6 +84,22 @@ class TestGetDecisionParameters(unittest.TestCase):
         params = decision.get_decision_parameters(FAKE_GRAPH_CONFIG, self.options)
         self.assertEqual(params["owner"], "ffxbld@noreply.mozilla.org")
 
+    def test_dontbuild(self):
+        repo_mock = unittest.mock.MagicMock()
+        repo_mock.commit_message = "Add FOO DONTBUILD"
+
+        self.options["tasks_for"] = "github-release"
+        params = decision.get_decision_parameters(FAKE_GRAPH_CONFIG, self.options)
+        self.assertNotEqual(params["target_tasks_method"], "nothing")
+
+        self.options["tasks_for"] = "github-push"
+        params = decision.get_decision_parameters(FAKE_GRAPH_CONFIG, self.options)
+        self.assertEqual(params["target_tasks_method"], "nothing")
+
+        self.options["tasks_for"] = "hg-push"
+        params = decision.get_decision_parameters(FAKE_GRAPH_CONFIG, self.options)
+        self.assertEqual(params["target_tasks_method"], "nothing")
+
 
 @pytest.mark.parametrize(
     "candidate_base_ref, base_rev, expected_base_ref",


src/taskgraph/decision.py
```diff
@@ -227,7 +227,9 @@ def get_decision_parameters(graph_config, options):
     # ..but can be overridden by the commit message: if it contains the special
     # string "DONTBUILD" and this is an on-push decision task, then use the
     # special 'nothing' target task method.
-    if "DONTBUILD" in commit_message and options["tasks_for"] == "hg-push":
+    if "DONTBUILD" in commit_message and (
+        options["tasks_for"] == "hg-push" or options["tasks_for"] == "github-push"
+    ):
```
JohanLorenzo, line 232: Nit: Python provides the `in` keyword[1] to make this condition less verbose. Here's an example[2] Let's use it here too 🙂 

[1] https://docs.python.org/3/reference/expressions.html#in
[2] https://github.com/taskcluster/taskgraph/blob/034305f5f77d2c6a0c2d0ca851ae683a9cb07d01/src/taskgraph/docker.py#L49

JohanLorenzo (commented):
Hi @miladHakimi! 

Thank you very much your interest in taskgraph and for making this patch! You're right, you found the right place to make the fix 👍 

There's a little nit to make the code more pythonic. That said, let's also seize this opportunity to improve the code coverage. Would you like to add a new unit like like this one? ⬇️ 

https://github.com/taskcluster/taskgraph/blob/936142a8c73510d0a0b530ac4a9a16d54cced6c9/test/test_decision.py#L82-L85

src/taskgraph/decision.py
miladHakimi: Thanks for the kind words!
I hope the test is correct.
I'm also not sure if I have to include the `github-pull-request` to the `DONTBUILD` list in `decision.py` or not.

JohanLorenzo: The test looks great! We could split the cases into multiple ones but we would have to convert the test class into a `pytest` case. No need to do it in this PR, this could be done in a follow-up one 🙂 

>  I'm also not sure whether I have to include the github-pull-request to the DONTBUILD list in decision.py or not.

That's a great question! Most of the usage of our `DONTBUILD` usage comes from CD (continuous deployment). For example: this task[1] created that `DONTBUILD` commit. We also use `DONTBUILD` for commits related to documentation and trivial fixes.

We could allow `DONTBUILD` on pull requests but I'm unsure of the side effects. A PR could contain multiple commits. If the last one has `DONTBUILD`, then the whole stack of commits won't be tested at all. I'm not sure we should do it today. That said, we may want to make taskgraph smarter about pull requests and detect whether a PR contains multiple commits and still build if one commit omits `DONTBUILD`. However, this improvement is definitely outside of the scope of https://github.com/taskcluster/taskgraph/issues/189 🙂

[1] https://firefox-ci-tc.services.mozilla.com/tasks/NU4C9bT8TLm6e-re1YfSdA
[2] https://treeherder.mozilla.org/jobs?repo=mozilla-beta&revision=c708a6fd4bfbf7dccf2a70b52fbbf2dfb0056a02

jcristau: IMO DONTBUILD on PRs should be ignored.

JohanLorenzo (approved):
Your changes look great 💯 Thank you again for taking the time to fix this issue 👍 I'm going to merge your PR as is. 

If you'd like to go further, I see 3 changes that directly follow this PR: 
 1. Convert `TestGetDecisionParameters` to use [pytest](https://docs.pytest.org/en/6.2.x/). Pytest is our preferred testing framework. Some of the tests cases like `TestGetDecisionParameters` are legacy. Using pytest will enable to parametrize and break down `test_dontbuild()`. See this example: 

https://github.com/taskcluster/taskgraph/blob/936142a8c73510d0a0b530ac4a9a16d54cced6c9/test/test_decision.py#L88-L107

 2. Make taskgraph smarter handle `DONTBUILD` in `github-pull-request` (see above)
 3. Take a stab at https://github.com/taskcluster/taskgraph/issues/190 

Let me know if you're interested in one of these things or if you'd like to try something else.

In the meantime, thank you! 😃

JohanLorenzo:
> The test looks great!

Oops, I missed one thing in my review. `test_dontbuild()` was green because the current commit message contains the word `DONTBUILD`. It wasn't the case anymore in the following commit, which lead to this revert #198. 

We should inject `repo_mock` in `get_decision_parameters()` by telling unittest to return the mock at this line. This can be achieved with the `@patch` decorator[2] on top of the function `test_dontbuild()`

[1] https://github.com/taskcluster/taskgraph/blob/54543ae5e7073f2a3a63ca1e149835ca53537d6e/src/taskgraph/decision.py#L165
[2] https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

src/taskgraph/decision.py
miladHakimi: I see. Thanks for the thorough explanation.

miladHakimi:
> > The test looks great!
> 
> Oops, I missed one thing in my review. `test_dontbuild()` was green because the current commit message contains the word `DONTBUILD`. It wasn't the case anymore in the following commit, which lead to this revert #198.
> 
> We should inject `repo_mock` in `get_decision_parameters()` by telling unittest to return the mock at this line. This can be achieved with the `@patch` decorator[2] on top of the function `test_dontbuild()`
> 
> [1]
> 
> https://github.com/taskcluster/taskgraph/blob/54543ae5e7073f2a3a63ca1e149835ca53537d6e/src/taskgraph/decision.py#L165
> 
> 
> [2] https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

Oh I see. I'll fix it.

miladHakimi:
I took a stab at the task and I have a challenge right now. 
The `get_repository` returns a `Repository` object. The commit message is not a member variable of this class and the `get_commit_message()` actually reads the commit message of the repository in the given directory. So, I have to mock this method as well. I tried to do this:
```python
    @unittest.mock.patch('taskgraph.util.vcs.get_repository', return_value=unittest.mock.MagicMock())
    def test_dontbuild(self, repo):
        repo.get_commit_message.return_value = "DONTBUILD"
``` 
But it didn't work. I couldn't make the `get_repository` to return what I want. Since the `get_repository` returns a `Repository`, I tried to mock the `get_commit_message()` for both `GitRepository` and `HgRepository`:
```python
    @unittest.mock.patch.object(
        GitRepository,
        "get_commit_message",
        unittest.mock.MagicMock(return_value="DONTBUILD"),
    )
    @unittest.mock.patch.object(
        HgRepository,
        "get_commit_message",
        unittest.mock.MagicMock(return_value="DONTBUILD"),
    )
    def test_dontbuild(self):
        self.options["tasks_for"] = "github-release"
        params = decision.get_decision_parameters(FAKE_GRAPH_CONFIG, self.options)
        self.assertNotEqual(params["target_tasks_method"], "nothing")
``` 
I can send this as a new PR.

JohanLorenzo:
Wonderful! I agree, let's open a new PR for this. I'd recommend adding another test case to make sure `target_tasks_method` equals `default` when another commit message is provided. This way, we're sure we won't hit the same type of regression. How does that sound to you @miladHakimi?

miladHakimi:
> Wonderful! I agree, let's open a new PR for this. I'd recommend adding another test case to make sure `target_tasks_method` equals `default` when another commit message is provided. This way, we're sure we won't hit the same type of regression. How does that sound to you @miladHakimi?

It sounds reasonable.

PR #193, opened 2023-02-23
Mock out decision._determine_more_accurate_base_ref in decision tasks

bhearsum:
We already mock out decision._determine_more_accurate_base_rev, so I'm guessing we want to do this as well. Without it, these tests end up running some `util/vcs.py` code against your local taskgraph repo. In my case, this [fails to guess the default branch](https://github.com/taskcluster/taskgraph/blob/936142a8c73510d0a0b530ac4a9a16d54cced6c9/src/taskgraph/util/vcs.py#L399-L412) because I have multiple remotes configured.

### DIFF
--- a/test/test_decision.py
+++ b/test/test_decision.py
@@ -66,10 +66,17 @@ class TestGetDecisionParameters(unittest.TestCase):
             decision._determine_more_accurate_base_rev
         )
         decision._determine_more_accurate_base_rev = lambda *_, **__: "abcd"
+        self.old_determine_more_accurate_base_ref = (
+            decision._determine_more_accurate_base_ref
+        )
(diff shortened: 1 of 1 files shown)


bhearsum:
@JohanLorenzo - it looks like you authored this code in https://github.com/taskcluster/taskgraph/commit/78d63f87ba6962def6aabeaf1bb63975afdb2803 and https://github.com/taskcluster/taskgraph/commit/9f26daa527f1044a8a387bafbd0e7c00fd79ce31 -- maybe you remember the original intent?

JohanLorenzo (approved):
I took a look at both revisions and that's an overlook of mine. Thanks for fixing it up! 👍

PR #196, opened 2023-02-24
Ignore as much outside VCS config as possible to minimize test interference

bhearsum:


test/test_util_vcs.py
```diff
@@ -14,6 +14,14 @@
 
 _FORCE_COMMIT_DATE_TIME = "2019-11-04T10:03:58+00:00"
 
+# Disable as much system/user level configuration as we can to avoid
+# inteference wit htests.
+# This ignores ~/.hgrc
+os.environ["HGRCPATH"] = ""
+# This ignores system-level git config (eg: in /etc). There apperas to be
```
jcristau, line 21:
```suggestion
# This ignores system-level git config (eg: in /etc). There appears to be
```

bhearsum:
I don't see how this patch could possibly decrease code coverage in `decision.py` as codecov.io claims....I'm going to go ahead and merge this. If I've missed something I'm happy to follow-up.

PR #201, opened 2023-02-28
fix: `dontbuild` on `github-push`.

miladHakimi:
If the commit message of a `github-push` or `hg-push` contains `dontbuild` (all caps), the build should be skipped. Previously, this was supported for only `hg-push`. This commit adds support for `github-push` and provides tests to make sure the behavior is as expected.

fixes: #189 #191

### DIFF
--- a/test/test_decision.py
+++ b/test/test_decision.py
@@ -13,6 +13,7 @@ from pathlib import Path
 import pytest
 
 from taskgraph import decision
+from taskgraph.util.vcs import GitRepository, HgRepository
 from taskgraph.util.yaml import load_yaml
 
 FAKE_GRAPH_CONFIG = {"product-dir": "browser", "taskgraph": {}}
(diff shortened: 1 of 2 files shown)


test/test_decision.py
```diff
@@ -91,6 +92,55 @@ def test_no_email_owner(self):
...
+        GitRepository,
+        "get_commit_message",
+        unittest.mock.MagicMock(return_value="Add Foo"),
+    )
+    @unittest.mock.patch.object(
+        HgRepository,
+        "get_commit_message",
+        unittest.mock.MagicMock(return_value="Add Foo"),
+    )
+    def test_default(self):
```
JohanLorenzo, line 105: Nit: Let's improve the test name with a more explicit name 🙂 

```suggestion
    def test_regular_commit_message_yields_default_target_tasks_method(self):
```

test/test_decision.py
```diff
@@ -91,6 +92,55 @@ def test_no_email_owner(self):
...
+        GitRepository,
+        "get_commit_message",
+        unittest.mock.MagicMock(return_value="DONTBUILD"),
+    )
+    @unittest.mock.patch.object(
+        HgRepository,
+        "get_commit_message",
+        unittest.mock.MagicMock(return_value="DONTBUILD"),
+    )
+    def test_dontbuild(self):
```
JohanLorenzo, line 128: Nit: Same rationale with this test message. This way, it'll be easier to future readers to understand these 2 tests are related. 

```suggestion
    def test_dontbuild_commit_message_yields_default_target_tasks_method(self):
```

JohanLorenzo (approved):
Great work! 💯  Thank you very much for improving the test coverage! I ran some manual tests locally to make sure we won't introduce a similar regression. I think we're good to go! 

I'll update the test names with more explicit names, just for the sake of future readers. Apart from that, let's go! 

If you're interested in continuing, the suggestions in https://github.com/taskcluster/taskgraph/pull/191#pullrequestreview-1313418701 are still up to date 🙂

test/test_decision.py
miladHakimi: Thanks for taking care of them.

miladHakimi:
> If you're interested in continuing, the suggestions in [#191 (review)](https://github.com/taskcluster/taskgraph/pull/191#pullrequestreview-1313418701) are still up to date 🙂

Sure, I'll take a look.

PR #202, opened 2023-03-01
unzip files in `fetch-content` quietly

bhearsum:
We already unzip tar files quietly, we probably ought to do the same for zips. It's been noted that this is very spammy in logs.

### DIFF
--- a/src/taskgraph/run-task/fetch-content
+++ b/src/taskgraph/run-task/fetch-content
@@ -374,7 +374,7 @@ def extract_archive(path, dest_dir, typ):
     elif typ == "zip":
         # unzip from stdin has wonky behavior. We don't use a pipe for it.
         ifh = open(os.devnull, "rb")
-        args = ["unzip", "-o", str(path)]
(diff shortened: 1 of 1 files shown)


hneiva (approved):

ahal (approved):
I want to double approve this.

bhearsum:
We have no test coverage for `fetch-content` AFAIK; I'm ignoring that error.

ahal:
@bhearsum for future reference, I did set up something:
https://github.com/taskcluster/taskgraph/blob/main/test/test_scripts_fetch_content.py

But it's just a single test that covers hardly anything, so no worries on skipping.. I'd have done the same in this case.

bhearsum:
> @bhearsum for future reference, I did set up something: https://github.com/taskcluster/taskgraph/blob/main/test/test_scripts_fetch_content.py
> 
> But it's just a single test that covers hardly anything, so no worries on skipping.. I'd have done the same in this case.

Ah, sorry I missed that! I clearly didn't squint hard enough. Noted for the future.

PR #207, opened 2023-03-09
refactor!: drop support for Python 3.6

ahal:
BREAKING CHANGE: Python 3.6 is no longer supported.

PR #208, opened 2023-03-09
Replace most of our Python lint tools with ruff

ahal:
This depends on the PR to drop support for Python 3.6 because it doesn't have 3.6 as an option in its settings.


jcristau (approved):
Cool :)

pyproject.toml
```diff
@@ -7,10 +7,26 @@ extend-exclude = """(\
...
+
+[tool.ruff]
+select = [
+    "E",    # pycodestyle
+    "F",    # pyflakes
+    "I",    # isort
+    "PLC",  # pylint convention
+    "PLE",  # pylint errors
+    "UP",   # pyupgrade
+]
```
hneiva, line 23: I suggest adding `"W"` as well
https://beta.ruff.rs/docs/rules/#warning-w

src/taskgraph/util/path.py
```diff
@@ -93,13 +93,8 @@ def basedir(path, bases):
...
 re_cache = {}
-# Python versions < 3.7 return r'\/' for re.escape('/').
-if re.escape("/") == "/":
-    MATCH_STAR_STAR_RE = re.compile(r"(^|/)\\\*\\\*/")
-    MATCH_STAR_STAR_END_RE = re.compile(r"(^|/)\\\*\\\*$")
-else:
-    MATCH_STAR_STAR_RE = re.compile(r"(^|\\\/)\\\*\\\*\\\/")
-    MATCH_STAR_STAR_END_RE = re.compile(r"(^|\\\/)\\\*\\\*$")
+MATCH_STAR_STAR_RE = re.compile(r"(^|/)\\\*\\\*/")
+MATCH_STAR_STAR_END_RE = re.compile(r"(^|/)\\\*\\\*$")
```
hneiva, line 97: Depends on #207 ?

hneiva (approved):

pyproject.toml
hneiva: Thinking about this. We should come up with a "releng standard set of rules for ruff"

src/taskgraph/util/path.py
ahal: Yep, sorry I was planning to rebase before requesting review.

pyproject.toml
ahal: Good call on adding "W", codebase already passes everything there.

PR #212, opened 2023-03-16
refactor!: rename '_target_task' decorator 'register_target_task'

ahal:
This decorator should be considered a public interface as many projects use it to add their own custom target task methods. Therefore it shouldn't have an underscore prefix. While we're at it, also add `register` to better describe what it is doing.

BREAKING CHANGE: rename `_target_task` to `register_target_task`


ahal:
Since we have a pending major version bump due to dropping Python 3.6 anyway, figured might as well improve this.

ahal:
I think I'd prefer to just do a hard cutover tbh, it's an easy enough change for consumers to make and next release is going to be major anyway.

ahal:
Doh, missed the major bump. Oh well. Think I'll just leave this hanging around until the next one.

jcristau (approved):

PR #217, opened 2023-03-28
Add smarter defaults for loaders, transforms, and treeherders

### LINKED ISSUE #184: Provide more defaults for kinds
One of the bigger barriers to using taskgraph is how intimidating it is to write configs. One particular part of this is how much there often is in a `kind` that doesn't directly relate to the task being written, or is essentially boilerplate that most kinds need. As a concrete example, this is a current `kind` definition used by MozillaVPN:
```
loader: taskgraph.loader.transform:loader

transforms:
    - taskgraph.transforms.job:transforms
    - taskgraph.transforms.task:transforms

tasks:
    taskgraph-definition:
        worker-type: b-linux
        worker:
            docker-image: {in-tree: base}
            max-run-time: 3600
        description: "Test the full `mozilla_vpn_taskgraph` to validate the latest changes"
        treeherder:
            symbol: test-taskgraph-definition
            kind: test
            platform: tests/opt
            tier: 1
        run:
            using: run-task
            use-caches: true
            cwd: '{checkout}'
            command: >-
                pip3 install -r taskcluster/requirements.txt &&
                taskgraph full --p taskcluster/test/params &&
                taskgraph full
```

The entirity of the `loader` and `transforms` keys are boilerplate, as well as much of the `run` section. Many other parts would be unnecessary if we provided better defaults in the `job` and/or `task` transforms. In the end, I think it's possible to shrink the above down to:
```
tasks:
    taskgraph-definition:
        worker:
            docker-image: {in-tree: base}
        description: "Test the full `mozilla_vpn_taskgraph` to validate the latest changes"
        run:
            command: >-
                pip3 install -r taskcluster/requirements.txt &&
                taskgraph full --p taskcluster/test/params &&
                taskgraph full

```

...at which point, the entire `kind` is something that I think we could reasonable expect a project owner to understand and write.

Concretely, I suggest we make the following changes:
* Use the `transform` loader as the default loader
* Provide the `job` and `task` transforms as the default transforms for the `transform` loader.
* Provide smart defaults for the entirety of the `treeherder` section:
  * Use the capitalized first letter of the `kind` for `symbol` (if multiple kinds share the same first letter, take the first two -- but after that make it required).
  * `kind` can obviously be trivially filled out
  * `tier` can default to 1
  * `platform` can default to the job/task name within the `kind`
* Parts of `worker` can have defaults as follows:
  * `chain-of-trust` defaults to `true`
  * `taskcluster-proxy` defaults to `true` (unless this is a security risk for a reason I'm unaware of)
  * `max-run-time` can default to some value (I'll throw out `1800`, but I have no real strong opinion)
  * `artifacts` should use the more or less standard entry of
```
        artifacts:
            - type: directory
              name: public/build
              path: /builds/worker/artifacts
```
* `run` can also provide a number of defaults
  * `using` can be `run-task`
  * `use-caches` can be `true`
  * `cwd` can be `{checkout}`
  * `worker-type` can be `b-linux`
    * I could be convinced this is a bad idea - but I _think_ it's true that unless you're building for Windows or macOS, your task is probably running on Linux.

Because these changes will be making many things implicit where they used to be explicit, we _must_ (as in, do not land without this) update the docs with better reference information if/when we do this. The goal here is to provide very sensible _defaults_ to lower the learning curve to adopting taskgraph (and by extension, Taskcluster) -- we do _not_ want to make it more difficult to use a more sophisticated configuration as projects outgrow early simplicity.

bhearsum:
This is a partial fix #184, dealing with the idea of a default loader, transforms, and autogenerated treeherder symbols. The individual commits have commentary on each of these - but it's largely in line with what's described in the issue.

### DIFF
--- a/src/taskgraph/generator.py
+++ b/src/taskgraph/generator.py
@@ -42,7 +42,7 @@ class Kind:
         try:
             loader = self.config["loader"]
         except KeyError:
-            raise KeyError(f"{self.path!r} does not define `loader`")
+            loader = "taskgraph.loader.default:loader"
         return find_object(loader)
 
     def load_tasks(self, parameters, loaded_tasks, write_artifacts):
--- a/src/taskgraph/util/treeherder.py
+++ b/src/taskgraph/util/treeherder.py
@@ -62,3 +62,23 @@ def inherit_treeherder_from_dep(job, dep_job):
     # Does not set symbol
     treeherder.setdefault("kind", "build")
     return treeherder
+
+
+def treeherder_defaults(kind, label):
+    defaults = {
+        # Despite its name, this is expected to be a platform+collection
+        "platform": "default/opt",
+        "tier": 1,
+    }
+    if "build" in kind:
+        defaults["kind"] = "build"
+    elif "test" in kind:
+        defaults["kind"] = "test"
+    else:
+        defaults["kind"] = "other"
+
+    # Takes the uppercased first letter of each part of the kind name, eg:
+    # apple-banana -> AB
+    defaults["symbol"] = "".join([c[0] for c in kind.split("-")]).upper()
+
+    return defaults
--- a/template/{{cookiecutter.project_name}}/taskcluster/ci/hello/kind.yml
+++ b/template/{{cookiecutter.project_name}}/taskcluster/ci/hello/kind.yml
@@ -1,10 +1,6 @@
 ---
-loader: taskgraph.loader.transform:loader
-
 transforms:
   - {{cookiecutter.project_slug}}_taskgraph.transforms.hello:transforms
-  - taskgraph.transforms.job:transforms
-  - taskgraph.transforms.task:transforms
 
 task-defaults:
   worker-type: t-linux-large
--- a/docs/concepts/kind.rst
+++ b/docs/concepts/kind.rst
@@ -30,7 +30,10 @@ look at a simple build kind example, you'll notice at the top the `loader` and
 `transforms` are defined. These point to standard taskgraph, but if you wanted
 to use a custom loader or transform it would be defined here pointing to
 `your_project_name_taskgraph` directory, eg
-`my_project_name_taskgraph.transforms.job`.
+`my_project_name_taskgraph.transforms.job`. If no loader is specified,
+`taskgraph.loader.default:loader` will be used, which will bring with it
+`taskgraph.transforms.job` and `taskgraph.transforms.task` as default
+transforms.
 
 This can also be a clue of where to look if you are trying to understand what a
 specific yaml attribute does in an existing kind file (keeping in mind that
--- a/docs/concepts/loading.rst
+++ b/docs/concepts/loading.rst
@@ -40,6 +40,14 @@ transform - whatever it expects.
 Available Loaders
 -----------------
 
-Taskgraph provides only a single loader, the :mod:`Transform Loader
-<taskgraph.loader.transform>`. Projects may optionally provide their own loaders
-that support different behavior.
+Taskgraph provides two loaders, although in truth they are virtually identical:
+
+The :mod:`Transform Loader<taskgraph.loader.transform>` is a barebones loader.
+
+The :mod:`Default Loader<taskgraph.loader.default>` does everything that the
+Transform Loader does, and additionally ensures that the
+:mod:`Job<taskgraph.transforms.job>` and :mod:`Task<taskgraph.transforms.task>`
+transforms are used on every task it loads.
+
+Projects may optionally provide their own loaders that support different
+behavior.
--- a/docs/tutorials/creating-a-task-graph.rst
+++ b/docs/tutorials/creating-a-task-graph.rst
@@ -83,26 +83,17 @@ starting definitions of the tasks themselves. If you followed the layout above,
 you have a ``hello`` kind. For this next section we'll be editing
 ``taskcluster/ci/hello/kind.yml``.
 
-#. First declare a loader. Loaders determine how the task definitions get read.
-   The most common is the :func:`transform loader
-   <taskgraph.loader.transform.loader>`:
-
-   .. code-block:: yaml
-
-    loader: taskgraph.loader.transform:loader
-
-#. Next declare the set of :term:`transforms <transform>` that will be applied
-   to tasks. Usually there is at least a kind specific set of transforms, as
-   well as the general purpose :mod:`-taskgraph.transforms.task` transforms.
-   Practically every task should use the latter, as they perform the final
-   steps to modify the tasks into the `format Taskcluster expects`_. In our
-   example:
+#. Declare the set of :term:`transforms <transform>` that will be applied
+   to tasks. By default, taskgraph will include the
+   :mod:`-taskgraph.transforms.job` and :mod:`-taskgraph.transforms.tasks`
+   transforms, which are used by the vast majority of simple tasks. It is also
+   quite common for kind-specific transforms to be used, which we will do here
+   for the purpose of demonstration. In our example:
 
    .. code-block:: yaml
 
     transforms:
         - myrepo_taskgraph.transforms.hello:transforms
-        - taskgraph.transforms.task:transforms
 
 #. Finally we define the task under the ``tasks`` key. The format for the
    initial definition here can vary wildly from one kind to another, it all
@@ -123,10 +114,8 @@ Here is the combined ``kind.yml`` file:
 
 .. code-block:: yaml
 
- loader: taskgraph.loader.transform:loader
  transforms:
      - myrepo_taskgraph.transforms.hello:transforms
-     - taskgraph.transforms.task:transforms
  tasks:
      taskcluster:
          description: "Says hello to Taskcluster"
--- a/src/taskgraph/loader/default.py
+++ b/src/taskgraph/loader/default.py
@@ -1,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+import logging
+
+from .transform import loader as transform_loader
+
+logger = logging.getLogger(__name__)
+
+
+DEFAULT_TRANSFORMS = [
+    "taskgraph.transforms.job:transforms",
+    "taskgraph.transforms.task:transforms",
+]
+
+
+def loader(kind, path, config, params, loaded_tasks):
+    """
+    This default loader builds on the `transform` loader by providing sensible
+    default transforms that the majority of simple tasks will need.
+    Specifically, `job` and `task` transforms will be appended to the end of the
+    list of transforms in the kind being loaded.
+    """
+    transform_refs = config.setdefault("transforms", [])
+    for t in DEFAULT_TRANSFORMS:
+        if t in config.get("transforms", ()):
+            raise KeyError(
+                f"Transform {t} is already present in the loader's default transforms; it must not be defined in the kind"
+            )
+    transform_refs.extend(DEFAULT_TRANSFORMS)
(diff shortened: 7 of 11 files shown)


src/taskgraph/generator.py
```diff
@@ -31,22 +31,47 @@ class KindNotFound(Exception):
...
+            # specified if they are going to be added automatically. This
+            # avoids issues such as including them twice, being in a different
+            # order in the kind vs. here, etc.
+            for t in DEFAULT_TRANSFORMS:
+                if t in transform_refs:
+                    raise KeyError(
+                        "neither job nor task transforms may be specified unless you also specify a loader"
+                    )
+            loader = "taskgraph.loader.transform:loader"
+            transform_refs.extend(DEFAULT_TRANSFORMS)
```
jcristau, line 65: It seems a bit surprising to me that we add the default transforms if you don't specify a loader in kind.yml, but not if you set it, even to the default.  I guess it sort of makes sense because it lets you keep backwards compat with existing kind.yml files, but conceptually it feels odd.

src/taskgraph/util/treeherder.py
```diff
@@ -62,3 +62,30 @@ def inherit_treeherder_from_dep(job, dep_job):
     # Does not set symbol
     treeherder.setdefault("kind", "build")
     return treeherder
+
+
+def treeherder_defaults(kind, label):
+    default_collection = "opt"
+    label_without_kind = label.split(f"{kind}-", 1)[1]
```
jcristau, line 69: Are we sure that the label starts with `{kind}-`, or could we end up with an IndexError here?

src/taskgraph/util/treeherder.py
```diff
@@ -62,3 +62,30 @@ def inherit_treeherder_from_dep(job, dep_job):
...
+    if "build" in kind:
+        defaults["kind"] = "build"
+    elif "test" in kind:
+        defaults["kind"] = "test"
+    else:
+        defaults["kind"] = "other"
+
+    # Takes the uppercased first letter of each part of the kind name, eg:
+    # apple-banana -> AB
+    defaults["symbol"] = "".join([c[0].upper() for c in kind.split("-")])
```
jcristau, line 89: nit: could call upper() just once at the end

jcristau (commented):
I guess you'll want to update the docs for this as well?

src/taskgraph/generator.py
bhearsum: I agree with you, actually, and I'm open to change here. A couple of things I experimented with were:
* A new loader called `default` that has these default transforms set. (The actual implementation just called out to the `transforms` loader - so it was really just a way to uniquely identify the "default" case - otherwise it functionally identical to the `transform` loader.) The main reason I didn't end up with this is because it felt like overcomplicating things.
* Adding the concept of default transforms to loaders in general, by putting that list in, eg: `loader/transform.py`. I didn't pursue this in the end because I wasn't sure how to handle the case of conflicts between the `kind` and the defaults in the loaders. (Perhaps this is the same thing we're discussing about the current patch though...)

src/taskgraph/util/treeherder.py
bhearsum: Good question, I will try to verify this assumption.

bhearsum:
> I guess you'll want to update the docs for this as well?

Good callout - indeed!

src/taskgraph/generator.py
```diff
@@ -31,22 +31,47 @@ class KindNotFound(Exception):
...
-            raise KeyError(f"{self.path!r} does not define `loader`")
-        return find_object(loader)
+            # To avoid confusion, we do not allow these transforms to be
+            # specified if they are going to be added automatically. This
+            # avoids issues such as including them twice, being in a different
+            # order in the kind vs. here, etc.
+            for t in DEFAULT_TRANSFORMS:
+                if t in transform_refs:
+                    raise KeyError(
+                        "neither job nor task transforms may be specified unless you also specify a loader"
```
ahal, line 62: Maybe change this to `f"Default transforms ({t}) cannot be specified unless..."`. That way we don't hardcode the names of the default transforms.

src/taskgraph/generator.py
ahal: Tbh I like the `default` loader idea the best and feel like it's less complicated than the current implementation. The main reason is that the current implementation is making the false assumption that loaders have transforms. But in reality, one could implement a loader that doesn't use transforms at all (I believe avoiding transforms was even the original intent of loaders)! Personally I feel like loaders are too complicated and should be removed, but as long as we have them, I'd prefer to try and avoid assuming they are transform based.

src/taskgraph/generator.py
```diff
@@ -31,22 +31,47 @@ class KindNotFound(Exception):
...
+                    )
+            loader = "taskgraph.loader.transform:loader"
+            transform_refs.extend(DEFAULT_TRANSFORMS)
+
+        transforms = TransformSequence()
+        for xform_path in transform_refs:
+            transform = find_object(xform_path)
+            transforms.add(transform)
+        return find_object(loader), transforms
```
ahal, line 71: See comment above, loaders don't necessarily have to use `TransformSequence` (though so far they all do).

src/taskgraph/util/treeherder.py
ahal: Labels don't have to start with `{kind}-`, that's only a convention. I am positive we have several places that break this convention.

src/taskgraph/generator.py
```diff
@@ -31,22 +31,47 @@ class KindNotFound(Exception):
     """
 
 
+# In an ideal world these would be part of each loader. Unfortunately, loaders
+# can only provide information as part of yielding task inputs, and we need to
+# know the set of transforms prior to starting to read from the loader.
+DEFAULT_TRANSFORMS = [
+    "taskgraph.transforms.job:transforms",
+    "taskgraph.transforms.task:transforms",
```
ahal, line 39: Maybe this could go in the graph config instead?

ahal (changes requested):
Thanks Ben!

I really like this cleanup and think it will be a positive improvement. That said, I can't help but feel that this is being hampered by trying to preserve backwards compatibility with things like the `multi_dep` loader and the like. Iterative improvement is great and something we should strive for.. But I still can't help but feel like maybe it would be better to be more aggressive here and remove loaders entirely.

Basically, convert the `multi_dep`, `single_dep` and `test` (from Gecko) loaders into transform sequences. Then everything uses `transform` loader and all they need to do is specify which transforms to use (which may or may not include defaults). This will be more work, but would get us closer to an ideal end state imo. If we don't have the will or time to take this on now, then let's just stick with iterative improvement!

src/taskgraph/generator.py
bhearsum: I'm not even sure how you get away without at least the `task` transform in this day and age :upside_down_face: - but you're right, this is definitely an assumption that isn't necessarily true.

I'll take another pass on the `default` idea.

src/taskgraph/generator.py
bhearsum: I do like the idea of a project being able to override or append to this - but especially if we're going with the `default` transform idea it feels like this belongs in taskgraph to make it more "batteries included".

src/taskgraph/generator.py
bhearsum: Something I noticed on a read -- it's actually the case right now that [we require them to have at least an empty array of transforms defined](https://github.com/taskcluster/taskgraph/blob/3018cf24e697b60ad067c276ac03ee046c63f6b8/src/taskgraph/generator.py#L59) :grimacing:.

And one more clarification: for the new, `default` loader, we clearly must assume that it uses transforms in order to set these defaults -- right? Is there a preference between leaving them in `generator.py` as they are now, or tossing them in the new `taskgraph.loader.default` module? (The latter may imply that all loaders support default transforms...which it sounds like we don't want to do?)

ahal: TIL! I feel less strongly about introducing further assumptions then.

I think they'd be better in `taskgraph.loader.default` because that is what is consuming it. To me `generator.py` feels like the wrong place.

bhearsum: Yeah, I agree that it feels like the better place for it. Given that though - should be looking in _all_ loaders for `DEFAULT_TRANSFORMS`, then? Or is it a special behaviour only for the `default` one? (Presumably we'd look for it regardless of whether the `default` loader was explicitly specified or not, too.)

ahal: Oh I see, I guess I was thinking the implementation logic could also go in `loader/default.py`, but I guess it needs to be in `generator.py` for some reason?

I guess it makes sense to look for it in all loaders in that case? I dunno.

bhearsum: Right -- it can't go in the `loader` function because that function doesn't yield anything until transforms begin yielding. We need to know which transforms are being used prior to running any existing loading code.

We could certainly add a new function that we look for in a loader module that returns default transforms, but anything we do that puts this proposed set of defaults into `loader/default.py` sortof implies that it's something that could be used by any loader.

bhearsum: ([Our call to the loader function](https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/generator.py#L57 return) returns a generator, and the yielding doesn't happen until [we start running transforms](https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/generator.py#L86). I don't _think_ there's a way to have it, eg: return a tuple of generator + transforms, although perhaps that's wrong.)

bhearsum:
I've pushed a few changes which I _think_ address all of the comments (as long as we're all happy with how the default loader is being handled), specifically:
- Doc updates
- Re-add `default` loader, and move `DEFAULT_TRANSFORMS` into it
- Add support for `DEFAULT_TRANSFORMS` for all loaders (I could be convinced not to do this, but it would feel a bit wrong to special case the `default` loader for it...)
- Default treeherder.platform to `default/opt` :disappointed:. Open to other ideas here, but there don't seem to many options here.

src/taskgraph/generator.py
jcristau: The loader function could e.g. modify `config["transforms"]` and then return the generator, instead of yielding things directly (as in https://github.com/jcristau/taskgraph/commit/2c420c06738c4fd30040054c5531cbb5a5ad85b4)?
Or just return a generator + transforms, indeed (as in https://github.com/jcristau/taskgraph/commit/6da7c2123ea13b6017d4772f7f5b64d048b28f78).
(I'm not saying that's necessarily better, e.g. I'm not a fan of the side effect in the first patch, or the backwards compat handling in the second one, but if we want to keep the logic to the default loader instead of the generator, it's a possibility)

jcristau (commented):
Looks good to me; thanks!

bhearsum:
@jcristau - thanks again for those better ideas. I've applied https://github.com/jcristau/taskgraph/commit/2c420c06738c4fd30040054c5531cbb5a5ad85b4 to this, as it seems like the most sensible interface (different return types for different loaders feels quite awkward...).

template/{{cookiecutter.project_name}}/taskcluster/ci/hello/kind.yml
```diff
@@ -1,10 +1,6 @@
 ---
-loader: taskgraph.loader.transform:loader
-
 transforms:
   - {{cookiecutter.project_slug}}_taskgraph.transforms.hello:transforms
-  - taskgraph.transforms.job:transforms
-  - taskgraph.transforms.task:transforms
```
ahal, line 7: The only downside to this is that things aren't as transparent anymore. It won't be obvious that the `job` then the `task` transforms are applied in that order.

I think longer term we'll want to combine these into a single `TransformSequence`, but let's leave that for a follow-up.

ahal (approved):
Thanks, this is great!

jcristau (approved):

PR #219, opened 2023-03-30
Bug 1805651 - Add scope-summaries support for config.yml

gbrownmozilla:


gbrownmozilla:
Is config.yml a good place for this? Is "scope-summaries" a good key name?

JohanLorenzo (approved):
Cool! Looks good to me 😃👍

JohanLorenzo:
> Is config.yml a good place for this? Is "scope-summaries" a good key name?

I think it makes sense, yeah. Looping in @jcristau who had an opinion on this topic in https://github.com/taskcluster/taskgraph/pull/173.

gbrownmozilla:
> I'd maybe make the name be just about index paths rather than scopes, and remove the `index:insert-path:` prefix so we can just compare the paths. That seems like it's easier to understand and better reflects what it's doing. WDYT?

I didn't like that idea at first, but I think you've won me over.

Now I'm thinking it's also time to remove _SCOPE_SUMMARY_REGEXPS: The existing regexes are unused and any new ones can be made via config.yml.

src/taskgraph/morph.py
```diff
@@ -139,18 +127,20 @@ def make_index_task(parent_task, taskgraph, label_to_taskid, parameters, graph_c
     # temporary credential.
     scopes = set()
     domain_scope_regex = re.compile(
```
jcristau, line 129: rename to `domain_index_regex`?

jcristau (approved):
LGTM!

PR #224, opened 2023-04-13
feat: use 'transforms' as object by default when resolving references

ahal:
This is a minor quality of life improvement to simplify kind.yml files a small amount.


ahal:
This seemed like a simple little fix that could work along side your other config cleanups.

bhearsum (approved):
Agreed, this is a nice discrete improvement!

PR #226, opened 2023-04-17
feat: implement a 'run-task-command' key in run_task.py transforms

ahal:
The current Python path to Mac hardcodes Python 3.6 into the name, and on Windows it assumes `mozilla-build` is being used!

While we can (and should) update this to a better location (I see Gecko moved to `/usr/local/bin/python3`), forcing hardcoded Python path at all is somewhat problematic as it makes assumptions about the workers that are not necessarily going to be valid.

I think ideally we would just run `./run-task` all the time and it should just work, maybe the shebang could be changed to `/usr/bin/env python3`? However I'm a bit afraid that doing this is going to break certain workers.

So this implements a method to override the command used to invoke `run-task` from a task definition. This way if we break things, there will be a viable workaround ready to go.

I'm not super happy with this approach, so maybe it should be removed again once we have a more standard story around invoking `run-task`.

### DIFF
--- a/src/taskgraph/transforms/job/run_task.py
+++ b/src/taskgraph/transforms/job/run_task.py
@@ -59,6 +59,9 @@ run_task_schema = Schema(
         },
         # What to execute the command with in the event command is a string.
         Optional("exec-with"): Any(*list(EXEC_COMMANDS)),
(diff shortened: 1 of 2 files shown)


jcristau (approved):
Seems like an improvement at least; thanks.

PR #231, opened 2023-04-24
run-task: abort git clone/fetch if http transfer is too slow

jcristau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1829604


gbrownmozilla (approved):
I had never heard of these. Nice find.

jcristau:
> I had never heard of these. Nice find.

Me neither tbh :)

PR #233, opened 2023-04-25
Rework treeherder defaults to fix indexes that get dropped sometimes.

bhearsum:
It turns out that `build_task` needs to run after `add_index_routes` (big whoops on my part...).

Fixing this means extracting all the treeherder goop from `build_task` so we can run it before `add_index_routes`. In the process I noticed that my original patch ended up adding unnecessary `kind` and `platform` keys to the `extra.treeherder` object, which I'm also correcting here. (This was due to using the `treeherder` object as both an intermediary object to merge the defaults and task provided treeherder information and the final thing we set on the task - something we weren't doing before.)

### DIFF
--- a/src/taskgraph/transforms/task.py
+++ b/src/taskgraph/transforms/task.py
@@ -932,42 +932,30 @@ def add_generic_index_routes(config, task):
 
 
 @transforms.add
-def build_task(config, tasks):
+def process_treeherder_metadata(config, tasks):
     for task in tasks:
-        level = str(config.params["level"])
-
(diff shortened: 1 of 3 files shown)


jcristau:
How painful would it be to add a testcase for this?

jcristau:
> > How painful would it be to add a testcase for this?
> 
> Probably ought to give it a try, at least.
> 
> I suppose you specifically mean a test case that we aren't dropping the index routes that taskgraph itself provides?

Something like that, yeah; I didn't really think it through though.

Issue #43, opened 2022-05-05
Print a nice error message when using `taskgraph {load|build}-image` without docker started


ahal:
Currently you get the following very confusing output:
```
$ taskgraph load-image --task-id=ROzeMiL1Rv-5A_uIhJr4UA
Traceback (most recent call last):
  File "/home/ahal/.pyenv/versions/taskgraph/lib/python3.7/site-packages/requests/adapters.py", line 470, in send
    low_conn.endheaders()
  File "/home/ahal/.pyenv/versions/3.7.12/lib/python3.7/http/client.py", line 1276, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/home/ahal/.pyenv/versions/3.7.12/lib/python3.7/http/client.py", line 1036, in _send_output
    self.send(msg)
  File "/home/ahal/.pyenv/versions/3.7.12/lib/python3.7/http/client.py", line 976, in send
    self.connect()
  File "/home/ahal/.pyenv/versions/taskgraph/lib/python3.7/site-packages/requests_unixsocket/adapters.py", line 41, in connect
    sock.connect(socket_path)
FileNotFoundError: [Errno 2] No such file or directory

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ahal/dev/taskgraph/src/taskgraph/main.py", line 544, in load_image
    ok = load_image_by_task_id(args["task_id"], args.get("tag"))
  File "/home/ahal/dev/taskgraph/src/taskgraph/docker.py", line 61, in load_image_by_task_id
    result = load_image(artifact_url, tag)
  File "/home/ahal/dev/taskgraph/src/taskgraph/docker.py", line 198, in load_image
    docker.post_to_docker(download_and_modify_image(), "/images/load", quiet=0)
  File "/home/ahal/dev/taskgraph/src/taskgraph/util/docker.py", line 49, in post_to_docker
    headers={"Content-Type": "application/x-tar"},
  File "/home/ahal/.pyenv/versions/taskgraph/lib/python3.7/site-packages/requests/sessions.py", line 577, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/home/ahal/.pyenv/versions/taskgraph/lib/python3.7/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
[...]

PR #237, opened 2023-05-03
Print a nice error message when using taskgraph {load|build}-image without docker started #43

pawanraocse:
Hi - I am new to open source contribution and this is my first commit, please review and share some feedback.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -508,7 +508,7 @@ def show_taskgraph(options):
 def build_image(args):
     from taskgraph.docker import build_context, build_image
 
-    validateDocker()
+    validate_docker()
     if args["context_only"] is None:
         build_image(args["image_name"], args["tag"], os.environ)
     else:
@@ -545,7 +545,7 @@ def load_image(args):
     if not args.get("image_name") and not args.get("task_id"):
         print("Specify either IMAGE-NAME or TASK-ID")
         sys.exit(1)
-    validateDocker()
+    validate_docker()
     try:
         if args["task_id"]:
             ok = load_image_by_task_id(args["task_id"], args.get("tag"))
@@ -558,7 +558,7 @@ def load_image(args):
         sys.exit(1)
 
 
-def validateDocker():
+def validate_docker():
     import docker
 
     try:
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -559,13 +559,15 @@ def load_image(args):
 
 
 def validate_docker():
-    import docker
-
     try:
-        client = docker.from_env()
-        containers = client.containers.list()
-        print(containers)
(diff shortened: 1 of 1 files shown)


ahal (changes requested):
Nice, thanks for the patch!

This is almost exactly what I was looking for, except that it looks like you're using some third party `docker` library that won't be installed on user's systems. We could add it as a dependency, but it doesn't feel worth depending on a new package just for this.

So instead, I propose you use the [subprocess](https://docs.python.org/3/library/subprocess.html) module to run something like `docker system info`. Then you can print this message if the return code is non-zero.

Let me know if you have any questions!

pawanraocse:
Thanks for the review. Make sense to remove the library. I have done the changes, please review it again.

src/taskgraph/main.py
```diff
@@ -556,6 +558,22 @@ def load_image(args):
...
+                "Error: Cannot connect to the Docker daemon. Is the docker daemon running?"
+            )
+            sys.exit(1)
+    except Exception:
+        traceback.print_exc()
+        print(
+            "Error: Cannot connect to the Docker daemon. Is the docker daemon running?"
+        )
+        sys.exit(1)
+
```
ahal, line 575: There shouldn't be any need to catch `Exception` here. By default the `subprocess.run` function has `check=False`, which means it won't raise an exception if the command fails. If we *do* somehow get an exception here, it's almost certainly not due to docker being stopped, so we should let it propagate.

src/taskgraph/main.py
```diff
@@ -556,6 +558,22 @@ def load_image(args):
         sys.exit(1)
 
 
+def validate_docker():
+    try:
+        p = subprocess.run(["docker", "ps"])
+        if p.returncode != 0:
+            print(
+                "Error: Cannot connect to the Docker daemon. Is the docker daemon running?"
```
ahal, line 566: I know I told you to add this print, but by default, the `subprocess.run` function will print the output from the command. On my system at least, `docker ps` prints out:

> Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?

So we'd be esssentially printing the same message twice. Maybe what you could do is add `capture_output=True` to the `run` call, then instead of printing this message, you would print something like:
```
print("Error connecting to Docker:\n" + proc.stderr)
```

ahal (changes requested):
Thanks, we're almost there. I thought of two more minor issues.

src/taskgraph/main.py
pawanraocse: Sure, Added the changes. Please have a look

ahal (approved):
This looks great, thank you very much for your contribution! If you're interested in contributing more feel free to peruse the issues or reach out if you're unsure!

Issue #90, opened 2022-08-09
artifact-reference should support non-public artifacts


escapewindow:
https://github.com/mozilla-extensions/xpi-template/issues/37
Essentially, we probably want to remove [this check](https://github.com/taskcluster/taskgraph/blob/ec733aa2c331390a0af079fa4a71e47c4622468f/src/taskgraph/util/parameterization.py#L82-L87).

Given that:

a) we can use artifact-reference for scriptworkers, which often do have private artifact scopes baked in (until we resolve https://github.com/mozilla-releng/scriptworker/issues/426, after which we'll need to grant the scope to the task as normal), and
b) if we grant the given task the proper private artifact scope and enable the taskcluster-proxy, non-scriptworker tasks can also download private artifacts,

I don't see a need for this artificial restriction.

jcristau:
Fixed by #485

PR #238, opened 2023-05-10
Allow artifact-reference to support non-public artifacts

willdurand:
Fixes https://github.com/taskcluster/taskgraph/issues/90

(assuming I got it right that we don't need more than that...)

### DIFF
--- a/src/taskgraph/util/parameterization.py
+++ b/src/taskgraph/util/parameterization.py
@@ -79,11 +79,6 @@ def resolve_task_references(label, task_def, task_id, decision_task_id, dependen
                         )
                     )
 
-            assert artifact_name.startswith(
-                "public/"
-            ), "artifact-reference only supports public artifacts, not `{}`".format(
-                artifact_name
-            )
             return get_artifact_url(task_id, artifact_name)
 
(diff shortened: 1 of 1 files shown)


jcristau:
@willdurand taking another look at this, we might be able to change `get_artifact_url` to not use a bewit URL, and download through the taskcluster proxy directly, since the comment there points at https://bugzilla.mozilla.org/show_bug.cgi?id=1405889 as the reason for doing that dance, and that was resolved years ago.

jcristau:
The other thing is that `resolve_task_references` calls `get_artifact_url` without enabling the proxy, which would need to change for private artifacts.

jcristau (changes requested):
Moving this out of review queue, see previous comments.

petemoore:
Related: https://github.com/taskcluster/taskgraph/issues/407

ahal:
I ran into this limitation myself, going to take a poke at fixing it.

> The other thing is that resolve_task_references calls get_artifact_url without enabling the proxy, which would need to change for private artifacts.

Does it? I think we can remove `use_proxy` entirely from `get_artifact_url` now.

ahal:
Subsumed by #485

PR #240, opened 2023-05-14
Updated check_task_dependencies to also check soft-dependencies and if-dependencies and added unit test to test it

ilhamalrahm:
Fixed check_task_dependencies to also check soft-dependencies and if-dependencies and added unit tests to test the fix.
This resolves #74
This now also resolves #241

PR #243, opened 2023-05-16
fix: add the python virtual environment to .gitignore

sewelseb:
a simple fix to add the python virtual environement to the .gitignore

@JohanLorenzo can you review it? :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,7 @@ __pycache__/
 .tox/
 
 #python virtual environement
-taskgraph/
+/taskgraph/
 
 # Coverage
 .coverage
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,9 @@ __pycache__/
 .ruff_cache/
 .tox/
 
+#python virtual environement
+taskgraph/
+
 # Coverage
 .coverage
 


.gitignore
```diff
@@ -17,6 +17,9 @@ __pycache__/
 .ruff_cache/
 .tox/
 
+#python virtual environement
+taskgraph/
```
JohanLorenzo, line 21: Nit: Let's make git ignore just the `taskgraph` folder at the root directory. The current implementation excludes any files in `src/taskgraph` which will eventually be an issue. This can easily be fixed by:

```suggestion
/taskgraph/
```

Source[1]: 

> The pattern `hello.*` matches any file or directory whose name begins with `hello.`. If one wants to restrict this only to the directory and not in its subdirectories, one can prepend the pattern with a slash, i.e. `/hello.*`; the pattern now matches `hello.txt`, `hello.c` but not `a/hello.java`.


[1] https://git-scm.com/docs/gitignore#_examples

JohanLorenzo (commented):
Thank you for making this change @sewelseb 😃  It turns out another first-time contributor faced the same issue this week: https://github.com/taskcluster/taskgraph/pull/240#discussion_r1193538958. I agree with the intent to fix this experience. That said, let's make a small change in order not to surprise future contributors. I left more details below.

JohanLorenzo (approved):
Sweet! It looks good to me 🙂

Issue #7, opened 2022-04-13
Dump Python and other runtime info in run-task


ahal:
Imported from: https://bugzilla.mozilla.org/show_bug.cgi?id=1696947

It's often handy to know what version of Python a task is running against. However this info is usually missing from the logs. While it would be tricky to tell which version of Python a command will use (e.g mach might decide to use 2 vs 3 and we have no way of telling that from the taskgraph), we could get run-task to dump the versions of all default Python executables. E.g:

$ python --version
$ python2 --version
$ python3 --version

Then we'd only have to know which of those are being used where.

JohanLorenzo:
Let's add a new line here:

https://github.com/taskcluster/taskgraph/blob/a0dd5d37b31b4912969b87bd57c14462cb0a1868/src/taskgraph/run-task/run-task#L1057-L1058

We can first start by telling which python is running `run-task`. This can be achieved directly from Python https://stackoverflow.com/a/1252164. 

Then, I agree with the first comment, we should run [`subprocess.check_output()`](https://docs.python.org/3/library/subprocess.html?highlight=subprocess#subprocess.check_output) with the 3 python commands to make sure what version are run.

I'm happy to mentor this bug and provide additional info.

sewelseb:
I am interrested to try to solve it :)

PR #245, opened 2023-05-17
Fix: dump python version number in run task

sewelseb:
This pull request is there to fix the issue:
https://github.com/taskcluster/taskgraph/issues/7

@JohanLorenzo : thanks for your help and advices. Can you review my PR? :)

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -28,6 +28,7 @@ import io
 import json
 import os
 from pathlib import Path
+import platform
 import re
 import shutil
 import signal
@@ -1054,7 +1055,9 @@ def maybe_run_resource_monitoring():
     return process
 
 def _display_pyhton_version():
-    print("Python version:")
+    print("Python versions: ")
+    print(platform.python_version())
+
 
 def main(args):
     os.environ["TASK_WORKDIR"] = os.getcwd()
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1055,7 +1055,7 @@ def maybe_run_resource_monitoring():
     return process
 
 def _display_pyhton_version():
-    print("Python versions: ")
+    print("Python version: ")
     print(platform.python_version())
 
 
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1055,10 +1055,9 @@ def maybe_run_resource_monitoring():
     return process
 
 def _display_pyhton_version():
-    pythonVersionOutput = "Python version: "+platform.python_version()
     print_line(
         b"setup",
-        bytes(pythonVersionOutput, encoding='utf8')
+        b"Python version: "+ bytes(platform.python_version(), 'utf-8')
     )
 
 
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1055,17 +1055,20 @@ def maybe_run_resource_monitoring():
     return process
 
 def _display_pyhton_version():
-    print("Python version: ")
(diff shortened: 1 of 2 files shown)


JohanLorenzo:
Nice! I'll defer the review to @ahal for faster feedback.

src/taskgraph/run-task/run-task
```diff
@@ -1053,9 +1054,14 @@ def maybe_run_resource_monitoring():
     monitor_process.start()
     return process
 
+def _display_pyhton_version():
```
ahal, line 1057: I don't think it's worth pulling this out into a new function, let's just keep it in-line in `main`.

src/taskgraph/run-task/run-task
```diff
@@ -1053,9 +1054,14 @@ def maybe_run_resource_monitoring():
...
+    print("Python version: ")
+    print(platform.python_version())
```
ahal, line 1059: Instead of `print`, you should use `print_line(b"setup", b"<msg>")`. This way it will go to the same place as the rest of the logs and look consistent with them :). I'm not really sure why we need to use byte strings, but oh well.. best to stay consistent.

src/taskgraph/run-task/run-task
```diff
@@ -1053,9 +1054,14 @@ def maybe_run_resource_monitoring():
...
+
 
 def main(args):
     os.environ["TASK_WORKDIR"] = os.getcwd()
+    _display_pyhton_version()
```
ahal, line 1064: Let's move this after the `print_line` call below. That log line signals the very start of the script, so it makes sense to me that it would be the first line in the log.

ahal (changes requested):
Thanks for the patch! Please address the review comments below and re-submit.

src/taskgraph/run-task/run-task
sewelseb: Thank you for your review. I did it on the advice of @JohanLorenzo . The objective was to isolated it for the unit tests. I can indeed send it back to the main function

src/taskgraph/run-task/run-task
sewelseb: ok, I will update that

sewelseb:
@ahal : I have fixed the two last comments, I have just kept the external function waiting for your repply.

src/taskgraph/run-task/run-task
ahal: Ah, my mistake I didn't notice it was being used by the test. That is fine then!

I do notice you have a typo in the function name though, `s/pyhton/python`. Please fix :)

ahal (changes requested):
Thanks for addressing the comment! I agree that keeping it in a separate function makes sense for testing. I have a couple more minor issues to fix, but looking good otherwise.

src/taskgraph/run-task/run-task
```diff
@@ -1053,13 +1054,21 @@ def maybe_run_resource_monitoring():
     monitor_process.start()
     return process
 
+def _display_pyhton_version():
+    pythonVersionOutput = "Python version: "+platform.python_version()
+    print_line(
+        b"setup",
+        bytes(pythonVersionOutput, encoding='utf8')
```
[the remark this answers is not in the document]
sewelseb: it doesn't work, but I have done something similar that works

src/taskgraph/run-task/run-task
sewelseb: Oups, yes, thanks, I have fixed it.

ahal (changes requested):
Sorry for the late review!

Though I still see the `pyhton` typo in several places, please fix that up :)

ahal (approved):
Thank you very much for the contribution!

PR #250, opened 2023-05-24
fix: ensure 'taskgraph' command exits non-zero if any parallel genera…

ahal:
…tions fail

This will fix a bug in a VPN task that is using this featuer to verify that all of the checked in parameters files produce valid graphs.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -189,7 +189,7 @@ def generate_taskgraph(options, parameters, logdir):
         spec = parameters[0]
         out = format_taskgraph(options, spec, logfile(spec))
         dump_output(out, options["output_file"])
-        return
+        return 0
 
     futures = {}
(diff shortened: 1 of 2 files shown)


src/taskgraph/main.py
```diff
@@ -838,7 +844,7 @@ def main(args=sys.argv[1:]):
     parser = create_parser()
     args = parser.parse_args(args)
     try:
-        args.command(vars(args))
+        return args.command(vars(args))
```
hneiva, line 847: should this be `sys.exit(args.command(...))` ?

ahal: Possibly, either that or we should be returning 1 in the error case instead of raising `SystemExit`. This function is kind of built to be called programmatically, so I'm leaning towards the latter..

That said, switching to be consistent either way would technically be backwards incompatible, so I wasn't going to bother with it (at least in this PR).

PR #251, opened 2023-05-24
Run pyright against codebase

ahal:


ahal:
This just excludes every file that has pre-existing errors. I didn't have time or motivation to fix everything, and this way we can improve things incrementally (and will at least get type checking on new files).

Once all existing errors are fixed, we can think about enforcing annotations.. but not yet.

bhearsum (approved):
Excellent!!

jcristau (approved):
Nice :)

PR #257, opened 2023-05-30
fix: move group-by logic to utility file

ahal:
This fixes a bug that results in a schema validation error anytime a custom group-by function is used. Essentially, the schema was using a snapshot of the GROUP_BY_MAP from the time it was imported. So third party taskgraph code that attempted to then add additional functions, would fail validation as the schema has already been built.

By moving this logic to an external utility module, it allows external consumers to register their custom functions prior to the schema being built.


hneiva (approved):
LGTM

PR #261, opened 2023-06-05
Fetch tags during run-task's git checkout in some scenarios

bhearsum:
This works around recent issues found when starting from an older cached checkout - more details in the comments.

@jcristau - I ended up making this pretty targeted to avoid pulling tags in unwanted scenarios. I think it covers the case we ran into today at the very least? (Which, for posterity, was a task starting from a worker that had cached a checkout before the tag existed, then run a `github-release` task that wanted to checkout the tag.)

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -654,6 +654,24 @@ def git_checkout(
 
         retry_required_command(b"vcs", args, cwd=destination_path, extra_env=env)
 
+    # If a ref was provided, it might be tag, so we need to make sure we fetch
+    # those. This is explicitly only done when base and head repo match,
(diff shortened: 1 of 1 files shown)


jcristau (commented):
Does this actually work if the tag gets moved around and we've cached an old version?
In my testing it looks like it errors out (git fetch rejects it with "would clobber existing tag").

bhearsum:
> Does this actually work if the tag gets moved around and we've cached an old version? In my testing it looks like it errors out (git fetch rejects it with "would clobber existing tag").

Hrm! I'll take a deeper look. Thanks for catching this.

bhearsum:
`--force` seems like the right thing to do here. I can't think of any downsides to using it...I think we always ought to be doing everything we can to get the requested ref?

PR #264, opened 2023-06-08
Replace attrs with dataclasses

ahal:
Now that the minimum version is 3.7, we can use dataclasses!

I'll be honest, the main motivation for this change is that Pyright doesn't support `attrs` (and they don't plan to unless attrs changes some things). But fewer dependencies and using stdlib is a nice side bonus.


gabrielBusta (approved):

ahal:
Nope, we've already added 3.6 incompatibilities anyway

PR #272, opened 2023-06-20
Fix/dump python version number in run task subtasks

sewelseb:
Hello,

here is my pull request to display the python version of the subprocess. 
it comes from the issue https://github.com/taskcluster/taskgraph/issues/7

@JohanLorenzo  : thanks for your help and advices. Can you review my PR? :)

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -570,6 +570,7 @@ def _clean_git_checkout(destination_path):
         cwd=destination_path,
         env=os.environ,
     )
+    _display_python_version_in_sub_process(subprocess)
     stdout = io.TextIOWrapper(p.stdout, encoding="latin1")
     ret = p.wait()
     if ret:
@@ -1078,6 +1079,11 @@ def _display_python_version():
         b"Python version: "+ bytes(platform.python_version(), 'utf-8')
     )
 
+def _display_python_version_in_sub_process(subprocess: subprocess):
+    print_line(
+        b"setup",
+        subprocess.check_output("python --version")
+    )
 
 def main(args):
     os.environ["TASK_WORKDIR"] = os.getcwd()
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1087,7 +1087,7 @@ def _display_python_version():
 def _display_python_version_in_sub_process(subprocess: subprocess):
     print_line(
         b"setup",
-        subprocess.check_output("python --version")
+        subprocess.check_output(["python", "--version"])
     )
 
 def main(args):
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1087,8 +1087,48 @@ def _display_python_version():
 def _display_python_version_in_sub_process(subprocess: subprocess):
     print_line(
         b"setup",
-        subprocess.check_output(["python", "--version"])
+        b"pyhton --version: ",
     )
+    try:
+        print_line(
+            b"setup",
+            subprocess.check_output(["python", "--version"])
+        )
+    except:
+         print_line(
+            b"setup",
+            b"Unknown command: 'pyhton --version'",
+        )
+         
+    print_line(
+        b"setup",
+        b"pyhton2 --version: ",
+    )
+    try:
+        print_line(
+            b"setup",
+            subprocess.check_output(["python2", "--version"])
+        )
+    except:
+        print_line(
+            b"setup",
+            b"Unknown command: 'pyhton2 --version'",
+        )
+
+    print_line(
+        b"setup",
+        b"pyhton3 --version: ",
+    )
+    try:
+        print_line(
+            b"setup",
+            subprocess.check_output(["python3", "--version"])
+        )
+    except:
+         print_line(
+            b"setup",
+            b"Unknown command: 'pyhton3 --version'",
+        )
 
 def main(args):
     os.environ["TASK_WORKDIR"] = os.getcwd()
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -275,6 +275,7 @@ def run_command(prefix, args, *, extra_env=None, cwd=None):
         cwd=cwd,
         env=env,
     )
+    _display_python_version_in_sub_process(subprocess)
 
     stdout = io.TextIOWrapper(p.stdout, encoding="latin1")
 
@@ -724,6 +725,7 @@ def git_checkout(
         args, cwd=destination_path, universal_newlines=True
     ).strip()
     assert re.match("^[a-f0-9]{40}$", commit_hash)
+    _display_python_version_in_sub_process(subprocess)
 
     if head_repo.startswith("https://github.com"):
         if head_repo.endswith("/"):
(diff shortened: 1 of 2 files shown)


jcristau:
I believe we'd want to run all of python --version, python2 --version and python3 --version, and report which of them work and what version they are

sewelseb:
@jcristau & @JohanLorenzo : I have made the update to try the 3 commands. do you think I need to also send a notification to another system if a version 2 is detected?

src/taskgraph/run-task/run-task
```diff
@@ -275,6 +275,7 @@ def run_command(prefix, args, *, extra_env=None, cwd=None):
         cwd=cwd,
         env=env,
     )
+    _display_python_version_in_sub_process(subprocess)
```
JohanLorenzo, line 278: Nit: I'm sorry, I'm not sure to follow. What's the intent about passing down `subprocess`? This module is available everywhere in this file since it's imported at the top-level: 

https://github.com/taskcluster/taskgraph/blob/4346754b8457dd53f8b541297b8d60d502c75c0c/src/taskgraph/run-task/run-task#L37

src/taskgraph/run-task/run-task
```diff
@@ -570,6 +571,7 @@ def _clean_git_checkout(destination_path):
         cwd=destination_path,
         env=os.environ,
     )
+    _display_python_version_in_sub_process(subprocess)
```
JohanLorenzo, line 574: @jcristau, correct me if I'm wrong: the python binaries this script has access to cannot change during the execution, right? If so, we don't need to display the version of python every time we shell out, do we?

src/taskgraph/run-task/run-task
```diff
@@ -1078,6 +1084,51 @@ def _display_python_version():
         b"Python version: "+ bytes(platform.python_version(), 'utf-8')
     )
 
+def _display_python_version_in_sub_process(subprocess: subprocess):
```
JohanLorenzo, line 1087: Refactor: this function can be made much simpler with a for-loop that loops over `("python", "python2", "python3")`.

src/taskgraph/run-task/run-task
```diff
@@ -1078,6 +1084,51 @@ def _display_python_version():
...
+    print_line(
+        b"setup",
+        b"pyhton --version: ",
+    )
+    try:
+        print_line(
+            b"setup",
+            subprocess.check_output(["python", "--version"])
+        )
+    except:
```
JohanLorenzo, line 1097: Bug: Let's not catch all possible exceptions here. This will likely bite us in the future (e.g.: the machine is in a bad state, the script doesn't manage to shell out for whatever reason but it considers it's just Python missing).

Instead, I'd recommend catching `FileNotFoundError` which is the one raised: 

```
>>> subprocess.check_output(["python2", "--version"])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 548, in run
    with Popen(*popenargs, **kwargs) as process:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 1024, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 1917, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'python2'
```

src/taskgraph/run-task/run-task
```diff
@@ -1078,6 +1084,51 @@ def _display_python_version():
...
 
+def _display_python_version_in_sub_process(subprocess: subprocess):
+    print_line(
+        b"setup",
+        b"pyhton --version: ",
+    )
+    try:
+        print_line(
+            b"setup",
+            subprocess.check_output(["python", "--version"])
```
JohanLorenzo, line 1095: Improvement: Let's add the `--version` argument a second time. This will print out more debug data. E.g.: 

```
>>> subprocess.check_output(["python", "--version"])
b'Python 3.11.3\n'

>>> subprocess.check_output(["python", "--version", "--version"])
b'Python 3.11.3 (main, Apr  7 2023, 20:13:31) [Clang 14.0.0 (clang-1400.0.29.202)]\n'
```

I'd recommend adding a code comment to highlight to future readers this is intentional.

src/taskgraph/run-task/run-task
```diff
@@ -1078,6 +1084,51 @@ def _display_python_version():
         b"Python version: "+ bytes(platform.python_version(), 'utf-8')
     )
 
+def _display_python_version_in_sub_process(subprocess: subprocess):
+    print_line(
+        b"setup",
+        b"pyhton --version: ",
+    )
```
JohanLorenzo, line 1091: Nit: I'd recommend storing the command in a variable and then pretty print it here with:

```python
command = [python_executable, "--version",  "--version"]
print_line(b"setup", " ".join(command))
```

This reduces the likelihood of a typo and that prevents modifying the command at multiple locations the day we want to modify its flags, for instance.

test/test_scripts_run_task.py
```diff
@@ -391,8 +400,10 @@ def test_display_python_version_should_output_python_versions_title(
     assert ("Python version:" in capsys.readouterr().out) is True
 
 
-def test_display_python_version_should_output_python_versions(run_task_mod, capsys):
-    run_task_mod._display_python_version()
+def test_display_python_version_in_sub_process_should_output_python_versions(
```
JohanLorenzo, line 403: Improvement: I think we can seize the opportunity of testing the function more thoroughly thanks to monkeypatch, just like you did in the `test_clean_git_checkout()`. Here we can make `_check_output()` behave differently depending on the command it's given. This way, we can check whether the `except` statement works as expected.

That would be something close to (I didn't test this code): 

```python
    def _check_output(args):
        command = args[0]
        python_executable = command[0]
        if python_executable == "python2":
              raise FileNotFoundError("[Errno 2] No such file or directory: 'python2'")
        return "Dummy python 3 version"
```

JohanLorenzo (commented):
Thanks for taking care of this improvement @sewelseb! 😃 I think this PR is headed in the right direction. I think there are some places we can improve the implementation. I left more details below, let me know what you think 🙂

src/taskgraph/run-task/run-task
sewelseb: ok, I will update

src/taskgraph/run-task/run-task
sewelseb: I understand and I tought about doing it. but the "business" objectives are different. for example if we want to treat differently the case when we have a python 2.x involved. this is why i kept them separated. I can of course refactor.

JohanLorenzo: Got it! Could you expand on the py2.x case you foresee? Maybe there's something I missed here 🙂

ahal:
Closing due to inactivity. Please feel to re-open if you would still like to address the feedback :)

PR #276, opened 2023-06-22
ci: Add task for syntax checking

hneiva:
We currently don't enforce developers to install pre-commit. Adding a syntax check task will help reviewing PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,10 @@ ignore = [
     "E741",
 ]
 target-version = "py37"
+exclude = [
+  "src/taskgraph/run-task/robustcheckout.py",
+  "taskcluster/scripts/external_tools/tooltool.py",
+]
 
 [tool.ruff.isort]
 known-first-party = ["taskgraph"]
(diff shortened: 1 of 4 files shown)


hneiva:
I'm not sure if this is really needed. I just thought it was easier than trying to fix pre-commit.ci  integration.
@ahal thoughts?

hneiva:
Mmmmm upon further testing I think I know what made me question the integration.

1. We're [ignoring everything](https://github.com/taskcluster/taskgraph/blob/main/.pre-commit-config.yaml#L46). under run-task
I noticed this [PR](https://github.com/taskcluster/taskgraph/pull/273/files) has single quote strings, and black should nag about that.

2. Even if we remove the above exclusion, the file `src/taskgraph/run-task/run-task` isn't checked by black or ruff because it lacks a file extension
Given that we've been maintaining that file, I'd suggest we at least include it in ruff and black.

hneiva:
Will create a new PR for proper fixes

PR #277, opened 2023-06-23
feat: add `purgeCaches` option to gw payload builder

matt-boris:
This feature was added to generic worker in https://github.com/taskcluster/taskcluster/pull/6148.

Only thing I wasn't sure on here was whether or not to include [this block](https://github.com/taskcluster/taskgraph/blob/d2a4cd791e9dea2dc8355585588daec5320c71a8/src/taskgraph/transforms/task.py#L445-L452).


ahal (approved):
Thanks!

PR #278, opened 2023-06-23
Add support for `all` grouping in `from_deps`

bhearsum:
I found a need for this while working on Translations, where there are multiple leaf nodes at the end of the graph - and having a single task that includes them all makes it easy to target everything.

### DIFF
--- a/docs/reference/transforms/from_deps.rst
+++ b/docs/reference/transforms/from_deps.rst
@@ -111,11 +111,34 @@ groups:
 Then the ``notify`` task will be duplicated into three, one for each group. The
 notify tasks will depend on each task in its associated group.
 
+You may also provide a special value of ``all`` to the ``group-by`` function.
+Using ``all`` will *always* result in one task being generated, with all tasks
+from the included kinds to be set as dependencies. It is usually useful to also
+set ``unique-kinds`` to ``False`` when using ``all``.
+
+If we alter the ``kind`` definition from above as follows:
+
+.. code-block:: yaml
+
+   kind-dependencies:
+     - build
+     - signing
+     - publish
+
+   tasks:
+     notify:
+       from-deps:
+         group-by: all
+
+We would end up with a single ``notify`` task that depends on all tasks from
+the ``build``, ``signing``, and ``publish`` kinds.
+
+
 Custom Grouping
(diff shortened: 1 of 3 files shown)


docs/reference/transforms/from_deps.rst
```diff
@@ -111,20 +111,43 @@ groups:
...
+
+   tasks:
+     notify:
+       from-deps:
+         group-by: all
+
+We would end up with a single ``notify`` task that depends on all tasks from
+the ``build``, ``signing``, and ``publish`` kinds.
+
+
```
ahal, line 136: Thanks for the docs! I guess we should have a dedicated reference section for all of the built-in `group-by` functions, but that can be a follow-up and this is great for now.

docs/reference/transforms/from_deps.rst
```diff
@@ -111,20 +111,43 @@ groups:
...
 implemented as well:
 
 .. code-block:: python
 
    from typing import List
 
    from taskgraph.task import Task
-   from taskgraph.transforms.from_deps import group_by
    from taskgraph.transforms.base import TransformConfig
+   from taskgraph.util.dependencies import group_by
```
ahal, line 150: Oops, nice catch!

src/taskgraph/util/dependencies.py
```diff
@@ -27,6 +27,11 @@ def group_by_single(config, tasks):
         yield [task]
 
 
+@group_by("all")
+def group_by_all(config, tasks):
+    return [[task for task in tasks]]
```
ahal, line 32: Shouldn't this just be `return [tasks]`?

ahal (approved):
Makes sense, thanks for the docs and tests!

src/taskgraph/util/dependencies.py
bhearsum: That was my instinct as well, but nope. [The transform](https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/transforms/from_deps.py#L151-L158) pretty clearly wants a list of lists.

(`group_by_attribute` also returns a list of lists. And `group_by_single` yields a list in each iteration of its loop.)

PR #279, opened 2023-06-23
chore: Fix syntax checking to include run-task file

hneiva:


hneiva:
Here is another example of the GH integration behaving different from local commit hooks.
When I committed my changes it didn't complain about fetch-content file 😵‍💫

pyproject.toml
```diff
@@ -26,6 +27,14 @@ ignore = [
     "E741",
 ]
 target-version = "py37"
+exclude = [
+  "src/taskgraph/run-task/robustcheckout.py",
+  "taskcluster/scripts/external_tools/tooltool.py",
```
ahal, line 32: Maybe we should just exclude all of `external_tools` as presumably we wouldn't want to lint anything in there.

ahal (approved):

PR #283, opened 2023-06-27
Logging the command used to invoke run-task

yaxhkabra:
Added the call to `print_line` for logging the command used to inoke the run-task script. Used `sys.argv` to capture the command invoking `run task`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## [5.4.0] - 2023-06-13
 
-### Added
-
-- Loggign the command invkoing the run-script
-
 ### Fixed
 
 - Regression in add-new-jobs and retrigger actions, introduced in 5.1.0.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [5.4.0] - 2023-06-13
 
+### Added
+
+- Loggign the command invkoing the run-script
+
 ### Fixed
 
 - Regression in add-new-jobs and retrigger actions, introduced in 5.1.0.
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1083,8 +1083,9 @@ def main(args):
     )
     calling_script =' '.join(args)
     print_line(
-    b"setup",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [5.4.0] - 2023-06-13
 
+### Added
+
+- Loggign the command invkoing the run-script
+
```
ahal, line 8: Nice thought, though we don't typically update the changelog until we make a release. The `5.4.0` release has already gone out the door :).

So please remove this change and I'll be sure to mention it the next time we release.

src/taskgraph/run-task/run-task
```diff
@@ -1081,6 +1081,12 @@ def main(args):
         b"setup",
         b"run-task started in %s\n" % os.environ["TASK_WORKDIR"].encode("utf-8"),
     )
+    #logging command which envoked the script
```
ahal, line 1084: nit: I don't think this comment is necessary, the code is pretty self-explanatory.

src/taskgraph/run-task/run-task
```diff
@@ -1081,6 +1081,12 @@ def main(args):
...
+    calling_script = sys.argv[0]
```
ahal, line 1085: We don't want to just log the script, but rather the full command line. We should also use `args` here rather than `sys.argv` in case something called this `main` function programmatically.

src/taskgraph/run-task/run-task
```diff
@@ -1081,6 +1081,12 @@ def main(args):
...
+    print_line(
+    b"script",
```
ahal, line 1087: I think the `setup` component is better for this.

ahal (changes requested):
Thanks for the PR!

yaxhkabra:
Sure! I'll do the changes.

yaxhkabra:
Updated print line function, used args to capture command. Removed the log file changes.

ahal (approved):
Sorry for the delay in review, thanks for the patch!

p.s I took the liberty of fixing a minor style nit rather than sending it back your way.

PR #286, opened 2023-07-04
fix(vcs): make `repo.find_latest_common_revision` return null revision if there's no common ancestor

ahal:


test/test_util_vcs.py
```diff
@@ -415,20 +415,31 @@ def test_find_latest_common_revision(repo_with_remote):
...
     assert repo.head_rev != expected_latest_common_revision
 
-    if repo.tool == "git":
-        base_ref = f"{remote_name}/{repo.branch}"
-        assert (
-            repo.find_latest_common_revision(base_ref, repo.head_rev)
-            == expected_latest_common_revision
-        )
-    else:
+    if repo.tool == "hg":
```
jcristau, line 418: What does find_latest_common_revision return for unrelated commits in the hg case?

ahal: It took me awhile to figure out how to do that (I don't think it's something that would come up in practice), but looks like it returns an empty string. But ok ok, I'll make it consistent :p.. I was being lazy.

jcristau (approved):
Thanks.  Please update commit message and PR title now that it's not just git :)

PR #290, opened 2023-07-07
feat: allow overriding default task deadline

bhearsum:
This has come up with the work on our Translations model training pipeline. Full training runs are expected to take weeks or even a month to complete. Easily allowing the entire graph to have a longer deadline is the easiest way to cope with this, and I imagine other things could benefit as well, eg: fuzzing.

### DIFF
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -35,9 +35,10 @@ graph_config_schema = Schema(
                 "lowest",
             ),
         ),
-        Optional("task-deadline-after",
-                 description="Default 'deadline' for tasks, in relative date format. "
-                 "Eg: '1 week'"
+        Optional(
+            "task-deadline-after",
+            description="Default 'deadline' for tasks, in relative date format. "
+            "Eg: '1 week'",
         ): optionally_keyed_by("project", str),
         Required("workers"): {
             Required("aliases"): {
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -35,6 +35,10 @@ graph_config_schema = Schema(
                 "lowest",
             ),
         ),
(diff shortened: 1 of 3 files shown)


ahal (changes requested):
Nice, makes sense!

I think you'll need to update the [graph_config_schema](https://github.com/taskcluster/taskgraph/blob/42f95496ada66b06e0250334d1d84ad79635fe78/src/taskgraph/config.py#L21) too. I think it should also go under the `taskgraph` subkey.

ahal:
> I think you'll need to update the [graph_config_schema](https://github.com/taskcluster/taskgraph/blob/42f95496ada66b06e0250334d1d84ad79635fe78/src/taskgraph/config.py#L21) too. I think it should also go under the `taskgraph` subkey.

Oh I guess this works because that schema has `extra: object` in it.. but since this is a core Taskgraph feature, it's best to document it.

bhearsum:
Updated the schema -- I'm happy to move this to the `taskgraph` section if you want, too. (Although based on our Matrix conversation I'm not sure it matters much...)

ahal:
Yeah, feel free to leave it.

src/taskgraph/config.py
```diff
@@ -35,6 +35,7 @@
                 "lowest",
             ),
         ),
+        Optional("task-deadline-after"): optionally_keyed_by("project", str),
```
ahal, line 38: Might be nice to add the `description` field here to describe the format it expects :)

ahal (approved):
Thanks!

src/taskgraph/config.py
bhearsum: That's a good idea, yeah.

PR #291, opened 2023-07-09
Add abillity to build firefox with mach from source tarball

smurfd:
Hey,
I was notified that changes i had done in https://bugzilla.mozilla.org/show_bug.cgi?id=1798746
to the `mozilla-central/third_party/python/taskcluster_taskgraph` were overwritten, and that they should have been changed here in the taskgraph github instead.

The reason for the change is to be able to build firefox from release tarball using the `mach` command (which was not previously doable)

### DIFF
--- a/src/taskgraph/generator.py
+++ b/src/taskgraph/generator.py
@@ -272,7 +272,7 @@ class TaskGraphGenerator:
         # Always add legacy target tasks method until we deprecate that API.
         if "target_tasks_method" not in filters:
             filters.insert(0, "target_tasks_method")
-        filters = [filter_tasks.filter_task_functions[f] for f in filters]
+        filters = [filter_tasks.filter_task_functions[f] for f in filters if f]
 
         yield self.verify("parameters", parameters)
 
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -121,15 +121,17 @@ def _get_defaults(repo_root=None):
             repo_url = ""
             project = ""
 
-        defaults.update({
-            "base_repository": repo_url,
-            "head_ref": repo.branch or repo.head_rev,
-            "head_repository": repo_url,
-            "head_rev": repo.head_rev,
-            "project": project,
-            "repository_type": repo.tool,
-            "version": get_version(repo_path),
-        })
+        defaults.update(
+            {
+                "base_repository": repo_url,
+                "head_ref": repo.branch or repo.head_rev,
+                "head_repository": repo_url,
+                "head_rev": repo.head_rev,
+                "project": project,
+                "repository_type": repo.tool,
+                "version": get_version(repo_path),
+            }
+        )
     return defaults
 
 
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -80,20 +80,8 @@ def get_version(repo_path):
 def _get_defaults(repo_root=None):
     repo_path = repo_root or os.getcwd()
     repo = get_repository(repo_path)
-    try:
-        repo_url = repo.get_url()
-        parsed_url = mozilla_repo_urls.parse(repo_url)
-        project = parsed_url.repo_name
-    except (
-        CalledProcessError,
-        mozilla_repo_urls.errors.InvalidRepoUrlError,
-        mozilla_repo_urls.errors.UnsupportedPlatformError,
-    ):
-        repo_url = ""
-        project = ""
-
-    return {
-        "base_repository": repo_url,
+    defaults = {
+        "base_repository": "SOURCE",
         "base_ref": "",
         "base_rev": "",
         "build_date": int(time.time()),
@@ -102,9 +90,9 @@ def _get_defaults(repo_root=None):
         "enable_always_target": True,
         "existing_tasks": {},
         "filters": ["target_tasks_method"],
-        "head_ref": repo.branch or repo.head_rev,
-        "head_repository": repo_url,
-        "head_rev": repo.head_rev,
+        "head_ref": "",
+        "head_repository": "",
+        "head_rev": "",
         "head_tag": "",
         "level": "3",
         "moz_build_date": datetime.now().strftime("%Y%m%d%H%M%S"),
@@ -112,14 +100,37 @@ def _get_defaults(repo_root=None):
         "optimize_strategies": None,
         "optimize_target_tasks": True,
         "owner": "nobody@mozilla.com",
-        "project": project,
+        "project": "",
         "pushdate": int(time.time()),
         "pushlog_id": "0",
-        "repository_type": repo.tool,
+        "repository_type": "",
         "target_tasks_method": "default",
         "tasks_for": "",
-        "version": get_version(repo_path),
+        "version": "",
     }
+    if repo:
+        try:
+            repo_url = repo.get_url()
+            parsed_url = mozilla_repo_urls.parse(repo_url)
+            project = parsed_url.repo_name
+        except (
+            CalledProcessError,
+            mozilla_repo_urls.errors.InvalidRepoUrlError,
+            mozilla_repo_urls.errors.UnsupportedPlatformError,
+        ):
+            repo_url = ""
+            project = ""
+
+        defaults.update({
+            "base_repository": repo_url,
+            "head_ref": repo.branch or repo.head_rev,
+            "head_repository": repo_url,
+            "head_rev": repo.head_rev,
+            "project": project,
+            "repository_type": repo.tool,
+            "version": get_version(repo_path),
+        })
+    return defaults
 
 
 defaults_functions = [_get_defaults]
@@ -195,13 +206,14 @@ class Parameters(ReadOnlyDict):
     @staticmethod
     def _fill_defaults(repo_root=None, **kwargs):
         defaults = {}
-        for fn in defaults_functions:
-            defaults.update(fn(repo_root))
+        if repo_root != "SOURCE":
+            for fn in defaults_functions:
+                defaults.update(fn(repo_root))
 
-        for name, default in defaults.items():
-            if name not in kwargs:
-                kwargs[name] = default
-        return kwargs
+            for name, default in defaults.items():
+                if name not in kwargs:
+                    kwargs[name] = default
+            return kwargs
 
     def check(self):
         schema = (
--- a/src/taskgraph/util/vcs.py
+++ b/src/taskgraph/util/vcs.py
@@ -511,8 +511,9 @@ def get_repository(path):
             return HgRepository(path)
         elif os.path.exists(os.path.join(path, ".git")):
             return GitRepository(path)
-
-    raise RuntimeError("Current directory is neither a git or hg repository")
+        elif os.path.exists(os.path.join(path, "moz.configure")):
+            return None
+    raise RuntimeError("Current directory is neither a git or hg repository, nor a release source")
 
 
 def find_hg_revision_push_info(repository, revision):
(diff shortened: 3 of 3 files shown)


src/taskgraph/parameters.py
```diff
@@ -80,46 +80,76 @@ def get_version(repo_path):
...
-        "owner": "nobody@mozilla.com",
-        "project": project,
-        "pushdate": int(time.time()),
-        "pushlog_id": "0",
-        "repository_type": repo.tool,
-        "target_tasks_method": "default",
-        "tasks_for": "",
-        "version": get_version(repo_path),
-    }
+    if repo:
```
ahal, line 83: Thanks, this seems fine. Though rather than a big `if` statement here, let's try to only adjust the defaults for the parameters that rely on VCS. For example something like:
```python
defaults ={
    "base_repository": "SOURCE",
    "base_ref": "",
    "base_rev": "",
    "build_date": int(time.time()),
    ...
}

if repo:
    defaults.update({
        "base_repository": repo.get_url(),
        "head_ref": repo.branch or repo.head_rev,
         ...
    })
return defaults

**Review by ahal — CHANGES_REQUESTED** [review]
Sorry I didn't notice this sooner! Normally I only notice PRs if I'm requested for review, but I guess you didn't have permission to do that here.

**Comment by smurfd** [comment]
Hey, no worries at all! To be honest, i did not even try to request review. Was not sure how things worked here over at github. :)

Yes that makes alot of sense.

File: src/taskgraph/parameters.py [hunk 2]
```diff
@@ -195,13 +208,14 @@ def format_spec(spec):
...
     def _fill_defaults(repo_root=None, **kwargs):
         defaults = {}
-        for fn in defaults_functions:
-            defaults.update(fn(repo_root))
-
-        for name, default in defaults.items():
-            if name not in kwargs:
-                kwargs[name] = default
-        return kwargs
+        if repo_root != "SOURCE":
```
ahal, line 211: Hm, I just noticed this line and I'm a bit confused now. If this is set to `SOURCE`, then we'll never call that `_get_defaults()` function in the first place. So are all the changes in there even necessary?

And come to think of it, could the calling code in `mozilla-central` just monkeypatch the `defaults_functions` global variable in this module? I think it might actually be possible to accomplish what you need here without any changes to upstream Taskgraph at all.

smurfd: Let me try to clearify. If repo_root == "SOURCE", that means the root source directory of firefox you are running `mach` in, is like a firefox-115.0.3.tar.xz release source, not a Mercurial or Git checkout. The idea with the patch for mozilla-central is to be able to run `mach build` in a non-cvs folder. (sorry if that was obvious :))

Correct me if im wrong, but will we not always call the `_get_defaults()` function on this row:
`defaults_functions = [_get_defaults]`

However, if we manage to make magic with the `_get_defaults()` without touching upstream.
We still need the changes in upstream taskgraph to `vcs.py` and `generator.py` for `mach` to recognize the folder its run from, if its not a git or mercurial checkout.

ahal: > Correct me if im wrong, but will we not always call the _get_defaults() function on this row:
defaults_functions = [_get_defaults]

No this isn't calling the function, it's just storing it for later. It isn't invoked until line 213 (`fn(repo_root)`). So all the changes in `_get_defaults` are not necessary (I think).

> We still need the changes in upstream taskgraph to vcs.py and generator.py for mach to recognize the folder its run from, if its not a git or mercurial checkout.

I think they might be unnesseary as well, but I'll start some new comment threads, one for each.

src/taskgraph/generator.py
```diff
@@ -272,7 +272,7 @@ def _run(self):
         # Always add legacy target tasks method until we deprecate that API.
         if "target_tasks_method" not in filters:
             filters.insert(0, "target_tasks_method")
-        filters = [filter_tasks.filter_task_functions[f] for f in filters]
+        filters = [filter_tasks.filter_task_functions[f] for f in filters if f]
```
ahal, line 275: Do you have a traceback of when this caused an error? The `if f` should only be checking that the filter name isn't `None` or an empty string or something. I'm not sure why that would ever happen.

src/taskgraph/util/vcs.py
```diff
@@ -511,8 +511,11 @@ def get_repository(path):
             return HgRepository(path)
         elif os.path.exists(os.path.join(path, ".git")):
             return GitRepository(path)
-
-    raise RuntimeError("Current directory is neither a git or hg repository")
+        elif os.path.exists(os.path.join(path, "moz.configure")):
+            return None
+    raise RuntimeError(
+        "Current directory is neither a git or hg repository, nor a release source"
+    )
```
ahal, line 518: I don't love putting Gecko specific logic here. Do you know where `get_repository` was being called from? We've established that the invocation in `parameters.py` can be bypassed. That only leaves `files_changed.py` (which definitely needs a VCS to work anyway), `decision.py` (which I wouldn't suspect would be used for your use case) and `main.py`.

Perhaps we could instead update `main.py` to catch `RuntimeError`? Maybe we could invent a new `Exception` class to better handle it too.

ahal: Or failing that, I'd rather we always return `None` here, and then assert that `repo is not None` for any existing callers that are expecting it.

ahal: (although that latter proposal is backwards incompatible, so would require a major version bump.. so I'm leaning towards the former)

src/taskgraph/generator.py
smurfd: this change seems like we can remove

src/taskgraph/parameters.py
smurfd: this `if` does not seem like its needed either

src/taskgraph/util/vcs.py
smurfd: `get_repository()` is called from `_get_defaults()`

If i run `./mach bootstrap` without any of the changes to taskgraph, what happens is that, the bootstrap passes through, installing os dependencies and python dependencies but the artifacts like `clang`,`clang-tidy`,`nasm` etc. is not installed, which could cause issues with `./mach build`:ing

for those to install i need to have the changes in `vcs.py` and `parameters.py`

I will dig deeper. i see where you come from, touch as little as possible in `taskgraph` :)
which i understand now, when this is not part of mozilla-central, but its own entity (which i didn't realize earlier until a few weeks ago :))

I have to remember how things held together, have forgotten that when the patch was pushed.
I get back here when i know more and have some better answers.

smurfd: hey @ahal 
I might have an idea. If it sounds good, we close this PR and i create a follow up bug in mozilla-bugzilla. mention this pr and cc you.

This is the callstack prettymuch for this.

maybe_install_private_packages_or_exit() is called from bootstrap() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozboot/mozboot/bootstrap.py#348
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozboot/mozboot/bootstrap.py#394

install_toolchain_artifact() is called from maybe_install_private_packages_or_exit()  in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozboot/mozboot/bootstrap.py#263

bootstrap_toolchain() is called from install_toolchain_artifact()  in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozboot/mozboot/base.py#347

bootstrap_path() is called from bootstrap_toolchain() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozbuild/mozbuild/bootstrap.py#52

bootstrap_toolchain_tasks() is called from bootstrap_path() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/build/moz.configure/bootstrap.configure#151

toolchain_task_definitions() is called from bootstrap_toolchain_tasks()  in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/build/moz.configure/bootstrap.configure#110

load_tasks_for_kind() is called from toolchain_task_definitions() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozbuild/mozbuild/toolchains.py#20

parameters_loader() is called from load_tasks_for_kind() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/generator.py#444

load_parameters_file() is called from parameters_loader() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#359

Parameters.__init__() is called from load_parameters_file() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#310

fill_defaults() is called from Parameters.__init__() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#159

get_defaults() via defaults_functions is called from fill_defaults() or extend_parameters_schema in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#146
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#198

get_repository() is called from get_defaults() in:
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/parameters.py#82

get_repository()
https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/third_party/python/taskcluster_taskgraph/taskgraph/util/vcs.py#505


And im thinking i would, just above https://searchfox.org/mozilla-central/rev/ddff4872b8220fae71155e39fd30ac3b760a4741/python/mozbuild/mozbuild/toolchains.py#20
call ´get_repository()` in a `try` / `except RuntimeError`.
If it hits `RuntimeError` i would check if a `moz.configure` file exists. Then make whatever calls i need from above callstack from the taskcluster things, either by hand or calling the actual functions from `python/mozbuild/mozbuild/toolchains.py`

If we dont hit the exception, we continue as usual.
Meaning no touching the `third_party/python/taskcluster_taskgraph/` stuff.

Makes sense?

src/taskgraph/generator.py
ahal: Yeah, just from reading the Python I'm pretty sure it'll be a no-op. If `f` is ever a false-y value, then it would cause an exception.. which is desired because it should never be a false-y value. Let's remove this.

src/taskgraph/util/vcs.py
ahal: Yes, that makes sense to me! And I think it's the right approach. While I'm not opposed to supporting running Taskgraph without a source checkout, it is very much an edge case and adds a fair amount of complexity. E.g, do only the paths touched by `./mach bootstrap` need to support it? Or do all transforms need to support it (I'd guess there are many places that assume we have a VCS in those)?

Detecting this in the code that consumes Taskgraph and taking appropriate action sounds great! If you're hitting errors and need some helping figuring out how to workaround them, let me know! (Though please use needinfo or ping me or I may miss it :))

ahal:
Closing per latest conversation.

PR #293, opened 2023-07-12
Much more broad support for task substitution

bhearsum:
While working on the Firefox Translations training pipeline I created a couple of very general transforms that had the effect of allowing substitution values into any part of a task definition. These values could be defined in the task templates themselves, or come from parameters. Both of these were inspired by the `command_context` support we already have in core Taskgraph, but take the idea much, much further.

My philosophy while working on this was to try to avoid kind-specific transforms at all costs, which meant having more generalized transforms that pushed some logic into the kinds via their configuration.

This pull request is a close-to-upstreamable version of those transforms. I'm not even sure we _want_ to support this here, so this is very much a draft at the moment, to be used to start a discussion.

Here are a couple of example from the Translations configuration that demonstrate the usage:
* In most kinds we substitute `src_locale` and `trg_locale` into a number of places (name, description, env or command, usually some dependencies and fetches). [This task](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/cefilter/kind.yml) is a good, simple example of that.
* In some cases, a context value may come from one of two places (think a specific value that can fallback to a default). [This kind shows that off](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/bicleaner/kind.yml#L57).
* Another good example is this kind where we have [tasks running a very similar command](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/merge-mono/kind.yml), with just [one argument being different between the two tasks](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/merge-mono/kind.yml#L97-L119).

It's worth calling out that task context is more or less just a generalized version of command context. The reason they're both here is because we don't have a good way to do multiple interpolations of the same field, so if, eg: you wanted to use `task-context` substitute something from parameters into `command`, but also use `command-context` to substitute something else - you'd end up getting a `KeyError` in the first `format()`.

If we decide we want to take this in full on in part there's definitely some clean-up that needs to happen - but this is a starting point.

### DIFF
--- a/docs/reference/transforms/task_context.rst
+++ b/docs/reference/transforms/task_context.rst
@@ -1,0 +1,137 @@
+.. _task_context:
+
+Task Context
+============
+
+The :mod:`taskgraph.transforms.task_context` transform can be used to
+substitute values into any field in a task with data that is not known
+until ``taskgraph`` runs.
+
+This data can be provided in a few ways, as described below.
+
+Usage
+-----
+
+from-parameters
+~~~~~~~~~~~~~~~
+
+Context can be pulled from parameters that are provided to ``taskgraph``.
+
+First, add the transform to the ``transforms`` key in your ``kind.yml`` file:
+
+.. code-block:: yaml
+
+   transforms:
+     - taskgraph.transforms.task_context
+     # ...
+
+Then create a ``task-context`` section in your task definition, e.g:
+
+.. code-block:: yaml
+
+   tasks:
+     build:
+       description: my description {foo}
+       task-context:
+         from-parameters:
+           foo: foo
+         substitution-fields:
+           - description
+
+When ``taskgraph`` is run the value of ``foo`` in the parameters it is provided
+will be substituted into the description. For example, if the following parameters
+are used:
+
+.. code-block:: yaml
+
+   foo: with some extra
+
+...the description will end up with a value of "my description with some extra".
+
+When using ``from-parameters`` you may also provide an ordered list of keys to
+look for in the parameters, with the first one found being used. For example,
+with the this ``kind``:
+
+.. code-block:: yaml
+
+   tasks:
+     build:
+       description: my description {foo}
+       task-context:
+         from-parameters:
+           foo:
+             - foo
+             - default
+         substitution-fields:
+           - description
+
+...the description will bring in the value ``foo`` from the parameters if
+present, or ``default`` otherwise.
+
+from-file
+~~~~~~~~~
+
+Context may also be provided from a defined yaml file. The provided file
+should usually only contain top level keys and values (eg: nested objects
+will not be interpolated - they will be substituted as text representations
+of the object).
+
+For example, with this kind definition:
+
+.. code-block:: yaml
+
+   tasks:
+     build:
+       description: my description {foo}
+       task-context:
+         from-file: some_file.yaml
+         substitution-fields:
+           - description
+
+And this in ``some_file.yaml``:
+
+.. code-block:: yaml
+
+   foo: from a file
+
+...description will end up with "my description from a file".
+
+
+from-object
+~~~~~~~~~~~
+
+You may also specify context as direct keys and values in the ``task-context``
+configuration in ``from-object`` . This can be useful in ``kinds`` that define
+most of their contents in ``task-defaults``, but have some values that may
+differ for various concrete ``tasks`` in the ``kind``.
+
+For example:
+
+.. code-block:: yaml
+
+   task-defaults:
+     description: my description {extra_desc}
+     task-context:
+       substitution-fields:
+         - description
+
+   tasks:
+     build1:
+       task-context:
+         from-object:
+           extra_desc: build1
+     build2:
+       task-context:
+         from-object:
+           extra_desc: build2
+
+This will give build1 and build2 descriptions with their ``extra_desc``
+included while allowing them to share the rest of their task definition.
+
+Precedence
+----------
+
+If the same key is found in multiple places the order of precedence
+is as follows: ``from-parameters``, ``from-object`` keys, ``from-file``.
+
+That is to say: parameters will always override anything else.
--- a/src/taskgraph/transforms/task_context.py
+++ b/src/taskgraph/transforms/task_context.py
@@ -1,11 +1,11 @@
-import copy
 from textwrap import dedent
+
+from voluptuous import ALLOW_EXTRA, Any, Optional, Required
 
 from taskgraph.transforms.base import TransformSequence
 from taskgraph.util.schema import Schema
 from taskgraph.util.templates import deep_get, substitute
 from taskgraph.util.yaml import load_yaml
-from voluptuous import ALLOW_EXTRA, Any, Extra, Optional, Required
 
 SCHEMA = Schema(
     {
--- a/src/taskgraph/transforms/task_context.py
+++ b/src/taskgraph/transforms/task_context.py
@@ -1,0 +1,121 @@
+import copy
+from textwrap import dedent
+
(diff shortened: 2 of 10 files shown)


src/taskgraph/util/dict_helpers.py
```diff
@@ -0,0 +1,10 @@
+def deep_get(dict_, field):
```
ahal, line 1: I think this would be a good fit alongside the functions in `util/templates.py` rather than a new file.

ahal: Also this can definitely be a follow-up, but there's similar logic for walking an object here:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/main.py#L96

Though in that case it's deleting rather than retrieving.

ahal (commented):
Nice!

Could this replace much of the `by-*` usage we have? I'm also wondering if it could entirely replace `command-context` in run_task. Did you just leave it in for backwards compatibilitiy?

bhearsum:
> Nice!
> 
> Could this replace much of the `by-*` usage we have?

I think it depends on the scenario. Certainly in cases where you have a bunch of concrete tasks that share a template/defaults, they could all specify their own `context` rather than using a `by-blah` in the template. I'm not sure I see how it can replace `by-blah` in, for example, [update-verify-config](https://searchfox.org/mozilla-central/rev/0b021daa5955f0288ef6477fc51be33f7f452f9a/taskcluster/ci/release-update-verify-config/kind.yml#30) -- where all the generated tasks end up with the same value.

I guess to put it generally: you can use task substitution in place of keyed by when the content varies by task, but not when the content varies by, eg: branch, release level, etc.

> I'm also wondering if it could entirely replace `command-context` in run_task. Did you just leave it in for backwards compatibilitiy?

Yes, I think it can fully replace `command-context`. I mostly left both in here for discussion purposes, and to avoid breaking compatibility. I don't have any real objection to completely replacing `command-context` though -- it'll just be more work when some projects pull in the change that drops it. (It will also mean we're even more different than `gecko_taskgraph` than we already are...but perhaps that ship has sailed already.)

bhearsum:
I _think_ I've addressed everything that came up in here and some of the common stuff from #294. I also tested it against Translations and Gecko to make sure it could properly replace the `command-context` and `task-substitution` that we do in those projects.

docs/reference/transforms/task_context.rst
```diff
@@ -0,0 +1,135 @@
...
+Then create a ``task-context`` section in your task definition, e.g:
+
+.. code-block:: yaml
+
+   tasks:
+     build:
+       description: my description {foo}
+       task-context:
+         from-parameters:
+           foo: foo
```
ahal, line 37: Ah so, these aren't necessarily the same as parameters in `parameters.yml`? Can we use a different name in that case? Maybe `from-object`? Or `from-dict`?

docs/reference/transforms/task_context.rst
```diff
@@ -0,0 +1,135 @@
...
+   task-defaults:
+     description: my description {extra_desc}
+     task-context:
+       substitution-fields:
+         - description
+
+   tasks:
+     build1:
+       task-context:
+         extra_desc: build1
```
ahal, line 121: I'm confused why the direct setting is needed, wouldn't this also work if they had:
```
task-context:
    from-parameters:
        extra_desc: build1
```

The `task-defaults` use a merge deep function, so it should work I think. But maybe I'm missing something.

docs/reference/transforms/task_context.rst
bhearsum: These are Parameters. The keys here are the names you'll use in the tasks, the values are where to find them in `parameters.yml`. (If you're wondering why we can't just use a list of strings...I guess we could, although it's not as obvious what you use in the `kind` when you're retrieving nested values, eg: https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/ci/collect-mono-trg/kind.yml#L27.)

docs/reference/transforms/task_context.rst
bhearsum: This is most useful when you have multiple entries in a `kind` that are mostly the same, but whose specific `task` entries may need different values in, eg: a command. https://github.com/mozilla/firefox-translations-training/blob/main/taskcluster/ci/merge-mono/kind.yml is a good example of this (it's still using `command-context`, but I think you'll get the point). These tasks are nearly identical, except that one uses `max_sent` that comes from `training_config.experiment.mono-max-sentences-src` in parameters, while the other uses `max_sent` from `training_config.experiment.mono-max-sentences-trg`.

ahal: Yeah sorry, I was confused because I thought `from-parameters` was arbitrary key/values as well and I didn't understand why you needed to have two methods of setting arbitrary key/values.

What if instead of setting these at the root of the `task-context` dict, we put them under a `from-object` or `from-dict` key instead? That way we don't pollute the top namespace with arbitrary keys and it's more consistent with the other two methods of providing context.

docs/reference/transforms/task_context.rst
ahal: Ah ok, my mistake. Thanks for clarifying!

I guess you could allow specifying nested values in the string as well and then the name could just include the dots or replace them with `_` or something... But this seems fine as well.

docs/reference/transforms/task_context.rst
bhearsum: That's an excellent idea - will do.

ahal:
Just noting that this will require a major version bump due to removing `command-context`. Also where is the Decision task?

ahal:
There it is.

src/taskgraph/transforms/task_context.py
```diff
@@ -0,0 +1,121 @@
...
+            their contents in `task-defaults`, but have some values that may
+            differ for various concrete `tasks` in the `kind`.
+
+            If the same key is found in multiple places the order of precedence
+            is as follows:
+              - Parameters
+              - `from-object` keys
+              - File
+
+            That is to say: parameters will always override anything else.
```
ahal, line 34: Wow, nice descriptions 👍

src/taskgraph/transforms/task_context.py
```diff
@@ -0,0 +1,121 @@
...
+                """
+            `task-context` can be used to substitute values into any field in a
+            task with data that is not known until `taskgraph` runs.
+
+            This data can be provided via `from-parameters` or `from-file`,
+            which can pull in values from parameters and a defined yml file
+            respectively.
+
+            Data may also be provided directly in the `from-object` section of
+            `task-context`.T his can be useful in `kinds` that define most of
```
ahal, line 24: nit: s/.T his/. This

src/taskgraph/transforms/task_context.py
```diff
@@ -0,0 +1,121 @@
...
+    },
+    extra=ALLOW_EXTRA,
+)
+
+transforms = TransformSequence()
+transforms.add_validate(SCHEMA)
+
+
+@transforms.add
+def render_command(config, jobs):
```
ahal, line 84: I'm guessing this may have been initially copied from the `command-context` feature? Should probably be called `render_task` or something.

ahal (approved):
Very nice, thanks for the thorough docs, descriptions and tests!

src/taskgraph/transforms/task_context.py
bhearsum: Err, yes!

PR #294, opened 2023-07-12
feat: transform that supports generic chunking and unchunking

bhearsum:
Chunking is a common feature in taskgraphs to run jobs in parallel. This transform contains two sequences to make it easy to set up basic chunking.

The `chunk` sequence is used to split a job into `total-chunks` chunks, and substitute `this_chunk` and `total_chunks` into any of the `substitution-fields`. Typically this is needed in `label` and `description` as well as at least one field that impacts the running job (eg: `env` or `command`).

The `unchunk` sequence is used to do the reverse: inject similar content into the task definition `total-chunks` times. This is very useful for adding dependencies and fetches when an upstream job in chunked.

I've been using a slightly modifide form of this in Firefox Translations. [This kind](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/train-teacher/kind.yml#L27) is a good example chunking, while [this task](https://github.com/mozilla/firefox-translations-training/blob/a6748776bc81be973951c47df65796668631a105/taskcluster/ci/collect-mono-src/kind.yml) is a good example of unchunking. A resulting graph can be found [over here](https://firefox-ci-tc.services.mozilla.com/tasks/groups/QzQ7DoOtR9mEeZqfcmwk3Q).


src/taskgraph/transforms/chunking.py
```diff
@@ -0,0 +1,94 @@
...
+        # Optional, so it can be used for a subset of jobs in a kind
+        Optional("chunk-config"): {
+            Required("total-chunks"): int,
+            Optional("substitution-fields"): [str],
+        }
+    },
+    extra=ALLOW_EXTRA,
+)
+
+chunk = TransformSequence()
```
ahal, line 23: Keep in mind that there's logic that defaults this variable to `transforms`. I think I'd keep that default for the `chunk` case here (I imagine `unchunk` will be much less common).

src/taskgraph/transforms/chunking.py
```diff
@@ -0,0 +1,94 @@
...
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import Schema
+from taskgraph.util.substitution import substitute
+
+CHUNK_SCHEMA = Schema(
+    {
+        # Optional, so it can be used for a subset of jobs in a kind
+        Optional("chunk-config"): {
+            Required("total-chunks"): int,
+            Optional("substitution-fields"): [str],
```
ahal, line 17: Could you add descriptions to these fields (see `from_deps` for an example). Also bonus points for adding documentation to:
https://github.com/taskcluster/taskgraph/tree/main/docs/reference/transforms

src/taskgraph/transforms/chunking.py
```diff
@@ -0,0 +1,94 @@
...
+
+                subcontainer = copy.deepcopy(container[subfield])
+                subfield = substitute(subfield, **subs)
+                container[subfield] = substitute(subcontainer, **subs)
+
+            yield subjob
+
+
+@unchunk.add
+def do_unchunking(config, jobs):
```
ahal, line 73: I'm a bit confused about the use case for unchunking. Do you have an example of when it might be used?

src/taskgraph/util/substitution.py
```diff
@@ -0,0 +1,16 @@
+def substitute(item, **subs):
```
ahal, line 1: Maybe this could also go in `util/templates.py`?

src/taskgraph/util/dict_helpers.py
```diff
@@ -0,0 +1,10 @@
+def deep_get(dict_, field):
```
ahal, line 1: As per your other PR, I think this is similar to the functions in `util/templates.py`. I agree the name isn't the greatest, but imo better to keep similar functions grouped together.

ahal (changes requested):

src/taskgraph/transforms/chunking.py
bhearsum: Good point, will do.

src/taskgraph/transforms/chunking.py
bhearsum: More generally: anywhere you want to depend on chunked jobs (regardless of whether or not you pull their artifacts), this can be useful.

ahal: Possible compromise idea:

Use `from_deps` to set dependencies as normal, then the `unchunk` transforms would iterate over all deps and scan for any that have a `chunk-substitutions` attribute (which was presumably set by the `chunk` transforms earlier). I think this way the downstream task collecting them wouldn't even need to define any `unchunk-config`, it would just implicitly get it from the dependency tasks.

bhearsum: You're absolutely right that there's a lot in common here with `from_deps`, and I definitely agree that it's the right choice to be setting the `dependencies` at the very lesat.

I'm a bit less sure what should be done about the `per-upstream-fields`. Both of the ideas that you proposed will obviously work, but neither feels very great to me because they separate what is currently all in one place into two places (either `from_deps` and an additional transform, or partly on the current task and partly on the upstream tasks). I think the `chunk-substitutions` idea also may not be flexible enough. For example: it's very common for chunking to inject `{this_chunk}` into the environment or command (not just the artifacts) - and we'd need a way for downstream tasks to pull in that bit.

With all of that in mind, I'm wondering about the idea of moving `unchunk` more or less wholesale into `from_deps` - and allowing it to do what `per-upstream-fields` currently does, and inject the explicitly specified content into tasks (rather than fetching it from the upstream tasks). Or perhaps an even more limited form of this where we allow defining specific per-task artifacts that should be fetched, eg:
```
kind-dependencies:
  - build

tasks:
  test:
    from-deps:
      group-by: chunk
      fetches:
        - artifact: build.tar.bz2
          dest: build-{this_chunk}.tar.bz2
```

Or if the artifacts have chunk numbers in them:
```
kind-dependencies:
  - build

tasks:
  test:
    from-deps:
      group-by: chunk
      total-chunks: 4
      fetches:
        - artifact: build-{this_chunk}.tar.bz2
```

ahal: Heh, it sounds like we have a bit of a philosophical difference of opinion, where one of us sees having everything in one place as a feature and the other sees it as a bug (and vice versa). I can see some pros and cons for both fwiw.

I guess I'm approaching this with the single responsibility principle in mind, where `from_deps` is designed to be really good at adding dependencies and nothing else. Then your new chunking and parameterization transforms could be really good at their single tasks and nothing else.

I don't know what exactly the magic combination is, but it feels like solving this "unchunking" problem would be do-able by composing some combination of the other three transforms together rather than having a single place that does it in a way that overlaps with the other two.

bhearsum: I don't think we're too far off, really. I agree that `from_deps` should be limited in what it does - but I guess in mind, because it's explicitly about dealing with dependencies, `fetches` (or `artifacts` if you prefer) are arguably in that scope as well. (I definitely agree that `from_deps` shouldn't be able to do anything outside of those two spaces though...) I guess the hitch there is that `fetches` are not something universal to all tasks...but rather to the `job` transforms, so maybe that doesn't fit in as well as it initially seemed it might.

I suppose the way I see it is that the goal here is "fetch artifacts from N upstreams" is the goal. The fact that we need to add `dependencies` along the way is an implementation detail, so needing to use two transforms to accomplish that one goal is a bit awkward. I guess this is also what we do in some other `from_deps` cases though...maybe this is just a fact of life with how `dependencies` and `fetches` are almost entirely separate concepts. (I would love it if they were more integrated...but that's obviously not something we're going to address here!)

As I write this I'm becoming more convinced that a follow-up transform in the way to go, especially considering it's similar to what we do in [places like firefox-android](https://github.com/mozilla-mobile/firefox-android/blob/32a1174dcb16cbddc53a7522808e9f84324b1606/taskcluster/ci/signing/kind.yml#L8). I'm still having trouble seeing how we can do without an `unchunk-config` (or equivalent) though. Whatever is adding the `fetches` could conceivably pull the total number of chunks from an upstream task, but per my earlier comment, I don't think we can simply pull everything that the `chunk` transform substituted (otherwise things like `description` or `env` are likely to get clobbered).

bhearsum: @ahal and I had a good synchronous chat about this yesterday. We both came to the conclusion that the main purpose of `unchunking` is to add `fetches` for artifacts from upstream jobs that were chunked.

With that in mind, `from_deps` can be fairly easily enhance to support adding dependencies for upstream tasks were chunked, and also extended to deal with adding `fetches` for these tasks as well.

Those changes are not entirely related to the `chunking` transform we're still adding here, so I've removed them here and will add them in a separate PR.

docs/reference/transforms/chunking.rst
```diff
@@ -0,0 +1,76 @@
+.. _chunking:
+
+Chunking Tasks
+==============
+
+The :mod:`taskgraph.transforms.chunking` transform contains a transforms
+sequence that aids in splitting a single entry in a ``kind`` into multiple
+multiple tasks. This is often used to parallelize expensive or slow work.
```
ahal, line 8: nit: multiple multiple

docs/reference/transforms/chunking.rst
```diff
@@ -0,0 +1,76 @@
+.. _chunking:
+
+Chunking Tasks
+==============
+
+The :mod:`taskgraph.transforms.chunking` transform contains a transforms
```
ahal, line 6: nit: ... module contains transforms that aid in splitting

src/taskgraph/transforms/chunking.py
```diff
@@ -0,0 +1,82 @@
...
+
+from taskgraph.transforms.base import TransformSequence
+from taskgraph.util.schema import Schema
+from taskgraph.util.templates import substitute
+
+CHUNK_SCHEMA = Schema(
+    {
+        # Optional, so it can be used for a subset of jobs in a kind
+        Optional(
+            "chunk-config",
```
ahal, line 17: I think I would mildly prefer this just be called `chunk` here (we don't append `-config` anywhere else). But not a big deal.

src/taskgraph/transforms/chunking.py
```diff
@@ -0,0 +1,82 @@
...
+    extra=ALLOW_EXTRA,
+)
+
+transforms = TransformSequence()
+transforms.add_validate(CHUNK_SCHEMA)
+
+
+@transforms.add
+def chunk_jobs(config, jobs):
+    for job in jobs:
```
ahal, line 54: nit: we've mostly migrated from `jobs` -> `tasks` in Taskgraph.

src/taskgraph/util/substitution.py
ahal: Or maybe `util/parameterization.py` would be even better.

ahal (approved):
Looks good, I have some minor nits but nothing fundamental.

src/taskgraph/transforms/chunking.py
bhearsum: That's a good call - this is the style we're using in `taskgraph` we may as well be consistent with it.

bhearsum:
> Or maybe util/parameterization.py would be even better.

Whoops - this actually ended up in `util/templates.py` as part of #293. We should probably just use that version instead :laughing:

PR #301, opened 2023-07-19
feat: allow 'enable_always_target' parameter to accept a list of kinds

ahal:
The driver behind this change is that in Gecko we want to always add 'docker-image' tasks even when `enable_always_target` would otherwise be False for all other kinds.

This will allow us to default to `enable_always_target: ["docker-image"]`, and then upgrade to `enable_always_target: True` on autoland or try.


ahal:
Here's roughly what the corresponding Gecko patch will look like once this is vendored:
https://hg.mozilla.org/try/rev/0312fa2739e1359fd761465f992b558b23acabee

jcristau (approved):
Clever.

PR #305, opened 2023-07-21
Detect and complain when group_by functions are registered with the s…

jcristau:
…ame name


bhearsum (approved):
Should we have a test for this? (I could go either way...)

jcristau:
I don't think I'll have to add a test today, so I'll land this for the sake of not leaving too many plates in the air as I disappear on PTO.

PR #313, opened 2023-08-03
feat: Support `fetches` and `set-name` in `from_deps`

bhearsum:
This came out of a long discussion that @ahal and I had about #294, where [we concluded that it would be better](https://github.com/taskcluster/taskgraph/pull/294#discussion_r1283529826) to have `from-deps` handle `fetches` than have a specialized transform that only deals with setting dependencies & fetches when an upstream task is chunked.

### DIFF
--- a/src/taskgraph/transforms/from_deps.py
+++ b/src/taskgraph/transforms/from_deps.py
@@ -16,8 +16,9 @@ from textwrap import dedent
 from voluptuous import Any, Extra, Optional, Required
 
 from taskgraph.transforms.base import TransformSequence
+from taskgraph.transforms.job import fetches_schema
 from taskgraph.util.attributes import attrmatch
-from taskgraph.util.dependencies import GROUP_BY_MAP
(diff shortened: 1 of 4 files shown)


src/taskgraph/transforms/from_deps.py
```diff
@@ -177,14 +201,40 @@ def from_deps(config, tasks):
...
+                attrs = new_task["attributes"]
                 new_task["attributes"] = primary_dep.attributes.copy()
                 new_task["attributes"].update(attrs)
 
+            if fetches:
+                new_task.setdefault("fetches", {})
+                task_fetches = new_task["fetches"]
+
+                for dep_label in new_task["dependencies"]:
+                    dep_task = config.kind_dependencies_tasks[dep_label]
```
ahal, line 221: I almost didn't bother commenting because it's such a trivial thing.. but you could use [get_dependencies](https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/util/dependencies.py#L50) at this point.

ahal (approved):
Looks good module some nits!

PR #315, opened 2023-08-08
feat: sort fetches by key and artifact to ensure a consistent ordering

bhearsum:
Fetches are not always processed in the same order across different runs of `taskgraph`. This change helps to make diffs of taskgraph output more readable. In cases where there are no changes to the fetches, it gets rid of the diffs altogether. In cases where there _are_ changes to the fetches, it helps to highlight the things that have actually changed.

In an ideal world we could pretty print these variables when doing diffs, but that's more effort than I'm willing to put in at the moment.


ahal (approved):
Thanks!

PR #316, opened 2023-08-08
fix: from-deps: don't override all existing dependencies; don't add fetch entry for kinds without fetches

bhearsum:
The former is arguably more than a "fix" in that it changes `from-deps` in a way that it is no longer the sole thing that manages upstream dependencies. In my opinion, if we are considering `from-deps` the official way to collect results from chunked/parallelized tasks, this is a necessary change, as we cannot predict the other things that such tasks may also depend on.

The latter is a pure bug fix: we shouldn't add a `fetch` block for kinds that have no `fetches` listed in `from-deps`.

(I can separate these two if it's useful.)

### DIFF
--- a/src/taskgraph/transforms/from_deps.py
+++ b/src/taskgraph/transforms/from_deps.py
@@ -184,9 +184,9 @@ def from_deps(config, tasks):
 
             new_task = deepcopy(task)
             new_task.setdefault("dependencies", {})
-            new_task["dependencies"].update({
-                dep.kind if unique_kinds else dep.label: dep.label for dep in group
-            })
+            new_task["dependencies"].update(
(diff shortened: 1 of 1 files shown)


src/taskgraph/transforms/from_deps.py
```diff
@@ -183,9 +183,10 @@ def from_deps(config, tasks):
                 )
 
             new_task = deepcopy(task)
-            new_task["dependencies"] = {
-                dep.kind if unique_kinds else dep.label: dep.label for dep in group
-            }
+            new_task.setdefault("dependencies", {})
+            new_task["dependencies"].update(
```
ahal, line 187: nit: `setdefault` returns the dict too, so you could just chain `.update` onto the end of that function.

ahal (approved):
Lgtm!

PR #327, opened 2023-08-17
Log the body of taskcluster errors

jcristau:
This means we log the entire error message instead of just the http status code.


jcristau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1705354 is an example where something like this can be helpful.

ahal (approved):

Issue #329, opened 2023-08-17
Exception running `taskgraph load-task`: TypeError: request() got an unexpected keyword argument 'chunked'


ahal:
STR:
```
$ taskgraph load-image --task-id=$TASK_ID
Traceback (most recent call last):
  File "/home/ahal/dev/taskgraph/src/taskgraph/main.py", line 558, in load_image
    ok = load_image_by_task_id(args["task_id"], args.get("tag"))
  File "/home/ahal/dev/taskgraph/src/taskgraph/docker.py", line 64, in load_image_by_task_id
    result = load_image(artifact_url, tag)
  File "/home/ahal/dev/taskgraph/src/taskgraph/docker.py", line 208, in load_image
    docker.post_to_docker(download_and_modify_image(), "/images/load", quiet=0)
  File "/home/ahal/dev/taskgraph/src/taskgraph/util/docker.py", line 50, in post_to_docker
    headers={"Content-Type": "application/x-tar"},
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/requests/sessions.py", line 637, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/requests/adapters.py", line 497, in send
    chunked=chunked,
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/urllib3/connectionpool.py", line 802, in urlopen
    **response_kw,
  File "/home/ahal/.pyenv/versions/taskgraph_37/lib/python3.7/site-packages/urllib3/connectionpool.py", line 504, in _make_request
    enforce_content_length=enforce_content_length,
TypeError: request() got an unexpected keyword argument 'chunked'
```

The error is happening here:
https://github.com/ahal/taskgraph/blob/dbf7a5b75e6b81c0083eb406130eb5e384555916/src/taskgraph/util/docker.py#L45

[...]

PR #330, opened 2023-08-17
fix: run docker commands in subprocess rather than over the unix socket

ahal:
The main reason for this is to work around issue #329. Between downgrading `requests` and adding `docker-py` as a dependency, this seems like the least bad option. As a bonus, it allows to remove a dependency as well!

Issue: #329


bhearsum (approved):
The only possible downside I can think of is that this command will no longer work if the docker client isn't installed. The only plausible place that could be an issue (AFAICT) is if taskgraph is running somewhere that's pointing at a remote docker daemon. I'm not aware of anywhere this actually happens, and I think it's an unlikely enough edge case that we ought not to worry about it.

ahal:
Agreed, plus `load-image` and `build-image` are only used for local debugging (afaik).

PR #332, opened 2023-08-17
fix(init): update bootstrap to use latest Taskgraph

ahal:


bhearsum (approved):

bhearsum:
I wonder if we should make updating this part of the taskgraph release process? I suppose there's a risk that we might need to do more than just change the requirements, but that should be pretty rare, and arguably should be done as part of individual PRs...

PR #333, opened 2023-08-18
docs: add a new tutorial that uses 'taskgraph init'

ahal:
We already have tutorials that cover building a `.taskcluster.yml` and `taskcluster` dir, however they do so from scratch and were created prior to the existence of `taskgraph init`.

This adds a tutorial for all the impatient people who just want to see tasks running and don't care about how things work.

### DIFF
--- a/docs/tutorials/getting-started.rst
+++ b/docs/tutorials/getting-started.rst
@@ -1,0 +1,124 @@
+Getting Started
+===============
+
+This tutorial will guide you towards a working Taskcluster + Taskgraph setup
+with your repo in as few steps as possible.
+
+.. note::
+
+   This tutorial has a heavy focus on the end result. If you prefer to build
+   understanding as you go, you may wish to follow the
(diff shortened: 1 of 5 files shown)


docs/tutorials/getting-started.rst
```diff
@@ -0,0 +1,125 @@
...
+to your repo. Some apps for known Taskcluster deployments are as follows:
+
+.. list-table:: Taskcluster Github Apps
+   :widths: 20 60
+   :header-rows: 1
+
+   * - Instance
+     - Github App
+   * - `Firefox-CI <https://firefox-ci-tc.services.mozilla.com>`_
+     - https://github.com/apps/firefoxci-taskcluster
```
JohanLorenzo, line 39: Thanks for documenting these URLs! I struggled so hard to find them, I had to bookrmark them 😅

JohanLorenzo (approved):
This is great! These docs are so nice to read. I like the fact it gives right enough information to get started while not overwhelming readers. Great job!

PR #338, opened 2023-08-25
Stop hardcoding the path to hg on mac

jcristau:
Best to just assume it's on $PATH, since that can be set as part of the task's payload.


jcristau:
This technically breaks backwards compatibility, but in practice I suspect nothing uses this.

gabrielBusta (approved):
I agree. If it does break something we can add hg to the path 🤷

Do you think we should do a major version bump for this?

gbrownmozilla (approved):

jcristau:
> Do you think we should do a major version bump for this?

Yes, I believe this warrants a major version bump.

PR #339, opened 2023-09-04
Fix load-image with an image name

jcristau:
- import `IndexSearch` from the right place
- pass missing deadline argument to `should_replace_task`
- fix error message to not refer to mozilla-central


hneiva (approved):
👍

jcristau:
Arguably this could do with test coverage...

ahal (approved):

PR #342, opened 2023-09-07
fix: ensure we can resolve default Parameters even if no repo is dete…

ahal:
…cted

This is being implemented to support a Gecko feature where they want to be able to run `./mach bootstrap` from a non-repo extracted artifact. Currently, since bootstrap depends on Taskgraph, it is not possible.

AIUI, they don't need to actually generate a graph and only the default parameters need to be able to work without a repo.

Bug: 1847288

### DIFF
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -11,6 +11,7 @@ from datetime import datetime
 from io import BytesIO
 from pprint import pformat
 from subprocess import CalledProcessError
+from unittest.mock import Mock
 from urllib.parse import urlparse
 from urllib.request import urlopen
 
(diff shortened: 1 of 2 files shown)


gabrielBusta (approved):

src/taskgraph/parameters.py
```diff
@@ -79,7 +80,13 @@ def get_version(repo_path):
 
 def _get_defaults(repo_root=None):
     repo_path = repo_root or os.getcwd()
-    repo = get_repository(repo_path)
+    try:
+        repo = get_repository(repo_path)
+    except RuntimeError:
+        # Use fake values if no repo is detected.
+        repo = Mock(branch="", head_rev="", tool="git")
```
gabrielBusta, line 87: You could do `repo = Mock(spec_set=GitRepository, branch="", head_rev="", tool="git")` so we can only set values in `GitRepository` here.

ahal: I just picked `git` because tool needs to be one of `hg` or `git` (and I figure `hg` is being phased out). But this is being used to mock out an hg repo, so I think it's fine to just leave it as a pure mock here and keep it updated if we depend on anything else.

Issue #345, opened 2023-09-19
`run-task` intermittently fails to clone repo on Windows with modern Git


ahal:
Basti from the VPN team noticed that their new `b-win2022` worker pool has intemittent Git permission failures while cloning the repo:
```
[taskcluster 2023-09-18T18:20:35.232Z] Executing command 0: C:/mozilla-build/python3/python3.exe run-task --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd
D:\task_169506064099579>C:/mozilla-build/python3/python3.exe run-task --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd 
[setup 2023-09-18T18:20:35.398Z] run-task started in D:\task_169506064099579
[setup 2023-09-18T18:20:35.398Z] Invoked by command: --mozillavpn-checkout=./build/src/ --task-cwd build/src/windows/installer -- cmd.exe /c build_prod.cmd
[setup 2023-09-18T18:20:35.398Z] Python version: 3.9.10
[vcs 2023-09-18T18:20:35.401Z] WARNING: vcs checkout path (D:\task_169506064099579\build\src) not in cache or volume; performance will likely suffer
[vcs 2023-09-18T18:20:35.401Z] executing ['git', 'fetch', '--tags', '--force', 'https://github.com/mozilla-mobile/mozilla-vpn-client', 'refs/heads/main']
[vcs 2023-09-18T18:20:35.414Z] fatal: detected dubious ownership in repository at 'D:/task_169506064099579/build/src'
[vcs 2023-09-18T18:20:35.414Z] 'D:/task_169506064099579/build/src' is owned by:
[vcs 2023-09-18T18:20:35.414Z] 	'S-1-5-21-867665065-978626105-2054798079-1000'
[vcs 2023-09-18T18:20:35.414Z] but the current user is:
[vcs 2023-09-18T18:20:35.414Z] 	'S-1-5-21-867665065-978626105-2054798079-1001'
[vcs 2023-09-18T18:20:35.414Z] To add an exception for this directory, call:
[vcs 2023-09-18T18:20:35.414Z] 
[vcs 2023-09-18T18:20:35.414Z] 	git config --global --add safe.directory D:/task_169506064099579/build/src
```
[...]

PR #346, opened 2023-09-19
fix: bypass Git's safe directory feature in 'run-task'

ahal:
As of Git 2.35.2, a security feature to mitigate a CVE on multi-user setups:
https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765

Basically Git will error out if the checkout dir isn't owned by the current user. This impacts `run-task` because checkouts can come from caches meaning they were cloned by a different task user.

As long as we aren't sharing caches across levels or pools (which we aren't), that other user is just as trusted as the current one. So it's safe to disable this check.

Issue: #345

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -624,6 +624,11 @@ def git_checkout(
             "Must specify both ssh_key_file and ssh_known_hosts_file, if either are specified",
         )
 
+    # Bypass Git's "safe directory" feature as the destination could be
+    # coming from a cache and therefore cloned by a different user.
+    args = ["git", "config", "--global", "--add", "safe.directory", Path(destination_path).as_posix()]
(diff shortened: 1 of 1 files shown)


jcristau (approved):
I feel like maybe it should be the worker's responsibility to handle ownership on caches...

jcristau:
(Ultimately here this means we trust other tasks not to do bad things in the git config of cached checkouts, which may or may not be reasonable...)

ahal:
> I feel like maybe it should be the worker's responsibility to handle ownership on caches...

I don't disagree with that! We'll need this to unblock things in the meantime though (and given we were previously using a Git that didn't even have this safety check, we're at least not making anything worse)

ahal:
I filed https://github.com/taskcluster/taskcluster/issues/6561

PR #348, opened 2023-09-20
Bug 1854152 - switch mobile decision image to Java 17

rvandermeulen:
reference-browser is hitting bustage trying to update to Gradle 8 because it depends on Java 17 and the decision images are still running 11.


rvandermeulen:
Green with A-S and Glean also.

jcristau (approved):

Issue #347, opened 2023-09-20
Could `taskgraph init` make it possible to configure the project name ?


julienw:
Currently `taskgraph init` uses the current directory name, but it's not always the best name to use for a project. Instead it should ask about it (with the directory name as a default) and/or make it possible to pass it on the command line.

Thanks!

ahal:
I think what we really want here is the name of the remote repo, as it'll need to match what's configured in `ci-configuration` and that's typically what we use when setting that up. Should be possible to extract this automatically, and fallback to the directory name if there aren't any remotes. There's already logic in Taskgraph for finding a default remote to use.

It should already be possible to run `taskgraph init --prompt`, but then you need to fill out all the context values, not just the name.

ahal:
This was fixed by #349

PR #349, opened 2023-09-20
fix(init): don't rely on local clone name for configuration

ahal:
Issue: #347

### DIFF
--- a/template/cookiecutter.json
+++ b/template/cookiecutter.json
@@ -1,7 +1,8 @@
 {
     "project_name": "my-project",
-    "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_').replace('-', '_') }}",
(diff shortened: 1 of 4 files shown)


template/cookiecutter.json
```diff
@@ -1,7 +1,8 @@
 {
     "project_name": "my-project",
-    "project_slug": "{{ cookiecutter.project_name.lower().replace(' ', '_').replace('-', '_') }}",
     "repo_host": ["github", "hgmo"],
+    "repo_name": "my-repo",
```
hneiva, line 4: what's the difference between repo_name and project_name?

ahal: Repo name is the name of the Github repository, project name is the name of the local checkout dir. Usually they're the same thing, but this fix is a result of that not being the case.

PR #355, opened 2023-10-06
Run an in-tree docker-image on generic-worker

gabrielBusta:
Just an experiment (I was messing with d2g)

### DIFF
--- a/src/taskgraph/transforms/job/run_task.py
+++ b/src/taskgraph/transforms/job/run_task.py
@@ -224,6 +224,38 @@ def generic_worker_run_task(config, job, taskdesc):
 
     if is_win:
         worker["command"] = [" ".join(command)]
+    elif worker.get("docker-image"):
+        worker["command"] = [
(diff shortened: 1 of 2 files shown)


gabrielBusta:
Maybe we could run d2g on the docker worker payloads, but we could tweak taskgraph like this ([got the idea back scrolling the lounge](https://matrix.to/#/!FLLnBnLWksRXuuPRBq:mozilla.org/$y06cTSqeXpa-mA-KpZqDPgJDGDVOu1DUBPH5odsMBVc?via=mozilla.org))

gabrielBusta:
Maybe we could have something like the community cluster?

https://community-tc.services.mozilla.com/tasks/KBsCjgSNTNq5ugvyHLHY5A/runs/0/logs/live/public/logs/live.log#L32-42

gabrielBusta:
Mm, I see that message mentions scopes... I haven't considered that at all

PR #357, opened 2023-10-11
Format chunks with zero padding

gabrielBusta:
I think these changes would solve https://github.com/mozilla/firefox-translations-training/issues/220

Since task can't have more than 99 dependencies I don't think we need to worry about people sorting things that go above 100.

### DIFF
--- a/src/taskgraph/transforms/chunking.py
+++ b/src/taskgraph/transforms/chunking.py
@@ -62,9 +62,10 @@ def chunk_tasks(config, tasks):
         for this_chunk in range(1, total_chunks + 1):
             subtask = copy.deepcopy(task)
 
+            # Format the chunks with zero padding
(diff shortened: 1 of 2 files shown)


gabrielBusta:
I have an in-tree fix but if it's not too disruptive, I think having these sort correctly would be nice

bhearsum (commented):
Although max dependencies is 99, there's no reason you can't chunk further than that. (You'll just need dummy tasks if you need downstream tasks.)

I have mixed feelings about this overall - the issue that brought this up is well outside the scope of chunking itself, and only arises when a downstream task in consuming artifacts from a chunked task that are named after the chunk _and_ the order in which they are processed is important. It's fairly edge case-y, and I think it's best handled where such cases come up (such as translations).

PR #361, opened 2023-11-07
feat: add optional local-run field to put "How to run job locally"

arai-a:
This is for [bug 1863187](https://bugzilla.mozilla.org/show_bug.cgi?id=1863187) and [bug 1863347](https://bugzilla.mozilla.org/show_bug.cgi?id=1863347), where I want to show "How to run job locally" to TreeHerder UI, to each job's view, in case the job fails on automation and people need to investigate the issue locally.

The optional `local-run` field is supposed to contain a string with list of commands, instructions (create file, etc), or a link to documentation.

example data is in https://phabricator.services.mozilla.com/D192856

### DIFF
--- a/src/taskgraph/transforms/job/__init__.py
+++ b/src/taskgraph/transforms/job/__init__.py
@@ -104,6 +104,8 @@ job_description_schema = Schema(
         # This object will be passed through to the task description, with additions
         # provided by the job's run-using function
         Optional("worker"): dict,
+        # Description for how to run the job locally.
+        Optional("local-run"): str,
     }
 )
 
--- a/src/taskgraph/transforms/task.py
+++ b/src/taskgraph/transforms/task.py
@@ -186,6 +186,8 @@ task_description_schema = Schema(
             Required("implementation"): str,
             Extra: object,
         },
+        # Description for how to run the job locally.
+        Optional("local-run"): str,
     }
 )
 


ahal:
In Gecko, these transforms are currently duplicated, e.g:
https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/transforms/task.py

So this code wouldn't be used from mozilla-central (though we would like to merge the two files at some point).

Though I wonder if it would be better to add this information to the `description` of a task? I don't think there's any reason the description needs to be limited to a single line. The problem with a separate key is that this won't actually end up in the task definition.. so it's kind of just serving the same purpose as a comment would. At least with the description, it'll end up in the final task definition.

ahal:
Given `m-c` doesn't use these files yet and this change needs to happen in `m-c` regardless, I'm going to close this PR here. We can discuss further in the bug / phab revision.

arai-a:
To my understanding, the files under `third_party` are used by m-c.
The try run fails if I modify only `taskcluster/gecko_taskgraph/transforms/` files, pointing the files under `third_party` in the stack trace.

https://treeherder.mozilla.org/jobs?repo=try&revision=8beeb4ccaea73af058522ffe5a2c034a090c8122
> [task 2023-11-06T17:10:09.294Z]   File "/builds/worker/checkouts/gecko/third_party/python/taskcluster_taskgraph/taskgraph/util/schema.py", line 30, in validate_schema
> [task 2023-11-06T17:10:09.294Z]     raise Exception("\n".join(msg) + "\n" + pprint.pformat(obj))

also, yeah, the field isn't yet copied to the task definition.  it's supposed to be happen later, after I put the example data in [bug 1863347](https://bugzilla.mozilla.org/show_bug.cgi?id=1863347).

PR #362, opened 2023-11-21
Bug 1865788 - retire decision-mobile docker image

jcristau:
Consumers have stopped using java in their decision task, so this is no longer necessary.


ahal (approved):
Awesome!

We'll want to update http://docs.mozilla-releng.net/en/latest/how-to/taskcluster/uploading_an_image.html#decision-images as well

jcristau:
docs updated in https://github.com/mozilla-releng/relengdocs/pull/255

PR #363, opened 2023-11-22
Split docker-worker caches by docker image (bug 1866109)

jcristau:
Sharing the same vcs checkout/storage dirs across different tasks running on different docker images runs the risk of incompatilibities, as shown e.g. when cloning a hg repo with one version of mercurial then trying to re-use it later using an older version.

Splitting by docker image means we lose some of the benefits of shared storage, but is safer in the face of tasks with (from taskgraph's point of view) unknown/arbitrary tool versions.  In practice this is hopefully good enough because there are few enough different docker images compared to the number of downstream tasks.

This requires a way to parametrize the cache name in the full task graph (at which stage in-tree docker images aren't resolved yet); we make it so that `{"task-reference": {"<foo>": "bar"}}` works and resolves `<foo>` to its task id at morph time.

### DIFF
--- a/src/taskgraph/util/parameterization.py
+++ b/src/taskgraph/util/parameterization.py
@@ -20,6 +20,12 @@ def _recurse(val, param_fns):
             if len(val) == 1:
                 for param_key, param_fn in param_fns.items():
                     if set(val.keys()) == {param_key}:
+                        if isinstance(val[param_key], dict):
+                            # handle `{"task-reference": {"<foo>": "bar"}}`
+                            return {
+                                param_fn(key): recurse(v)
+                                for key, v in val[param_key].items()
+                            }
                         return param_fn(val[param_key])
             return {k: recurse(v) for k, v in val.items()}
         else:
--- a/src/taskgraph/transforms/job/common.py
+++ b/src/taskgraph/transforms/job/common.py
@@ -130,12 +130,6 @@ def support_vcs_checkout(config, job, taskdesc, repo_configs, sparse=False):
     if sparse:
         cache_name += "-sparse"
 
-    # Workers using Mercurial >= 5.8 will enable revlog-compression-zstd, which
-    # workers using older versions can't understand, so they can't share cache.
-    # At the moment, only docker workers use the newer version.
-    if is_docker:
-        cache_name += "-hg58"
-
     add_cache(job, taskdesc, cache_name, checkoutdir)
 
     env = taskdesc["worker"].setdefault("env", {})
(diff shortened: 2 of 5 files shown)


src/taskgraph/util/parameterization.py
```diff
@@ -20,6 +20,11 @@ def recurse(val):
             if len(val) == 1:
                 for param_key, param_fn in param_fns.items():
                     if set(val.keys()) == {param_key}:
+                        if isinstance(val[param_key], dict):
```
bhearsum, line 23: For future readers: this is necessary because of the changes to `payload["cache"]`. Its direct value may be a dict with a `task-reference`, and the value _inside_ that `task-reference` may have a reference that needs resolving. Previously, I believe that values could be assumed to be a string.

(Should this be commented? Not going to make it a blocker seeing as this is already a very opaque, confusing function.)

bhearsum (approved):
This code is quite opaque to me, so just to sanity check one thing. This comment:
> This requires a way to parametrize the cache name in the full task graph (at which stage in-tree docker images aren't resolved yet); we make it so that {"task-reference": {"<foo>": "bar"}} works and resolves <foo> to its task id at morph time.

...explains why we can't just flip on [this code](https://github.com/taskcluster/taskgraph/blob/0401b911ec0a5d3a8b66bdacc183362aa7811871/src/taskgraph/transforms/task.py#L518-L522) for in-tree images?

(It's really annoying that we have two code paths for this sort of thing. It makes me wonder if we should consider desupporting external docker images, and require them to come in via a `fetch` - that's obviously not something we'll address here though!)

jcristau:
> This code is quite opaque to me, so just to sanity check one thing. This comment:
> 
> > This requires a way to parametrize the cache name in the full task graph (at which stage in-tree docker images aren't resolved yet); we make it so that {"task-reference": {"": "bar"}} works and resolves  to its task id at morph time.
> 
> ...explains why we can't just flip on [this code](https://github.com/taskcluster/taskgraph/blob/0401b911ec0a5d3a8b66bdacc183362aa7811871/src/taskgraph/transforms/task.py#L518-L522) for in-tree images?
> 
Right, at that stage we only know the name of the docker image, but since the contents can change we need to wait until we've computed the hash and/or task id to populate the cache name.

ahal (approved):

src/taskgraph/util/parameterization.py
jcristau: Added a comment.

jcristau:
I've updated this to add/update comments, and to remove `_run_task_suffix()` from the cache name suffix when using an in-tree docker image, since that should be part of the docker image's hash (via `%include-run-task` in a Dockerfile), and thus redundant.

jcristau:
> [...] remove `_run_task_suffix()` from the cache name suffix when using an in-tree docker image, since that should be part of the docker image's hash (via `%include-run-task` in a Dockerfile), and thus redundant.

Hah, except it breaks `check_run_task_caches`.  Oops.

PR #365, opened 2023-11-27
Normalize Path Separator in util.hash.py

strseb:
Hey! :) 

I just did setup taskgraph locally on a windows machine, and wanted to run the full moz-vpn taskgraph. That it was trying to match `taskcluster/scripts/toolchain/compile_openssl.ps1` vs the windows path `taskcluster\scripts\toolchain\compile_openssl.ps1` therefore failed :c


ahal (approved):
Lgtm!

Yeah unfortunately Taskgraph doesn't get run from a Windows environment too often (as Decision tasks are typically Linux). So thanks for catching!

PR #366, opened 2023-11-27
Update decision image base from ubuntu 20.04 to debian 12

jcristau:


jcristau:
This should go along with a major version bump, due to the downstream effects.

ahal:
> This should go along with a major version bump, due to the downstream effects.

I don't think so because at the moment decision image updates are decoupled from Taskgraph version updates. I think if we had consumers using e.g `FROM  taskgraph:v6.3.1` in their Dockerfiles, then it would make sense to tie into the version numbers.

ahal (approved):
Nice, thanks for doing this!

PR #376, opened 2023-12-04
refactor!: change 'root_dir' to point to main 'taskcluster' directory

ahal:
BREAKING CHANGE: root dir should now point to the 'taskcluster' dir rather than 'taskcluster/kinds'.

This only impacts projects that use `taskgraph.config.load_graph_config` or `taskgraph.generator.TaskGraphGenerator` directly.


ahal:
I debated whether this should truly be considered a breaking change or not, but given we are about to do a major bump, might as well err on the side of caution.

jcristau (approved):
lgtm

Issue #381, opened 2023-12-06
Stop installing `<PROJECT>_PIP_REQUIREMENTS` into system Python with `run-task`


ahal:
When a repo sets the `<PROJECT>_PIP_REQUIREMENTS` variable, `run-task` will automatically install the requirements file after cloning the repo. This is primarily used for bootstrapping Taskgraph in the Decision task, but can also be used for other things.

We recently upgraded the image to Debian 12, which includes Python 3.11. This version of Python fails when you attempt to pip install into the system Python:
```
[pip-install 2023-12-05T21:39:50.324Z] × This environment is externally managed
[pip-install 2023-12-05T21:39:50.324Z] ╰─> To install Python packages system-wide, try apt install
[pip-install 2023-12-05T21:39:50.324Z]     python3-xyz, where xyz is the package you are trying to
[pip-install 2023-12-05T21:39:50.324Z]     install.
[pip-install 2023-12-05T21:39:50.324Z]     
[pip-install 2023-12-05T21:39:50.324Z]     If you wish to install a non-Debian-packaged Python package,
[pip-install 2023-12-05T21:39:50.324Z]     create a virtual environment using python3 -m venv path/to/venv.
[pip-install 2023-12-05T21:39:50.324Z]     Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
[pip-install 2023-12-05T21:39:50.324Z]     sure you have python3-full installed.
[pip-install 2023-12-05T21:39:50.324Z]     
[pip-install 2023-12-05T21:39:50.324Z]     If you wish to install a non-Debian packaged Python application,
[pip-install 2023-12-05T21:39:50.324Z]     it may be easiest to use pipx install xyz, which will manage a
[pip-install 2023-12-05T21:39:50.324Z]     virtual environment for you. Make sure you have pipx installed.
[pip-install 2023-12-05T21:39:50.324Z]     
[pip-install 2023-12-05T21:39:50.324Z]     See /usr/share/doc/python3.11/README.venv for more information.
```

We'll work around this by passing `--break-system-packages` for now (this is probably fine anyway, especially with Docker). But it highlights some poor behaviour. It would be better to create a virtualenv first or use a non-system Python (perhaps installed via pyenv).

PR #382, opened 2023-12-06
fix(run-task): pass '--break-system-packages' when installing pip req…

ahal:
…uirements

This is needed on newer Pythons that raise an error when attempting to pip install to the system installation.


ahal:
See also issue #381 to track fixing this properly.

gbrownmozilla (approved):

Issue #222, opened 2023-04-10
Cached tasks don't work for pull-requests on level 1-only repositories


ahal:
There's some logic in Taskgraph such that the index used for cached tasks in pull request gets elevated to level 3:
https://github.com/taskcluster/taskgraph/blob/f21088f16bf2ad4f9b2973297bccec13b5051a01/src/taskgraph/util/cached_tasks.py#L61

We do this so PRs use cached tasks that were generated from a push (rather than another PR). However for repos where everything is level 1 (including pushes), these index routes will never exist and we'll never optimize the cached_task away.

Then again using cached_tasks in a level-1 only repo means that PRs could overwrite index tasks from pushes. One possible way of fixing this would be to add the `tasks_for` value to the index in this circumstance. Or maybe it's easiest to say that cached_tasks aren't supported in level-1 only repos.

bhearsum:
An interesting wrinkle I discovered here is where something starts off level 3, and then moves to level 1 (as we did with https://github.com/mozilla/firefox-translations-training). We ended up with a bunch of level 3 caches, and started getting very unexpected rebuilds of tasks in PRs when they started to expire.

Although it's quite rare to go from level 3 to level 1, it highlights how hacky the code around [cache level for github pull requests is](https://github.com/taskcluster/taskgraph/blob/d8f608f720b4598fdc5a54440b40befdca0ea489/src/taskgraph/util/cached_tasks.py#L54-L66). That block forces the usage of level 3 caches for pull requests without consideration for how the repositories is configured (which taskgraph doesn't even know about...).

Not supporting cached tasks on PRs is not really great. Any situation where, eg: you have an expensive toolchain to build wastes money and slows down development.

Adding `tasks_for` in the index route isn't a terrible idea, but it will require a bit more specialized code to make pull requests still use cached tasks from pushes? (Because we no longer just use `level`, we need to specifically at `push` - but only for pull requests...)

A potential alternative might be to avoid using level 1 both for pushes and pull requests -- and use level 2 for pushes instead. That way we have segregated caches and might even be able to get rid of the hack cited above?

ahal:
Another idea I just had was we could create a new route for cached tasks that include the branch name. Then PRs can be configured to use the cached task for whatever branch they are based on (or maybe just always the default branch?)

This feels nicer to me than using `tasks_for`.

I still like the idea of using level 2 to denote pushes, but that's a bigger change that will likely require an RFC and an RRA.

ahal:
[...]

Issue #359, opened 2023-10-17
revisit using cached tasks in PRs


bhearsum:
Our [cached tasks implementation explicitly refuses to replace tasks with tasks from the level 1 index](https://github.com/taskcluster/taskgraph/blame/01ba699bc214ddf323db02a59cdaacaf66c41dc2/src/taskgraph/util/cached_tasks.py#L54). The commit comment as well as [original phabricator revision](https://phabricator.services.mozilla.com/D42343) talk about chain of trust not handling "tasks built from commits not at the tip of a pull requests well" as the reason for this.

Up until now this hasn't been much of an issue, as it's fairly rare to have long lived pull requests that could benefit from re-using tasks in prior pushes to the PR. In [firefox translations](https://github.com/mozilla/firefox-translations-training), this is a bit different - where some PR tasks can both be lengthy and expensive. (Arguably any project that has PR tasks that run for more than a few minutes could benefit though...)

I suggest that we find out whether or not CoT still has the aforementioned issue, and fix it if it does.

We may also want to consider not allow different PRs to re-use tasks from other PRs (either by adjusting the index, or some sort of hard coded safeguard), although I'm not entirely sure it's necessary.

bhearsum:
Some of RelEng talked about this over Zoom today. Here's an attempt as a summary:
- In the past, when using level 1 cached tasks was allowed in PRs, there were issues with different PRs stomping on each other.
  - This could be prevented by including the PR number in the cached task digest, or index route. In either case we'd still want to use level 3 caches first if they exist.
- Due to the way scopes work, any PR can publish to any level 1 index. This means that this is a potential security issue to allow, particularly in places where `push` or `release` tasks also run on level 1 workers.

All of this means that this is not a trivial fix, and any changes most likely need an RRA.

bhearsum:
For context and a different perspective, one of the folks working on https://github.com/mozilla/firefox-translations-training offered this as what he expected as a user:
> I would expect the following behavior: if all of the data and code files that serve as a step input (including ci kinds and transforms) are unchanged, it should use the cached output. Sometimes you want to debug the last step of the pipeline on taskcluster and CI is a convenient way to do that. Then it shouldn’t rerun the whole thing every time. It adds to costs and wait time. I’ll file an issue in the repo

ahal:
[...]

PR #389, opened 2023-12-13
fix: ensure pull-requests properly use cached tasks

ahal:
Fixes issues #222 and #359

### DIFF
--- a/src/taskgraph/util/cached_tasks.py
+++ b/src/taskgraph/util/cached_tasks.py
@@ -7,6 +7,9 @@ import hashlib
 import time
 
 TARGET_CACHE_INDEX = "{cache_prefix}.cache.level-{level}.{type}.{name}.hash.{digest}"
+TARGET_PR_CACHE_INDEX = (
+    "{cache_prefix}.cache.head.{head_ref}.{type}.{name}.hash.{digest}"
+)
 EXTRA_CACHE_INDEXES = [
     "{cache_prefix}.cache.level-{level}.{type}.{name}.latest",
     "{cache_prefix}.cache.level-{level}.{type}.{name}.pushdate.{build_date_long}",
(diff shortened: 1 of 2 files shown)


src/taskgraph/util/cached_tasks.py
```diff
@@ -7,6 +7,9 @@
 import time
 
 TARGET_CACHE_INDEX = "{cache_prefix}.cache.level-{level}.{type}.{name}.hash.{digest}"
+TARGET_PR_CACHE_INDEX = (
+    "{cache_prefix}.cache.head.{head_ref}.{type}.{name}.hash.{digest}"
```
ahal, line 11: In theory, adding `head_ref` to this index is probably unnecessary and we could just make it a constant like "pr" or similar. But I also don't think it hurts and guarantees we only used cached tasks that were generated from the same PR. Which just feels cleaner, even if the digest basically guarantees this anyway.

src/taskgraph/util/cached_tasks.py
```diff
@@ -53,31 +56,43 @@ def add_optimization(
...
-        taskdesc["optimization"] = {"index-search": index_routes}
+    # Pull requests use a different target cache index route. This way we can
+    # be confident they won't be used by anything other than other pull
+    # requests.
+    if config.params["tasks_for"].startswith("github-pull-request"):
+        # Check the PR bucket first.
+        subs["head_ref"] = config.params["head_ref"]
+        if subs["head_ref"].startswith("refs/heads/"):
+            subs["head_ref"] = subs["head_ref"][11:]
+        index_routes.insert(0, TARGET_PR_CACHE_INDEX.format(**subs))
```
ahal, line 73: Hm, maybe we should check for the cached task in the PR namespace last?

src/taskgraph/util/cached_tasks.py
ahal: Although, if we switch to checking the PR index last (see comment below), then we'd be virtually guaranteed to never hit another PRs cache anyway. Only time it would happen is if both PRs make the exact same change to a resource file.

bhearsum: My gut says that `head_ref` is probably the best way to go. `pr` has some theoretical benefits...but they seem a little unlikely and farfetched for the amount of confusion that we'd be adding. (We could always change that later though...)

src/taskgraph/util/cached_tasks.py
bhearsum: I think that would be more consistent with what we do now: we look at the most trustworthy source first. In reality though, I don't see why it would make a difference unless someone finds hash collisions, or a cache digest is seriously misconfigured.

bhearsum (commented):
This needs tests and ideally some real world testing but it looks very sane to me. If you'd like, I can give it a try in Translations before it goes up for proper review and/or lands (I doubt it would cause any serious bustage if it causes an issues at all...)

bhearsum:
I gave this a quick try in a PR in translations: https://github.com/mozilla/firefox-translations-training/pull/303

It looks good there. The [first push](https://firefox-ci-tc.services.mozilla.com/tasks/groups/JVIh_-3LRA-c-bsscSY_qQ) busted all the caches due to the taskgraph update, but the tasks appear to have the correct routes. An [additional push](https://firefox-ci-tc.services.mozilla.com/tasks/groups/WGPNfZKiQ2eVL__Hg5FcnQ) optimized away all of the fetch and toolchain tasks, as expected.

This doesn't verify anything about fallback back to level 1 caches, but it's a good sign for this patch!

ahal:
Thanks for testing! And yes, I still plan on adding tests to this

ahal:
The tests were added in #394, so this PR just updates the expectations.

bhearsum (approved):

PR #397, opened 2023-12-18
chore: version bump to 7.0.1

bhearsum:


CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Change Log
 
+## [7.0.1] - 2023-12-18
+
+### Added
+
+- Ensure pull requests properly use cached tasks
```
ahal, line 7: Typically I go through the commit log and add everything that's changed. Happy to do that later too though.

ahal (approved):

CHANGELOG.md
ahal: And not that it really matters, but I typically map the commit prefixes like:

fix: -> goes under Fixed
feat: -> goes under Added

Would love for this to be auto-generated though

bhearsum: Ah, sorry! I missed the other items in the log. I'll fix this up.

PR #402, opened 2023-12-21
fix(init): strip '.git' out of repo_url if it exists

ahal:
This was causing directories like `my_repo.git_taskgraph` to be created.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -862,6 +862,8 @@ def init_taskgraph(options):
         return 1
 
     context["repo_name"] = urlparse(repo_url).path.rsplit("/", 1)[-1]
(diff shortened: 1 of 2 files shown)


bhearsum (approved):

src/taskgraph/main.py
```diff
@@ -862,6 +862,8 @@ def init_taskgraph(options):
         return 1
 
     context["repo_name"] = urlparse(repo_url).path.rsplit("/", 1)[-1]
+    if context["repo_name"].endswith(".git"):
```
JohanLorenzo, line 865: Improvement for another time: we could likely factor out this special case and use https://github.com/mozilla-releng/mozilla-repo-urls/. No need to handle this now, though.

PR #409, opened 2024-01-04
Allow expiration to be overridden in fetch and docker-image kinds. Allow setting the default expiry on "try" tasks

### LINKED ISSUE #350: The docker_image and fetch transforms don't accept deadline or expiration attributes
The defaults are no longer a safe assumption now that we can configure the maximum deadline in Firefox-CI (since https://github.com/taskcluster/taskcluster/pull/6490.) This contributed to https://github.com/mozilla/firefox-translations-training/issues/201.

### LINKED ISSUE #351: We need a way to set the expiration on tasks at the project level (like the deadline)
To fix https://github.com/mozilla/firefox-translations-training/issues/201 on https://github.com/mozilla/firefox-translations-training/pull/203 I had to set the task expiration of every task to a expiration compatible with the deadline set at the project level. We need a way to set the default expiration on the tasks at the project level. The default of 28 on pull-request tasks is not compatible with deadlines greater than the previously allowed Taskcluster maximum (https://github.com/taskcluster/taskcluster/pull/6490.)

### LINKED ISSUE #392: allow fetch expiration to be overridden in kinds
This came up in translations, where both in PRs and pushes (which run on level 1), the hardcoded 28 day expire is not sufficient.

@gabrielBusta has a short term fix in the translations repo (https://github.com/mozilla/firefox-translations-training/pull/203), but that should be upstreamed.

### LINKED ISSUE #393: allow docker-image expiration to be overridden in kinds
The same as #392, but for docker images -- we have a longer minimum expiry time for tasks in translations.

gabrielBusta:
fixes #392 
fixes #393 
fixes #350
fixes #351

### DIFF
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -42,7 +42,7 @@ graph_config_schema = Schema(
         ): optionally_keyed_by("project", str),
         Optional(
             "task-expires-after",
-            description="Default 'expires-after' for non level 3 tasks, in relative date format. "
+            description="Default 'expires-after' for level 1 tasks, in relative date format. "
             "Eg: '90 days'",
         ): str,
         Required("workers"): {
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -40,6 +40,11 @@ graph_config_schema = Schema(
             description="Default 'deadline' for tasks, in relative date format. "
             "Eg: '1 week'",
         ): optionally_keyed_by("project", str),
+        Optional(
+            "task-expires-after",
+            description="Default 'expires-after' for non level 3 tasks, in relative date format. "
+            "Eg: '90 days'",
+        ): str,
         Required("workers"): {
             Required("aliases"): {
                 str: {
--- a/src/taskgraph/transforms/docker_image.py
+++ b/src/taskgraph/transforms/docker_image.py
@@ -128,10 +128,7 @@ def fill_template(config, tasks):
         # burn more CPU once to reduce image size.
         zstd_level = "3" if int(config.params["level"]) == 1 else "10"
 
-        if "task-expires-after" in config.graph_config:
-            expires = config.graph_config["task-expires-after"]
-        else:
-            expires = "28 days"
+        expires = config.graph_config._config.get("task-expires-after", "28 days")
 
         # include some information that is useful in reconstructing this task
         # from JSON
--- a/src/taskgraph/transforms/docker_image.py
+++ b/src/taskgraph/transforms/docker_image.py
@@ -128,6 +128,11 @@ def fill_template(config, tasks):
         # burn more CPU once to reduce image size.
         zstd_level = "3" if int(config.params["level"]) == 1 else "10"
 
+        if "task-expires-after" in config.graph_config:
+            expires = config.graph_config["task-expires-after"]
+        else:
+            expires = "28 days"
+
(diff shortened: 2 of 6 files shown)


src/taskgraph/transforms/docker_image.py
```diff
@@ -128,6 +128,11 @@ def fill_template(config, tasks):
         # burn more CPU once to reduce image size.
         zstd_level = "3" if int(config.params["level"]) == 1 else "10"
 
+        if "task-expires-after" in config.graph_config:
+            expires = config.graph_config["task-expires-after"]
```
ahal, line 132: This should be added to `config.graph_config_schema`

src/taskgraph/transforms/docker_image.py
```diff
@@ -128,6 +128,11 @@ def fill_template(config, tasks):
...
+        else:
+            expires = "28 days"
```
ahal, line 134: suggestion:
```
expires = config.graph_config.get("task-expires-after", "28 days")

File: src/taskgraph/transforms/fetch.py [hunk 3]
```diff
@@ -108,6 +109,8 @@ def make_task(config, tasks):
     # essentially forever.
     if config.params["level"] == "3":
         expires = "1000 years"
+    elif "task-expires-after" in config.graph_config:
+        expires = config.graph_config["task-expires-after"]
     else:
         expires = "28 days"
```
ahal, line 115: Same suggestion as above

ahal (changes requested):
I think I'd prefer a name like `default-task-expiry`, but I see we already have `task-deadline-after`, so this is fine as it's better to be consistent.

ahal:
Also bonus points for adding some tests to `test_transforms_fetch.py` and `test_transforms_docker_image.py` to fix the coverage :)

gabrielBusta:
Thanks! I improved the patch.

I am gonna delete so much code from the translations repo when this lands :)

gabrielBusta:
> I think I'd prefer a name like `default-task-expiry`, but I see we already have `task-deadline-after`, so this is fine as it's better to be consistent.

Yeah, I was going for consistency there.

src/taskgraph/config.py
```diff
@@ -40,6 +40,11 @@
             description="Default 'deadline' for tasks, in relative date format. "
             "Eg: '1 week'",
         ): optionally_keyed_by("project", str),
+        Optional(
+            "task-expires-after",
+            description="Default 'expires-after' for non level 3 tasks, in relative date format. "
+            "Eg: '90 days'",
+        ): str,
```
ahal, line 47: I think saying "non level 3" is technically incorrect because `is_try()` only catches pull requests and projects that contain the string "try". Saying "level 1 tasks" would be more accurate.

That said, it's kind of a shame that this only impacts level 1 tasks. What if we added an `optionally_keyed_by("level", str)` (like how `task-deadline-after` is optionally keyed by project above).

This way the graph config could look like:
```
task-expires-after:
    by-level:
        1: 90 days
        3: 1 year
```

Also, you could just set a single default for all levels:
```
task-expires-after: 1 year
```

Feel free to leave as a follow-up, but would be nice to have an issue on file at least.

ahal (approved):

src/taskgraph/config.py
gabrielBusta: I looked into it a bit and filed #410

PR #414, opened 2024-01-22
Add cache-pull-requests config option to disable caching of tasks on …

jcristau:
…pull requests

Caching tasks from pull requests currently breaks chain-of-trust, because it can't rebuild task definitions from earlier pushes.  This provides an escape hatch for projects which need this.


jcristau:
Updated to also remove the PR specific path from index-search.

ahal (approved):

PR #418, opened 2024-01-29
feat: add helper to walk up a task graph through the Taskcluster API

bhearsum:
This can be useful in places like action tasks where you don't have access to a generated task graph, or where the tasks you're interested in (eg: to help optimize other tasks away) may span over a large number of task groups.

### DIFF
--- a/src/taskgraph/util/taskcluster.py
+++ b/src/taskgraph/util/taskcluster.py
@@ -7,6 +7,7 @@ import datetime
 import functools
 import logging
 import os
+from typing import Dict, List, Union
 
 import requests
 import taskcluster_urls as liburls
@@ -367,3 +368,39 @@ def list_task_group_incomplete_tasks(task_group_id):
             params = {"continuationToken": resp.get("continuationToken")}
         else:
             break
+
+
+def _get_deps(task_ids, use_proxy):
+    upstream_tasks = {}
+    for task_id in task_ids:
+        task_def = get_task_definition(task_id, use_proxy)
+        upstream_tasks[task_def["metadata"]["name"]] = task_id
+
+        upstream_tasks.update(_get_deps(task_def["dependencies"], use_proxy))
+
+    return upstream_tasks
+
+
+def get_ancestors(
+    task_ids: Union[List[str], str], use_proxy: bool = False
+) -> Dict[str, str]:
+    """Gets the ancestor tasks of the given task_ids as a dictionary of label -> taskid.
+
+    Args:
+        task_ids (str or [str]): A single task id or a list of task ids to find the ancestors of.
+        use_proxy (bool): See get_root_url.
+
+    Returns:
+        dict: A dict whose keys are task labels and values are task ids.
+    """
+    upstream_tasks: Dict[str, str] = {}
+
+    if isinstance(task_ids, str):
+        task_ids = [task_ids]
+
+    for task_id in task_ids:
+        task_def = get_task_definition(task_id, use_proxy)
+
+        upstream_tasks.update(_get_deps(task_def["dependencies"], use_proxy))
+
+    return upstream_tasks
--- a/test/test_util_taskcluster.py
+++ b/test/test_util_taskcluster.py
@@ -355,3 +355,150 @@ def test_list_task_group_incomplete_tasks(responses, root_url):
         },
     )
     assert list(tc.list_task_group_incomplete_tasks(tgid)) == ["1", "2", "3"]
+
+
+def test_get_ancestors(responses, root_url):
+    base_url = f"{root_url}/api/queue/v1/task"
+    responses.add(
+        responses.GET,
+        f"{base_url}/fff",
+        json={
+            "dependencies": ["eee", "ddd"],
+            "metadata": {
+                "name": "task-fff",
+            },
+        },
+    )
+    responses.add(
+        responses.GET,
+        f"{base_url}/eee",
+        json={
+            "dependencies": [],
+            "metadata": {
+                "name": "task-eee",
+            },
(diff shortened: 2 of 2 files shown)


src/taskgraph/util/taskcluster.py
```diff
@@ -7,6 +7,7 @@
 import functools
 import logging
 import os
+from typing import Dict, List
```
ahal, line 10: Nice!

src/taskgraph/util/taskcluster.py
```diff
@@ -367,3 +368,22 @@ def list_task_group_incomplete_tasks(task_group_id):
             params = {"continuationToken": resp.get("continuationToken")}
         else:
             break
+
+
+def get_recursive_dependencies(
+    task_ids: List[str], add_given_tasks: bool = False, use_proxy: bool = False
+) -> Dict[str, str]:
+    """Returns a list of all the recursive dependencies of the task_ids provided as
+    a dictionary of a dictionary of label -> taskId."""
```
ahal, line 377: Couple typos here:

* s/Returns a list of all/Returns all
* s/a dictionary of a dictionary/a dictionary

Would also be nice to explain the args in a format like `def state_task` has.

src/taskgraph/util/taskcluster.py
```diff
@@ -367,3 +368,22 @@ def list_task_group_incomplete_tasks(task_group_id):
             params = {"continuationToken": resp.get("continuationToken")}
         else:
             break
+
+
+def get_recursive_dependencies(
```
ahal, line 373: What about `get_ancestors()`?

(feel free to ignore this bikeshed)

src/taskgraph/util/taskcluster.py
```diff
@@ -367,3 +368,22 @@ def list_task_group_incomplete_tasks(task_group_id):
...
+    task_ids: List[str], add_given_tasks: bool = False, use_proxy: bool = False
```
ahal, line 374: Is passing a list of ids here necessary? Could the caller just call this function in a loop instead?

src/taskgraph/util/taskcluster.py
```diff
@@ -367,3 +368,22 @@ def list_task_group_incomplete_tasks(task_group_id):
...
+) -> Dict[str, str]:
+    """Returns a list of all the recursive dependencies of the task_ids provided as
+    a dictionary of a dictionary of label -> taskId."""
+    upstream_tasks: Dict[str, str] = {}
+
+    for task_id in task_ids:
+        task_def = get_task_definition(task_id, use_proxy)
+        if add_given_tasks:
+            upstream_tasks[task_def["metadata"]["name"]] = task_id
```
ahal, line 383: I'm a bit fuzzy on why `add_given_tasks` would be useful. If the caller has the list of ids here already, can't they add these in themselves?

ahal (changes requested):

src/taskgraph/util/taskcluster.py
ahal: Oh I'm silly, it's to make the recursion easier. Makes sense.

ahal: Actually might be nice to support this first argument as a simple string as well (which is probably the common case for the initial call).

src/taskgraph/util/taskcluster.py
bhearsum: Yeah, that's a much better name.

src/taskgraph/util/taskcluster.py
bhearsum: I think it's necessary for this to work, unless we always add the given `task_ids` to the return value. (We have to after we've recursed, otherwise there's no base case, and the return value will always be empty.)

We could just always add them, but then the function becomes `get_ancestor_and_given_task_ids` or something, which I'm not sure is as useful.

The other way to do this is move the recursion out to a separate function:
```
def _get_deps(task_ids, use_proxy):
    upstream_tasks = {}
    for task_id in task_ids:
        task_def = get_task_definition(task_id, use_proxy)
        upstream_tasks[task_def["metadata"]["name"]] = task_id

        upstream_tasks.update(
            _get_deps(task_def["dependencies"], use_proxy)
        )

    return upstream_tasks

def get_recursive_dependencies(
    task_ids: List[str], use_proxy: bool = False
) -> Dict[str, str]:
    """Returns a list of all the recursive dependencies of the task_ids provided as
    a dictionary of a dictionary of label -> taskId."""
    upstream_tasks: Dict[str, str] = {}

    for task_id in task_ids:
        task_def = get_task_definition(task_id, use_proxy)

        upstream_tasks.update(
            _get_deps(task_def["dependencies"], use_proxy)
        )

    return upstream_tasks
```

Which certainly cleans up the interface. As you pointed out, there's really no good use case for `add_given_tasks` in the public interface, so maybe this is just better with no downsides...

src/taskgraph/util/taskcluster.py
bhearsum: That's a good call, yeah.

src/taskgraph/util/taskcluster.py
ahal: Thanks, I missed that this wasn't part of the public interface.. In that case I think this is fine. Maybe you could prefix it with an underscore and leave it out of the docstring (if you do end up documenting args there).

bhearsum: I ended up switching to the above form, which I think reads nicer and improves the interface.

ahal (approved):
Thanks, looks great!

PR #420, opened 2024-01-31
chore: version bump 7.1.2

bhearsum:


ahal:
I think this is technically a feature, but also it has zero chance of impacting existing code so I don't really care :p

ahal (approved):

PR #423, opened 2024-02-02
fix: pick up upstream enhancements for tarfile reproducability (fixes #421)

### LINKED ISSUE #421: pick up upstream archive.py enhancements
AFAICT we are missing at least two enhancements to archive.py that improve determinism:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1347582
- https://phabricator.services.mozilla.com/D165647

I ran into this while trying to reproduce an optimized taskgraph earlier today, and kept ending up with different tarfiles for the docker context, despite the files inside the archive being the same.

bhearsum:
This is mostly copy/pasting from https://searchfox.org/mozilla-central/source/python/mozbuild/mozpack/archive.py, with a couple of tweaks because we don't have the mozpack `file` classes here.
    
The enchancements come from https://bugzilla.mozilla.org/show_bug.cgi?id=1807872 and https://bugzilla.mozilla.org/show_bug.cgi?id=1347582.

### DIFF
--- a/src/taskgraph/util/archive.py
+++ b/src/taskgraph/util/archive.py
@@ -43,7 +43,7 @@ class TarInfo(tarfile.TarInfo):
         info["type"] = HackedType(info["type"])
         # ignore type checking because it looks like pyright complains because we're calling a
         # non-public method
-        return tarfile.TarInfo._create_header(info, format, encoding, errors) # type: ignore
+        return tarfile.TarInfo._create_header(info, format, encoding, errors)  # type: ignore
 
 
 def create_tar_from_files(fp, files):
--- a/src/taskgraph/util/archive.py
+++ b/src/taskgraph/util/archive.py
@@ -12,6 +12,40 @@ import tarfile
 DEFAULT_MTIME = 1451606400
 
 
+# Python 3.9 contains this change:
+#  https://github.com/python/cpython/commit/674935b8caf33e47c78f1b8e197b1b77a04992d2
+# which changes the output of tar creation compared to earlier versions.
+# As this code is used to generate tar files that are meant to be deterministic
+# across versions of python (specifically, it's used as part of computing the hash
+# of docker images, which needs to be identical between CI (which uses python 3.8),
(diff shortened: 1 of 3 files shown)


src/taskgraph/util/archive.py
```diff
@@ -12,6 +12,40 @@
...
+# What we do is subblass TarInfo so that if used on python >= 3.9, it reproduces the
+# behavior from python < 3.9.
+# Here's how it goes:
+# - the behavior in python >= 3.9 is the same as python < 3.9 when the type encoded
+# in the tarinfo is CHRTYPE or BLKTYPE.
+# - the value of the type is only compared in the context of choosing which behavior
+# to take
+# - we replace the type with the same value (so that using the value has no changes)
+# but that pretends to be the same as CHRTYPE so that the condition that enables the
+# old behavior is taken.
```
hneiva, line 32: praise: Nice explanation!

src/taskgraph/util/archive.py
```diff
@@ -12,6 +12,40 @@
...
+        return self == other
+
+
+class TarInfo(tarfile.TarInfo):
+    @staticmethod
+    def _create_header(info, format, encoding, errors):
+        info["type"] = HackedType(info["type"])
+        # ignore type checking because it looks like pyright complains because we're calling a
+        # non-public method
+        return tarfile.TarInfo._create_header(info, format, encoding, errors)  # type: ignore
```
hneiva, line 46: nit: would it still complain if we used `self._create_header`?

hneiva (approved):
Minor nit, but LGTM
Nice hacking!

src/taskgraph/util/archive.py
bhearsum: AIUI, that would end up calling this method recursively.

However, I wonder if `super` might work...

bhearsum: Then again, this is copied from upstream, and I don't think I want to diverge for that more than needed :stuck_out_tongue:

PR #443, opened 2024-02-07
Add support for pr-action in .taskcluster.yml (bug 1641282)

jcristau:
"tasks_for" can now be "pr-action", which runs as level1 and asserts a scope based on the base repo instead of head.


gabrielBusta:
Do we need this bit https://github.com/taskcluster/taskgraph/pull/426/files#diff-061c9948f646c0ab4376bedcd5a6b4781e846f8b0a572b9d29d573493b07ee47R303-R313

to land this?

gabrielBusta:
nvm just saw ur msg on matrix

gabrielBusta (approved):

jcristau:
For the record, landing the .tc.yml changes separately from the code changes, because the former is necessary for updating hooks, which is itself necessary to actually test out the code changes.

PR #447, opened 2024-02-08
actions: support non-indexed decision tasks

jcristau:
The decision task id is the action task's group id, so it's already available without a round-trip through the tc index.  This helps in case of e.g. PRs that don't necessarily index decision tasks (and even if they do, don't index them where `find_decision_task` is currently looking for them).

### DIFF
--- a/src/taskgraph/actions/util.py
+++ b/src/taskgraph/actions/util.py
@@ -32,8 +32,15 @@ def get_parameters(decision_task_id):
     return get_artifact(decision_task_id, "public/parameters.yml")
 
 
-def fetch_graph_and_labels(parameters, graph_config):
-    decision_task_id = find_decision_task(parameters, graph_config)
+def fetch_graph_and_labels(parameters, graph_config, task_group_id=None):
+    try:
+        # Look up the decision_task id in the index
+        decision_task_id = find_decision_task(parameters, graph_config)
+    except KeyError:
+        if not task_group_id:
+            raise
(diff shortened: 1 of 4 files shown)


gabrielBusta:
Might not have to mock the function in [this test](https://github.com/taskcluster/taskgraph/blob/main/test/test_actions_rebuild_cached_tasks.py#L25) anymore

src/taskgraph/actions/util.py
```diff
@@ -32,8 +32,9 @@ def get_parameters(decision_task_id):
     return get_artifact(decision_task_id, "public/parameters.yml")
 
 
-def fetch_graph_and_labels(parameters, graph_config):
-    decision_task_id = find_decision_task(parameters, graph_config)
+def fetch_graph_and_labels(parameters, graph_config, decision_task_id=None):
+    if decision_task_id is None:
+        decision_task_id = find_decision_task(parameters, graph_config)
```
jcristau, line 37: One bit I'm not sure about is whether the `task_group_id` passed to the action is ever `None`, and thus whether we still need this `find_decision_task` fallback?

gabrielBusta: Mmm, good point, I think it's probably always populated

jcristau:
Updated this to still use the index by default: in the case of e.g. release-promotion, an action task can create a new group, and `find_decision_task` will go back and find the on-push decision, whereas using the group id might end up with another action task.
Only if we can't find the decision task in the index do we fall back to the group id.

PR #450, opened 2024-02-08
feat(init): support pr-action in the template .taskcluster.yml

jcristau:


template/{{cookiecutter.project_name}}/taskcluster.github.yml
```diff
@@ -268,7 +278,7 @@ tasks:
 
                           extra:
                               $merge:
-                                  - $if: 'tasks_for == "action"'
+                                  - $if: 'tasks_for in ["action", "pr-action"]'
```
ahal, line 281: It might be worth creating a new `isAction` variable for this, but I'm ok with leaving it too.

ahal (approved):

template/{{cookiecutter.project_name}}/taskcluster.github.yml
jcristau: Thanks for the review.
I did consider a new variable, but there's a few cases where action and pr-action are different, or where they're also the same as cron, so there'd still be a need to distinguish, and leaving it like this might be clearer in the end?

PR #454, opened 2024-02-08
ci: Use docker slim images where possible

hneiva:


bhearsum (approved):
If you want to stick with this approach I'm good with it. I think we also talked about using separate images per Python version to avoid building a bunch of Python versions ourselves in one giant image? Up to you.

hneiva:
> If you want to stick with this approach I'm good with it. I think we also talked about using separate images per Python version to avoid building a bunch of Python versions ourselves in one giant image? Up to you.

Yeah, since it's a bigger refactor I think we should do it in a separate issue.

PR #458, opened 2024-02-08
actions: remove hardcoded gecko trust domain from retrigger-decision action

jcristau:


bhearsum (approved):
Am I missing something? It looks like this is just a formatting change...

PR #465, opened 2024-02-14
Create a base docker image containing only run-task

ahal:


docs/howto/bootstrap-taskgraph.rst
```diff
@@ -4,13 +4,74 @@ Bootstrap Taskgraph in Decision Task
...
+     image: mozillareleases/taskgraph:decision-v7.3.0@sha256:f954e54d427d08b18954a4ff4541d43c76c476d0738f341e4fa82a959939d3fa
+
+The hash can be found on the Docker Hub page for the specific tag you are
+pinning to.
+
+.. _mozillareleases/taskgraph: https://hub.docker.com/r/mozillareleases/taskgraph/tags
+
+
+Using a Python Requirements File
+--------------------------------
```
hneiva, line 45: Nit: Looks like you're using ~~~~ down below, but ----- here, is that intentional?

taskcluster/docker/decision/Dockerfile
```diff
@@ -21,9 +15,5 @@ ADD topsrcdir/requirements/base.txt /setup/requirements.txt
 ADD system-setup.sh /setup/system-setup.sh
 RUN bash /setup/system-setup.sh
```
hneiva, line 16: Isn't this already run in the parent image?

docs/howto/bootstrap-taskgraph.rst
ahal: Yeah, that makes it a smaller heading in rST

taskcluster/docker/decision/Dockerfile
ahal: The two files are different now, the parent one sets up run-task and this one sets up the rest of Taskgraph

hneiva (approved):
Good stuff!

PR #469, opened 2024-02-22
ci: ensure 'push-image' tasks don't overwrite pre-existing tags

ahal:


ahal:
The failing `push-image` tasks demonstrate this PR is working as expected (they were triggered via a release event, but on the same revision so are showing up here I guess).

hneiva (approved):

PR #475, opened 2024-03-21
Ignore type check errors on a line by line basis

ahal:
This will catch more type checking errors than if we ignore the entire file.


bhearsum (approved):
It looks like the test failure here is a strange pip issue? All of our tests pass. I'm in support of this change.

ahal:
> It looks like the test failure here is a strange pip issue? All of our tests pass. I'm in support of this change.

Yeah, I'm not sure what's going on here. I can't reproduce it locally. It's definitely a permafail and caused by this PR though :/

ahal:
I decided to spin the dep updates into a separate PR and hit the same issue:
https://github.com/taskcluster/taskgraph/pull/518

I'll figure out what is causing the problem and pin accordingly.

PR #477, opened 2024-03-21
feat: memoize get_task_definition and _get_deps

bhearsum:
This greatly improves performance of get_ancestors for certain inputs.

### DIFF
--- a/src/taskgraph/util/taskcluster.py
+++ b/src/taskgraph/util/taskcluster.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
+import copy
 import datetime
 import functools
 import logging
@@ -247,6 +248,7 @@ def get_task_url(task_id, use_proxy=False):
     return task_tmpl.format(task_id)
 
 
+@memoize
 def get_task_definition(task_id, use_proxy=False):
     response = _do_request(get_task_url(task_id, use_proxy))
     return response.json()
@@ -372,13 +374,14 @@ def list_task_group_incomplete_tasks(task_group_id):
             break
 
 
+@memoize
 def _get_deps(task_ids, use_proxy):
     upstream_tasks = {}
     for task_id in task_ids:
         task_def = get_task_definition(task_id, use_proxy)
         upstream_tasks[task_def["metadata"]["name"]] = task_id
 
-        upstream_tasks.update(_get_deps(task_def["dependencies"], use_proxy))
+        upstream_tasks.update(_get_deps(tuple(task_def["dependencies"]), use_proxy))
 
     return upstream_tasks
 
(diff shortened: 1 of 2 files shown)


jcristau (commented):
The downside here is that memoizing mutable objects is a bit of a footgun, because if any of the callers modifies the returned object, that pollutes the cache.
I don't know whether / how much that is a concern here, but it's bit us before.

bhearsum:
> The downside here is that memoizing mutable objects is a bit of a footgun, because if any of the callers modifies the returned object, that pollutes the cache. I don't know whether / how much that is a concern here, but it's bit us before.

Yeah, that's a good point. Perhaps only memoizing `get_task_definition` would be preferable, as it's the only thing making any network requests. Alternatively, accepting tuples instead of lists would make it impossible to change the argument after it has been passed...

bhearsum:
Updated patch just memoizes `get_task_definition`. This does mean we will only fetch it once ever per taskgraph invocation. I...think this is OK, especially given the other things we memoize in taskgraph, but if this is of concern we could have `get_ancestors` and `_get_deps` use their own cache instead to avoid interfering with other uses of `get_task_definition`.

bhearsum:
> Updated patch just memoizes `get_task_definition`. This does mean we will only fetch it once ever per taskgraph invocation. I...think this is OK, especially given the other things we memoize in taskgraph, but if this is of concern we could have `get_ancestors` and `_get_deps` use their own cache instead to avoid interfering with other uses of `get_task_definition`.

@ahal - you may have thoughts on this too?

ahal:
Should `@memoize` return a copy of the object instead? Then again I wouldn't be surprised if there are places where we are relying on the mutable behaviour.

This seems pretty safe to me, we don't typically mutate an already submitted task definition.

ahal (approved):

bhearsum:
> Should `@memoize` return a copy of the object instead? Then again I wouldn't be surprised if there are places where we are relying on the mutable behaviour.
> 
> This seems pretty safe to me, we don't typically mutate an already submitted task definition.

If the scenario we're talking about is callers relying on multiple calls to `get_task_definition` returning the exact same object, and that it can be mutated and shared between those callers, I'd argue that's undefined behaviour that shouldn't be relied on anyways :stuck_out_tongue:. I could be convinced either way on this, but I'm tempted to leave it as is to avoid the extra copy operation.

jcristau:
The scenario I think is more realistic is one where we call `get_task_definition` and modify the object e.g. to submit a new task, possibly in a loop, as part of something like an action task.  (I'm not actually seeing that pattern anywhere right now, from a quick look.)

bhearsum:
> The scenario I think is more realistic is one where we call `get_task_definition` and modify the object e.g. to submit a new task, possibly in a loop, as part of something like an action task. (I'm not actually seeing that pattern anywhere right now, from a quick look.)

Right, OK. I guess the question is whether or not the current behaviour is a fair expectation by the caller. I don't really have a strong opinion here though...it seems like if we want to avoid risk, adding the copy would be good. Otherwise, we could always adjust any callers that break when they pick up this change (I think?).

PR #484, opened 2024-04-04
feat: more featureful set-name for from-deps

bhearsum:
This is useful in cases where many tasks are only distinguished by a common suffix, and have a common task downstream of them.

As a concrete example, in Translations, tasks are named after their kind + the language pair they're running against, and I'm working on getting beetmover running against all of them. At the moment, `from-deps` generate names like `beetmover-ru-en` for every single task, causing collisions.

Note that we already do this sort of naming for things like beetmover and signing tasks in Firefox, where we end up with task names like `beetmover-repackage-macosx64-shippable/opt` and `shippable-l10n-mac-signing-macosx64-shippable-17/opt`. (We don't use `from-deps` there, but this or similar support is likely to be needed to every move to it.)

### DIFF
--- a/src/taskgraph/transforms/from_deps.py
+++ b/src/taskgraph/transforms/from_deps.py
@@ -20,6 +20,8 @@ from taskgraph.transforms.run import fetches_schema
 from taskgraph.util.attributes import attrmatch
 from taskgraph.util.dependencies import GROUP_BY_MAP, get_dependencies
 from taskgraph.util.schema import Schema, validate_schema
+from taskgraph.util.set_name import SET_NAME_MAP
+
 
 FROM_DEPS_SCHEMA = Schema(
     {
@@ -41,12 +43,14 @@ FROM_DEPS_SCHEMA = Schema(
                 "set-name",
                 description=dedent(
                     """
-                When True, `from_deps` will derive a name for the generated
-                tasks from the name of the primary dependency. Defaults to
-                True.
+                UPDATE ME AND DOCS
                 """.lstrip()
                 ),
-            ): bool,
+            ): Any(
+                None,
+                *SET_NAME_MAP,
+                {Any(*SET_NAME_MAP): object},
+            ),
             Optional(
                 "with-attributes",
                 description=dedent(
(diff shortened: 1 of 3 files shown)


ahal (changes requested):
How would you feel about implementing the ability to specify registered name functions similar to how we handle `group-by`? I worry that we aren't going to find a sane set of options here that are going to work for every conceivable case.

I'm picturing we would change `set-name` to take a string that references a registered function (see `group-by`). The `default` function would do the current behaviour and we could provide another built-in function that doesn't strip out the kind.

Another option I like is removing `set-name` entirely from `from_deps`, and then creating a builtin `taskgraph.transforms.set_name` transform that can be run afterwards.

Note that I have a couple PRs that break backward compat teed up already, so if we time this right we don't have to worry about making a major version bump just for this.

src/taskgraph/transforms/from_deps.py
```diff
@@ -41,12 +70,14 @@
...
-                tasks from the name of the primary dependency. Defaults to
-                True.
+                UPDATE ME AND DOCS
                 """.lstrip()
                 ),
-            ): bool,
+            ): Any(
+                None,
+                *SET_NAME_MAP,
+                {Any(*SET_NAME_MAP): object},
```
ahal, line 79: This is going to break the ability to create custom `set_name` functions because the Schema is created at import time which happens before said custom function will have a change to register themselves. That's why I ultimately moved the `group_by` function off into a utility file. That way we have a chance to register functions before the schema gets created.

Alternatively we could just allow any string rather than the actually registered ones. That would be fine with me.

ahal (changes requested):
Thanks, this looks just like I was thinking!

src/taskgraph/transforms/from_deps.py
bhearsum: Ah, good point. It's pretty easy to just move this out elsewhere, and matches what we're doing for other things. Let's do that.

I am curious though: is there a reason not to just accept a callable here instead?

ahal: Like the python path notation? Yeah that would work too.

For `group-by` I was just copy/pasting how the existing `multi_dep` loader did it. One benefit of this way is that consumers don't need to care about where the function lives, e.g `set-name: default` vs `set-name: taskgraph.transforms.from_deps.default`. It looks cleaner.

ahal (approved):
Thanks!

PR #485, opened 2024-04-04
fix: support 'artifact-reference' with private artifacts

ahal:
Issue: #90


jcristau (approved):
Thanks, this looks plausible to me.

PR #486, opened 2024-04-04
refactor!: remove 'head_ref' from cached_task PR route

ahal:
BREAKING CHANGE: changed cached_task pull request index route format


ahal:
This depends on:
https://phabricator.services.mozilla.com/D206701

bhearsum (approved):
This looks reasonable. AFAICT, we'll need to keep the `cache.head` grants in ci-config until all relevant repositories have upgraded to a version of taskgraph with this patch?

(I assume tests will pass once the ci-config change is landed.)

PR #494, opened 2024-04-24
fix: pass in base + head revisions when computing files_changed

ahal:


src/taskgraph/decision.py
```diff
@@ -185,7 +185,9 @@ def get_decision_parameters(graph_config, options):
 
     # Define default filter list, as most configurations shouldn't need
     # custom filters.
-    parameters["files_changed"] = repo.get_changed_files("AM")
+    parameters["files_changed"] = repo.get_changed_files(
```
bhearsum, line 188: To be clear: we need rev/base_rev because otherwise [we default to the previous revision](https://github.com/taskcluster/taskgraph/blob/29f5d601b6bb639e54ba7ded8a3d013f68706a92/src/taskgraph/util/vcs.py#L442-L444)? (And something else for hg...)

bhearsum (approved):

src/taskgraph/decision.py
ahal: Yes, also this is what the `files_changed.py` module was doing before my refactor.. I just totally missed the difference

ahal: Oh and technically we were in the top half of that conditional, so we were detecting the base_rev automatically. The detection works great for pushes, but not for pull requests.

PR #496, opened 2024-04-29
feat: use new batched queries in task replacement

Alphare:
Batching queries to get from task index to status reduces the number of (sometimes trans-continental) queries from 2*900+ to ~2. This reduces the time spent replacing tasks from 20% to 75% depending on the use-case.

The 20% improvement in wall time was observed when running `mach taskgraph morphed` in a CI worker, while the 75% improvement was observed in a developer machine in France running `mach taskgraph full`.

More information in https://github.com/taskcluster/taskcluster-rfcs/pull/189.

### DIFF
--- a/src/taskgraph/optimize/base.py
+++ b/src/taskgraph/optimize/base.py
@@ -22,6 +22,7 @@ from taskgraph.graph import Graph
 from taskgraph.taskgraph import TaskGraph
 from taskgraph.util.parameterization import resolve_task_references, resolve_timestamps
 from taskgraph.util.python_path import import_sibling_modules
+from taskgraph.util.taskcluster import find_task_id_batched, status_task_batched
 
 logger = logging.getLogger(__name__)
 registry = {}
@@ -51,6 +52,9 @@ def optimize_task_graph(
     Perform task optimization, returning a taskgraph and a map from label to
     assigned taskId, including replacement tasks.
     """
+    # avoid circular import
+    from taskgraph.optimize.strategies import IndexSearch
+
     label_to_taskid = {}
     if not existing_tasks:
         existing_tasks = {}
@@ -70,6 +74,23 @@ def optimize_task_graph(
         do_not_optimize=do_not_optimize,
     )
 
+    # Gather each relevant task's index
+    indexes = set()
+    for label in target_task_graph.graph.visit_postorder():
+        if label in do_not_optimize:
+            continue
+        _, strategy, arg = optimizations(label)
+        if isinstance(strategy, IndexSearch) and arg is not None:
+            indexes.update(arg)
+
+    index_to_taskid = {}
+    taskid_to_status = {}
+    if indexes:
+        # Find their respective status using TC index/queue batch APIs
+        indexes = list(indexes)
+        index_to_taskid = find_task_id_batched(indexes)
+        taskid_to_status = status_task_batched(list(index_to_taskid.values()))
+
     replaced_tasks = replace_tasks(
         target_task_graph=target_task_graph,
         optimizations=optimizations,
@@ -78,6 +99,8 @@ def optimize_task_graph(
         label_to_taskid=label_to_taskid,
         existing_tasks=existing_tasks,
         removed_tasks=removed_tasks,
+        index_to_taskid=index_to_taskid,
+        taskid_to_status=taskid_to_status,
     )
 
     return (
@@ -259,12 +282,17 @@ def replace_tasks(
     label_to_taskid,
     removed_tasks,
     existing_tasks,
+    index_to_taskid,
+    taskid_to_status,
 ):
     """
     Implement the "Replacing Tasks" phase, returning a set of task labels of
     all replaced tasks. The replacement taskIds are added to label_to_taskid as
     a side-effect.
     """
+    # avoid circular import
+    from taskgraph.optimize.strategies import IndexSearch
+
     opt_counts = defaultdict(int)
     replaced = set()
     dependents_of = target_task_graph.graph.reverse_links_dict()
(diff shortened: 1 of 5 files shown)


ahal:
I have no idea what's up with those task failures, never seen that before..

jcristau:
Wow this decision task generated quite a messed up graph. Things end up using the index-task docker image instead of whatever docker image they wanted. [full-task-graph.json](https://firefoxci.taskcluster-artifacts.net/BrnPPIkTTPaSKUvmd4L1qw/0/public/full-task-graph.json) looks sane, but [task-graph.json](https://firefoxci.taskcluster-artifacts.net/BrnPPIkTTPaSKUvmd4L1qw/0/public/task-graph.json) has [PebYGdYSQgu0_iTKTCg_fQ](https://firefox-ci-tc.services.mozilla.com/tasks/PebYGdYSQgu0_iTKTCg_fQ) as dependency where it shouldn't be.

Alphare:
I *think* I didn't have this problem when I hadn't yet changed the tests (to check if the local failures were the same). Does that give any hints as to what I should do @jcristau ?

Alphare:
Tests are fixed, I hadn't fixed a problem in both places it was and the result was confusing to me. Everything seems good now.

Alphare:
Tests are fixed, I hadn't fixed a problem in both places it was and the result was confusing to me. Everything seems good now.

src/taskgraph/optimize/base.py
```diff
@@ -307,7 +328,9 @@ def replace_tasks(
             deadline = max(
                 resolve_timestamps(now, task.task["deadline"]) for task in dependents
             )
-        repl = opt.should_replace_task(task, params, deadline, arg)
+        repl = opt.should_replace_task(
+            task, params, deadline, arg, label_to_taskid, taskid_to_status
+        )
```
ahal, line 333: This will break custom optimization strategies defined in project specific code.

That in and of itself isn't a huge deal (we can do a major version bump if necessary).. but these values are specific to the `IndexSearch` strategy, so I don't love that we would pass them in to every strategy.

I'm not sure what the best alternative here is however..

There is a way to pass arguments into the strategies' constructor, but that happens in the decorator at module import time, so we wouldn't have `target_task_graph` yet.

The other option could be to pass the necessary data in via classmethod / classproperty. E.g, we could set `target_task_graph` as a class property on all strategies, then `IndexSearch` can use that to lazily make the calls to Taskcluster (and memoize the result). That feels a bit cleaner to me.

ahal (changes requested):
Thanks, looking good! This needs a bit of refactoring though

src/taskgraph/optimize/base.py
```diff
@@ -70,6 +71,24 @@ def optimize_task_graph(
         do_not_optimize=do_not_optimize,
     )
 
+    # Gather each relevant task's index
+    indexes = set()
+    for label in target_task_graph.graph.visit_postorder():
+        if label in do_not_optimize:
+            continue
+        index = optimizations(label)[2]
+        if index is not None:
```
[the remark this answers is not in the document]
ahal: Actually my suggestion further down of moving this logic to the `IndexSearch` class itself plays nicely here. That way we can avoid having `IndexSearch` specific logic in the core optimization code.

src/taskgraph/optimize/base.py
Alphare: I agree that this is suboptimal. I don't really understand your last paragraph; why would we use a `classmethod/classproperty` specifically and why would that help to pass down the arguments?

I've pushed a hacky version of passing conditionally the arguments so we can clear the rest of the pending comments.

src/taskgraph/optimize/base.py
```diff
@@ -307,7 +335,13 @@ def replace_tasks(
...
                 resolve_timestamps(now, task.task["deadline"]) for task in dependents
             )
-        repl = opt.should_replace_task(task, params, deadline, arg)
+        if isinstance(opt, IndexSearch):
+            repl = opt.should_replace_task(
+                task, params, deadline, arg, index_to_taskid, taskid_to_status
+            )
+        else:
+            repl = opt.should_replace_task(task, params, deadline, arg)
+
```
ahal, line 344: I'd prefer to leave the interface unchanged here, so I suggest you do something like:
```
if isinstance(opt, IndexSearch):
    arg = (arg, index_to_task_id, taskid_to_status)

repl = opt.should_replace_task(task, params, deadline, arg)
```

Then unpack `arg` in the `IndexSearch` strategy correspondingly. We use a similar technique in a few other places so this would be a bit more consistent.

src/taskgraph/optimize/base.py
```diff
@@ -423,7 +457,7 @@ def should_remove_task(self, task, params, arg):
         True to remove."""
         return False
 
-    def should_replace_task(self, task, params, deadline, arg):
+    def should_replace_task(self, *args, **kwargs):
```
ahal, line 460: This can be left alone with above comment fixed.

src/taskgraph/util/taskcluster.py
```diff
@@ -193,6 +193,35 @@ def find_task_id(index_path, use_proxy=False):
     return response.json()["taskId"]
 
 
+def find_task_id_batched(index_paths, use_proxy=False):
```
ahal, line 196: Would be nice to include the docstring like you did for `status_task_batched` :)

ahal (changes requested):
Thanks, this looks correct. I have one minor refactor request, then I think this can merge. I'll handle the merge + new release after the next iteration.

Alphare:
@ahal Thanks, all of your comments should be addressed now.

ahal (approved):
Awesome, lgtm!

PR #498, opened 2024-05-02
Add debug logging for decision tasks

jcristau:


bhearsum (approved):
I'm tempted to say this should be the default :stuck_out_tongue:. Easy enough to enable where we want it though.

PR #506, opened 2024-05-15
fix: migrate over the last untouched `IndexSearch.should_replace_task`

Alphare:
This was forgotten in 5207917d69793784b9556716fc0a7279d26bd23a. Now all usages of this API in this repo have been migrated.


Alphare:
This is not run through this CI (I did a broken push first and it passed), but you can check it running with `mach taskcluster-load-image <IMAGE>` with `<IMAGE>` being a valid image like `updatebot`.

Alphare:
@ahal Here is the docker fix I was talking about earlier.

ahal (approved):
Thanks!

PR #507, opened 2024-05-15
Merge Gecko image_builder (with all history) into Taskgraph

ahal:
I'd like to start building and deploying the `image_builder` from Taskgraph. This PR is just focusing on importing the files along with their history, and standing up a `docker-image` task to build the image on push.

A future PR will handle release / deployment / bootstrapping of this image. I haven't fully figured out how that will work yet.


ahal:
I just noticed that there was a move from `testing/docker/image_builder` to `taskcluster/docker/image_builder` 8 years ago.. So the history isn't complete.

I'm inclined to think this is good enough, but if anyone feels strongly about it then lmk and I'll see if I can figure out how to include that too.

hneiva:
> I just noticed that there was a move from `testing/docker/image_builder` to `taskcluster/docker/image_builder` 8 years ago.. So the history isn't complete.
> 
> I'm inclined to think this is good enough, but if anyone feels strongly about it then lmk and I'll see if I can figure out how to include that too.

Agreed. The history still lives on the old repo if we really need to dig for it.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -21,6 +21,8 @@ tasks:
         parent: run-task
     fetch:
         symbol: I(fetch)
+    image-builder:
```
hneiva, line 24: The task name becomes `build-docker-image-image-builder`. Any way we could make the name a bit more sensible?

hneiva (approved):
just one question, but LGTM.
Thanks for porting this over! :)

taskcluster/kinds/docker-image/kind.yml
ahal: Yeah it's weird, but I'd vote we leave it. Otherwise we no longer follow the `<kind>-<name>` format, which I'd argue is even more confusing.

ahal:
> Agreed. The history still lives on the old repo if we really need to dig for it.

Actually this ended up being pretty easy thanks to `git filter-repo`. Just re-pushed and now we have the full history.

bhearsum (approved):
Thank you for making the effort to import history.

PR #508, opened 2024-05-17
fix: Add logging when a cached task cannot be found (bug 1891183)

JohanLorenzo:
Here's what the log lines are going to look like[1]:

```
[task 2024-05-17T09:32:15.638Z] 2024-05-17 09:32:15,638 - DEBUG - replace_tasks: build-docker-image-decision
[task 2024-05-17T09:32:15.638Z] 2024-05-17 09:32:15,638 - WARNING - Couldn't find cached task for index path "taskgraph.cache.level-3.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a". Reason: <class 'KeyError'> 'taskgraph.cache.level-3.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a'
[task 2024-05-17T09:32:15.639Z] 2024-05-17 09:32:15,639 - WARNING - Couldn't find cached task for index path "taskgraph.cache.level-2.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a". Reason: <class 'KeyError'> 'taskgraph.cache.level-2.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a'
[task 2024-05-17T09:32:15.639Z] 2024-05-17 09:32:15,639 - WARNING - Couldn't find cached task for index path "taskgraph.cache.level-1.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a". Reason: <class 'KeyError'> 'taskgraph.cache.level-1.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a'
[task 2024-05-17T09:32:15.639Z] 2024-05-17 09:32:15,639 - WARNING - Couldn't find cached task for index path "taskgraph.cache.pr.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a". Reason: <class 'KeyError'> 'taskgraph.cache.pr.docker-images.v2.decision.hash.6089794e1aa2d0d0faf72ef49150dbd56a81f50c7bef0e39e0d8e4c0050f4b6a'
[task 2024-05-17T09:32:15.639Z] 2024-05-17 09:32:15,639 - DEBUG - replace_tasks: build-docker-image-decision kept by optimization strategy
```

I'm lucky my PR needs to build a docker image. I'll keep canceling this task until we agree on what the message should output. That's the reason why I'm sending a red PR for review 🙂 

[1] https://firefox-ci-tc.services.mozilla.com/tasks/Z6ar04cXSn6ZIJIVGpvC4w/runs/0/logs/public/logs/live.log#L207

### DIFF
--- a/src/taskgraph/optimize/strategies.py
+++ b/src/taskgraph/optimize/strategies.py
@@ -40,8 +40,9 @@ class IndexSearch(OptimizationStrategy):
         for index_path in index_paths:
             try:
                 if batched:
-                    task_id = label_to_taskid[index_path]
-                    status = taskid_to_status[task_id]
+                    task_id, status = _find_task_id_and_status_from_batch(
+                        label_to_taskid, taskid_to_status, index_path
+                    )
                 else:
                     # 404 is raised as `KeyError` also end up here
                     task_id = find_task_id(index_path)
@@ -57,11 +58,27 @@ class IndexSearch(OptimizationStrategy):
                     continue
 
                 return task_id
-            except KeyError:
-                # go on to the next index path
-                pass
+            except KeyError as e:
+                logger.debug(e)
 
         return False
+
+
+def _find_task_id_and_status_from_batch(label_to_taskid, taskid_to_status, index_path):
+    try:
+        task_id = label_to_taskid[index_path]
+    except KeyError as e:
+        raise KeyError(
+            f'Couldn\'t find cached task for index path "{index_path}".'
+        ) from e
+    try:
+        status = taskid_to_status[task_id]
+    except KeyError as e:
+        raise KeyError(
+            f'Couldn\'t find cached status for taskId {task_id} (index path "{index_path}").'
+        ) from e
+
+    return task_id, status
 
 
 @register_strategy("skip-unless-changed")


src/taskgraph/optimize/strategies.py
```diff
@@ -40,9 +40,10 @@ def should_replace_task(self, task, params, deadline, arg):
                     continue
 
                 return task_id
-            except KeyError:
-                # go on to the next index path
-                pass
+            except KeyError as e:
+                logger.warning(
+                    f'Couldn\'t find cached task for index path "{index_path}". Reason: {type(e)} {e}'
```
JohanLorenzo, line 45: In most cases, the `index_path` is going to be output twice (see example in my first comment). That's because a `KeyError` just contains the key as the error message. That said this line can also throw a `KeyError`:

https://github.com/taskcluster/taskgraph/blob/41a10aaf10dbe5cfe3c5cb3fc3622bcdce8c66f5/src/taskgraph/optimize/strategies.py#L31

So, I'm tempted to keep outputting both `{index_path}` and `{e}`. 

Let me know what you think!

src/taskgraph/optimize/strategies.py
```diff
@@ -40,9 +40,10 @@ def should_replace_task(self, task, params, deadline, arg):
                     continue
 
                 return task_id
-            except KeyError:
-                # go on to the next index path
-                pass
+            except KeyError as e:
+                logger.warning(
```
ahal, line 44: This should definitely be a `debug` level, if not `trace` :)

src/taskgraph/optimize/strategies.py
ahal: I would prefer to leave the exception out of the log. If it's important to have a different error message depending on where the `KeyError` comes from, then let's do two separate `try/except` blocks with different messages.

ahal (changes requested):
Nice, this is going to be useful!

I'm a little worried about the verbosity of the log however, it's already pretty massive thanks to all the optimization logging and this looks like it will make it way way bigger.

What if we move all optimization phase logging to a separate `optimization.log` artifact? It would keep the main log nice and compact, but give us the chance to dive into optimization decision in depth when needed.

ahal:
Another misc thought unrelated to this patch:
It would be nice if the logs for a given task had the label as a standard prefix, e.g: `[builld-linux-64/opt] kept due to ...`. Then you can grep the log for that label tag to see all relevant log lines just for that task.

This is a separate issue from this one though :)

src/taskgraph/optimize/strategies.py
JohanLorenzo: Fixed

src/taskgraph/optimize/strategies.py
JohanLorenzo: Fixed by raising multiple `KeyError`s and catching them as one single place

bhearsum:
Superceded by #525, which includes a separate optimization log.

PR #509, opened 2024-05-21
feat: allow for both batched and single queries in task replacement

Alphare:
This patch removes the need for migrating all call sites, especially in `gecko`.

This is the result of more discussion following the two previous patches¹. Turns out that while duck typing in interfaces is not great, leaking implementation details to the user is probably worse.

[1] 41a10aaf10dbe5cfe3c5cb3fc3622bcdce8c66f5
[1] 5207917d69793784b9556716fc0a7279d26bd23a

### DIFF
--- a/src/taskgraph/docker.py
+++ b/src/taskgraph/docker.py
@@ -17,10 +17,8 @@ except ImportError as e:
 
 from taskgraph.util import docker
 from taskgraph.util.taskcluster import (
-    find_task_id_batched,
     get_artifact_url,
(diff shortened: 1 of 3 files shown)


ahal (changes requested):
Minus the task failures :)

ahal:
Not sure what the `image-decision` failure was about.. looks like possible intermittent infra issue. Let's see if a re-run is better.

ahal (approved):

PR #511, opened 2024-05-22
Permit relative script paths in toolchain jobs

oskirby:
I encountered a minor annoyance while writing a toolchain job that tries to use a relative path like this:

```
flatpak-sdks:
    description: "Flatpak Bundle SDKs"
    treeherder:
        symbol: TL(flatpak-sdks)
    run:
        use-caches: true
        script: ../podman-worker.py
        arguments:
            - '/root/bundle_sdk.sh'
        toolchain-alias: flatpak-sdks
        toolchain-artifact: public/build/flatpak-sdks.tar.xz
```

This results in an error when trying to compute the file hashes, because it winds up failing to match `taskcluster/scripts/toolchain/../podman-worker.py` against the globbed path of `taskcluster/scripts/podman-worker.py` to work around the path we should run `path.normpath()` on the script before passing it into `hash_paths()`

### DIFF
--- a/src/taskgraph/transforms/run/toolchain.py
+++ b/src/taskgraph/transforms/run/toolchain.py
@@ -15,6 +15,7 @@ from taskgraph.transforms.run.common import (
     get_vcsdir_name,
 )
 from taskgraph.util.hash import hash_paths
+from taskgraph.util import path as mozpath
 from taskgraph.util.schema import Schema
 from taskgraph.util.shell import quote as shell_quote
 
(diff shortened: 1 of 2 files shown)


ahal (approved):
Makes sense to me, thanks for the test!

ahal:
I put some thought into whether it would be possible to use `..` to invoke a binary outside of the task's workdir, but we run `normpath` *and then* join it to `{srcdir}`. So no matter what the resulting path will always be something in-repo.

ahal:
Ah right, I forgot normpath doesn't do that. And yes it's totally possible to just do this in the script so I agree there's no additional security concern here.

I made that comment as a quick justification for not putting more thought into it. It was wrong, but there's still no issue here.

Issue #504, opened 2024-05-10
[run-task] Add an option to select specific submodules to update


ahal:
Currently we just update all submodules we detect, but these are likely not all needed by every task. It would be nice if tasks could pass a list of submodules they need (or disable them entirely) so we can speed up cloning times.

This is particularly painful for `mozilla-vpn-client` which has a bunch of submodules that take a good while to clone and are hardly ever needed.

ahal:
We'll have to take care with caches here. I.e, if the checkout is cached we'll need to ensure the appropriate submodules have been initialized.

PR #512, opened 2024-05-24
[run-task] Add the ability to configure submodule checkouts.

oskirby:
Thi Implements #504 and adds a bunch of tests to verify the behaviour of `run-task` when it comes to handling checkouts and their git submodules.

The first major change is to `run-task` which now checks for an environment variable of the form `<reponame>_SUBMODULES` to control how to handle git submodules. This variable can take the following values:
- unset: No submodules are checked out.
- `true`, `auto` or `yes`: All submodules are checked out.
- otherwise, a list of submodule paths to checkout, delimited by colons.

The second set of changes addresses the run-task transforms so that this behaviour can be controlled by a YAML task description. In this stream of work we extend the `RepoConfig` class to include an optional `submodules` value, which can take one of the following values:
- `None` or `False`: No submodules are checked out.
- `True`: All submodules are checked out.
- Otherwise, a list of strings naming the submodules to check out.

Note! This changes the default behavior of tasks! Submodules are *not* checked out by default as a result of this change. But I figured I would write it in the most sensible way first and then grapple with the challenges of how to handle compatibility.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -598,6 +598,7 @@ def git_checkout(
     commit: Optional[str],
     ssh_key_file: Optional[Path],
     ssh_known_hosts_file: Optional[Path],
+    submodules: Optional[str | bool],
 ):
     env = {
         # abort if transfer speed is lower than 1kB/s for 1 minute
@@ -705,23 +706,40 @@ def git_checkout(
 
     run_required_command(b"vcs", args, cwd=destination_path)
 
-    if os.path.exists(os.path.join(destination_path, ".gitmodules")):
-        args = [
-            "git",
-            "submodule",
-            "init",
+    if submodules is not None:
+        def _filter_submodule(status):
+            # A boolean value selects/disables all submodules.
+            if isinstance(submodules, bool):
+                return submodules
+
+            # Otherwise, a colon-separated stringlist of selected submodules.
+            smpath = status.split()[1]
+            return smpath in submodules.split(':')
+
+        submodpaths = [
+            status.split()[1]
+            for status in subprocess.check_output(["git", "submodule", "status"],
+                cwd=destination_path, universal_newlines=True).splitlines()
+            if _filter_submodule(status)
         ]
 
-        run_required_command(b"vcs", args, cwd=destination_path)
+        for p in submodpaths:
+            args = [
+                "git",
+                "submodule",
+                "init",
+                p
+            ]
+            run_required_command(b"vcs", args, cwd=destination_path)
 
-        args = [
-            "git",
-            "submodule",
-            "update",
-            "--force",  # Overrides any potential local changes
-        ]
-
-        run_required_command(b"vcs", args, cwd=destination_path)
+            args = [
+                "git",
+                "submodule",
+                "update",
+                "--force",  # Overrides any potential local changes
+                p
+            ]
+            run_required_command(b"vcs", args, cwd=destination_path)
 
     _clean_git_checkout(destination_path)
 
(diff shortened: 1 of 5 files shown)


test/test_scripts_run_task.py
```diff
@@ -346,6 +462,7 @@ def test_git_checkout(
             commit=None,
             ssh_key_file=None,
             ssh_known_hosts_file=None,
+            submodules=submodules,
         )
 
         # Check desired files exist
```
bhearsum, line 468: There's an assert further down to check whether the outer repo is at the right revision. There's an assert checking for expected files makes sure _something_ was checked out for the submodules, which I guess is good enough here, since the submodules themselves control which revs are used?

src/taskgraph/run-task/run-task
```diff
@@ -705,23 +706,40 @@ def git_checkout(
...
+            if _filter_submodule(status)
         ]
 
-        run_required_command(b"vcs", args, cwd=destination_path)
+        for p in submodpaths:
+            args = [
+                "git",
+                "submodule",
+                "init",
+                p
```
bhearsum, line 731: Checkouts are often cached on individual workers. Will this work correctly if a checkout already contains an initialized submodule at the same (or different) revision?

bhearsum (commented):
Thanks for writing this!

I think it largely looks good. There's an important open question further down about caches/existing checkouts that I'd like to clarify before approving this, and one other less important thing.

> true, auto or yes: All submodules are checked out.

TIL that `auto` is boolean True in python.

> Note! This changes the default behavior of tasks! Submodules are NOT checked out by default as a result of this change. But I figured I would write it in the most sensible way first and then grapple with the challenges of how to handle compatibility.

I could go either way on what the default behaviour should be, but I think what you have here is fine. I'm not too stressed about compat - it's a simple enough change for consumers to make when they upgrade, and few enough are using submodules anyways.

src/taskgraph/run-task/run-task
ahal: I think it would be technically possible to have different submodules checked out at different revisions, but technically a task should only rely on the submodules that it declares, and those would all get the `git submodule update` treatment below.

That being said, I'm starting to think we should default to updating all submodules if `SUBMODULES` is not set. I don't think it's going to be easy to audit tasks to tell whether / which submodules they may or may not be using. So preserving backwards compatibility by defaulting to all of them will probably save us some headache.

ahal (changes requested):
I think conceptually I do prefer this default behaviour of not checking out any submodules unless `SUBMODULES` is specified..

But thinking about it, this might cause us some headaches as it may be difficult to determine which tasks depend on what submodules as we upgrade Taskgraph places.

Can we refactor things so that not specifying the env preserves backwards compatibility?

test/test_scripts_run_task.py
oskirby: That is correct, a submodule in the superproject specifies the exact hash of child project to be checked out, and the `git submodule update --force` flag will ensure that the revision specified by the superproject is checked out.

src/taskgraph/run-task/run-task
oskirby: It will take a little bit of rethinking how what the env variables should mean to properly convey that (default disabled is easy because an unset environment variable is read as an empty string), but I will see what I can do.

oskirby:
Sorry for the delay in getting back to you, I got distracted by PTO :)
I'll try and revise this work so that the default behavior checks out submodules by default.

oskirby:
Going to close this as stale

PR #514, opened 2024-06-05
fix: chain of trust should influence cache digest

bhearsum:
Chain of trust being enabled or disabled influences the outcome of a task. Specifically, it controls whether or not chain of trust artifacts will be present on the task. Without this impacting the cache digest, cached tasks will continue to be cached when chain of trust is enabled on them, and if there is a chain of trust verifier downstream, this will cause verification issues.

BREAKING CHANGE: worker.chain-of-trust now influences cache digests

### DIFF
--- a/src/taskgraph/transforms/cached_tasks.py
+++ b/src/taskgraph/transforms/cached_tasks.py
@@ -77,7 +77,14 @@ def cache_task(config, tasks):
                         task["label"], p
                     )
                 )
+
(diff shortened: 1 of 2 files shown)


gabrielBusta:
I don't fully understand this one. So I am deferring to someone else in @taskcluster/releng

jcristau (approved):

ahal:
I'm not sure this should be considered a breaking change.. It doesn't require any code updates in consumer repos right? It'll just invalidate their caches.

PR #515, opened 2024-06-06
BREAKING CHANGE: Delete taskgraph.util.memoize (fixes #491)

### LINKED ISSUE #491: Delete taskgraph.util.memoize
In https://github.com/taskcluster/taskgraph/pull/490 we're getting rid of our custom `memoize` implementation in favour of `functools.lru_cache`. Next time we do a major bump we should just delete this utility file entirely and consumers can just use `functools.lru_cache` or `functools.cache` directly.

bhearsum:


ahal (approved):
Thanks!

PR #516, opened 2024-06-07
Upstream some fixes in `taskgraph.util.templates` from gecko_taskgraph

ahal:
This upstreams a pair of changes to this file, one of which is breaking.


hneiva (approved):
LGTM

PR #518, opened 2024-06-12
chore: update dependencies

ahal:


ahal:
The failure is a file-lock regression that release 30 minutes ago:
https://github.com/tox-dev/filelock/issues/337

Which in turn breaks all virtualenv creation:
https://github.com/pypa/virtualenv/issues/2735

ahal:
The bad release was yanked from pypi.

gbrownmozilla (approved):

PR #519, opened 2024-06-12
ci: switch unit tests to one task per Python version

ahal:


ahal:
Sorry for not splitting the commit up better.. I didn't have a coherent plan and needed to backtrack a bunch. In the end it was easier to just put everything in one big commit. This commit is doing a few things:

1. Creating one docker-image / test task per Python version we support
2. Dropping dependency on pyenv
3. Moving coverage combine and reporting to the `codecov` task (used to be done with tox)
4. Dropping dependency on tox (no point in it anymore)
5. Splitting the pyright task into a new `check` kind

ahal:
I'll update the required checks once this lands. Though we should probably set up a `complete-pr` task to avoid needing to manually update them. I'll leave that for a separate PR though.

hneiva (approved):
Very nice! Thanks for figuring this out 🚀

PR #520, opened 2024-06-13
Update `image_builder` Docker image to version 5.1.0

jcristau:


taskcluster/docker/python/Dockerfile
```diff
@@ -7,10 +7,10 @@ FROM python:$PYTHON_VERSION-slim
 LABEL maintainer="Release Engineering <release@mozilla.com>"
 
 # Add worker user
-RUN mkdir /builds && \
+RUN mkdir -p /builds && \
```
bhearsum, line 10: nit: I don't think `-p` serves any purpose here?

bhearsum (approved):

taskcluster/docker/python/Dockerfile
jcristau: Without `-p`, `mkdir` fails because the directory already exists (AFAICT because the newer version of kaniko handles the `VOLUME` instructions before anything else).  Arguably this makes the `mkdir` unnecessary, but I opted to keep it there in case the Dockerfile ever gets used with another tool that behaves differently.

bhearsum: Ah, right. I forgot that `-p` does more than create intermediate directories.

PR #524, opened 2024-06-20
fix: make IndexSearch optimization log reasons why it may refuse to u…

bhearsum:
…se a replacement task


ahal (approved):
Thanks! This is rare enough that it shouldn't add much verbosity to the log.

PR #526, opened 2024-06-21
fix(init): set level to 1 for pull requests on level 3 projects

jcristau:


template/{{cookiecutter.project_name}}/taskcluster.github.yml
```diff
@@ -93,6 +92,10 @@ tasks:
                       'head_ref[:10] == "refs/tags/"': '${head_ref[10:]}'
                       'head_ref[:11] == "refs/heads/"': '${head_ref[11:]}'
                       $default: '${head_ref}'
+              level:
+                  $if: 'isPullRequest || tasks_for == "pr-action"'
+                  then: "1"
+                  else: "{{cookiecutter.level}}"
```
ahal, line 98: This could result in an `then: "1", else: "1"` block which would look a little weird.. but probably not a big deal. If we really wanted we could use Jinja2 if statements to toggle between this and the original `level: "{{cookiecutter.level}}"` statement. But I'm fine with this.

ahal (approved):

template/{{cookiecutter.project_name}}/taskcluster.github.yml
jcristau: It's been 3 weeks and I didn't get back to fixing that template, so I'm going to merge as-is as it's an improvement already.

PR #528, opened 2024-06-25
feat(init): use versioned decision image

### LINKED ISSUE #493: `taskgraph init` should use version pinned decision image and no requirements file
Right now it uses a plain hash pinned one and adds a requirements file that uses the wrong version of taskgraph :laughing:

jcristau:
Fixes #493


template/cookiecutter.json
```diff
@@ -1,5 +1,6 @@
 {
     "project_name": "my-project",
+    "taskgraph_version": "{{ cookiecutter.taskgraph_version }}",
```
ahal, line 3: I don't think this will work. The template context comes from the other variables in this file, so this seems like it would lead to some kind of recursion error?

Instead this is supposed to be the default version. I think this isn't a question that we should prompt for and instead we should either use `latest` or the same version as what they are running `taskgraph init` with.

ahal (changes requested):

template/cookiecutter.json
ahal: If we wanted to be really fancy, we could query pypi for the latest version and use that.

ahal: On second thought, we should definitely use the same version as what they ran `taskgraph init` with. Otherwise, the generated graph might fail inside the Decision task.

ahal: Oh, I think I missed the extra context being passed in there. And thanks for the link.

ahal (approved):

PR #529, opened 2024-06-26
Bug 1904784 - retry batched taskcluster requests

jcristau:
POST requests aren't retried by default, because they may not be idempotent.  However both find_task_id_batched and status_task_batched are read-only so there's no reason not to retry.


jcristau:
A couple of things:
- as far as I can tell there's no test coverage for the batched requests
- we may want to audit all other calls to `_do_request` and decide that retrying all POSTs is safe

ahal (approved):

PR #545, opened 2024-07-16
feat: support fully configured fetches for toolchain and fetch upstreams

bhearsum:
This allows for toolchains for things such as Python wheels to be fetched without being extracted in the wake of https://github.com/taskcluster/taskgraph/pull/541

### DIFF
--- a/src/taskgraph/transforms/run/__init__.py
+++ b/src/taskgraph/transforms/run/__init__.py
@@ -14,7 +14,7 @@ import copy
 import json
 import logging
 
-from voluptuous import Any, Exclusive, Extra, Optional, Required
+from voluptuous import Exclusive, Extra, Optional, Required
 
 from taskgraph.transforms.base import TransformSequence
 from taskgraph.transforms.cached_tasks import order_tasks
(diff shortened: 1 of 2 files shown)


src/taskgraph/transforms/run/__init__.py
```diff
@@ -252,7 +251,18 @@ def use_fetches(config, tasks):
...
-                for fetch_name in sorted(artifacts):
+                for artifact in sorted(artifacts):
+                    # Convert name only fetch entries to full fledged ones for
+                    # easier processing.
+                    if isinstance(artifact, str):
+                        artifact = {
+                            "artifact": artifact,
+                            "dest": None,
+                            "verify-hash": False,
+                            "extract": True,
```
jcristau, line 262: Those are the defaults so maybe don't specify them here?

src/taskgraph/transforms/run/__init__.py
```diff
@@ -85,7 +85,6 @@
         },
         # A list of artifacts to install from 'fetch' tasks.
         Optional("fetches"): {
-            Any("toolchain", "fetch"): [str],
```
jcristau, line 88: ruff says Any is now imported but unused

jcristau (approved):

src/taskgraph/transforms/run/__init__.py
bhearsum: ah, true

bhearsum:
This patch results in an identical taskgraph for a Gecko autoland on-push graph. I think that's good enough upfront verification.

PR #547, opened 2024-07-18
Bug 1853082 - Don't rely on file extension to guess archive type in f…

jcristau:
…etch-content. r=taskgraph-reviewers,bhearsum

I want to add a fetch for a rust crate archive from crates.io, and those don't come with an extension in the url, so we can't rely on that. Content sniffing is easy enough to put in place.

Differential Revision: https://phabricator.services.mozilla.com/D188152

[Attempting to re-land this as a breaking change]

### DIFF
--- a/src/taskgraph/run-task/fetch-content
+++ b/src/taskgraph/run-task/fetch-content
@@ -10,6 +10,7 @@ import contextlib
 import datetime
 import gzip
 import hashlib
+import io
 import json
 import lzma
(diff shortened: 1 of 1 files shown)


jcristau:
This code originated in gecko's run-task, previously landed in #541, and was reverted in #543 for breaking backwards compatibility.

bhearsum (approved):

PR #555, opened 2024-07-31
Fix renames not getting detected as file changes on git

Eijebong:
Because git detects file renames, they have a custom way of handling it in diffs and a custom filter for it. Using `--no-renames` gets git to behave like mercurial here and show a renamed file as a removal of the old name and an addition of the new one.


ahal (approved):
Thanks for the fix!

PR #565, opened 2024-08-19
Remove useless tasks list allocation in `transforms.docker_image`

Eijebong:
This was allocating a list of tasks from the passed in `tasks` generator before iterating over it. The reason this was here is because when that code got imported into this repo, (see be320c2b3a22c03d5fe7583f761456f75267c9b0), it was reordering the tasks before iterating over them. Said reordering got removed in ff0097edea821f8bab8ac9975f3ffc740956d683 but the allocation stayed.

I'm not expecting this to have any impact whatsoever but it confused me and ended up being useless so we might as well remove it.

### DIFF
--- a/src/taskgraph/transforms/docker_image.py
+++ b/src/taskgraph/transforms/docker_image.py
@@ -75,8 +75,6 @@ def fill_template(config, tasks):
 
     context_hashes = {}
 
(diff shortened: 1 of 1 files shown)


jcristau (approved):
Thanks!

PR #566, opened 2024-08-28
fix!: stop hardcoding path to Python in MacOS generic-worker tasks

ahal:
BREAKING CHANGE: Path to Python installation on MacOS no longer hardcoded.

### DIFF
--- a/docs/reference/migrations.rst
+++ b/docs/reference/migrations.rst
@@ -3,6 +3,20 @@ Migration Guide
 
 This page can help when migrating Taskgraph across major versions.
 
+10.x -> 11.x
+------------
+
+* A hardcoded path to a Python installation was removed for MacOS
+  generic-workers. If Mac tasks start failing after upgrade and you are able to
+  change the worker environment, ensure the ``python3`` binary is available on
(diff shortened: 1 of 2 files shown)


docs/reference/migrations.rst
```diff
@@ -3,6 +3,18 @@ Migration Guide
...
+------------
+
+* A hardcoded path to a Python installation was removed for MacOS
+  generic-workers. If Mac tasks start failing after upgrade, add the following
+  to their definitions:
+
+  .. code-block:: yaml
+
+     run:
+       run-task-command: ["/tools/python36/bin/python3", "run-task"]
```
jcristau, line 16: Should the advice be instead to ensure that python3 is on $PATH?

ahal: That assumes the people doing the upgrade can easily change the workers. This suggestion keeps things working entirely in-task. But I suppose we could suggest both options.

jcristau: That's fair, I was thinking environment can also be set in the task payload, but doing that for PATH is probably not a good idea in general, and we may prefer to rely on the default set on the worker.

jcristau (approved):

PR #569, opened 2024-09-03
fix: get_ancestors should ignore 404s

bhearsum:
This can happen when attempting to fetch tasks (either directly or indirectly) that have expired. My view on this is that such tasks should be treated as no longer being dependencies since they no longer exist.

I can see a case for maybe making this optional behaviour, but I don't know of any use cases where we'd want to raise an exception for this case. Certainly when being used as part of actions or other CI code it ends up being a ticking time bomb.

### DIFF
--- a/src/taskgraph/util/taskcluster.py
+++ b/src/taskgraph/util/taskcluster.py
@@ -462,7 +462,13 @@ def list_task_group_incomplete_tasks(task_group_id):
 def _get_deps(task_ids, use_proxy):
     upstream_tasks = {}
     for task_id in task_ids:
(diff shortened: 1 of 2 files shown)


hneiva (approved):
I agree that this should be the expected behaviour.

ahal (approved):
Makes sense to me, I'm surprised we haven't hit this before now.

We could return a tuple with `expired_ancestors` being the second variable.. But it's probably not worth the breaking change. Also it wouldn't even be fully accurate as we'd have no way of finding dependencies of the expired tasks in turn.

PR #570, opened 2024-09-04
Update Dockerfile special `ARG` syntax

matt-boris:
To adhere to https://github.com/taskcluster/taskgraph/blob/f20ebf1921e2feed7682b074f91fc96609d5319b/src/taskgraph/util/docker.py#L103-L106


ahal (approved):
Huh... Good catch.

Due to https://github.com/taskcluster/taskgraph/blob/5b9cd0a6054f36c725cc4e4c462467dcad61f142/src/taskgraph/util/docker.py#L158 I have absolutely no idea how this is working. But looking at the latest image builder tasks for this Dockerfile, it does look like each task is using the right image. I'm very confused.

PR #572, opened 2024-09-05
chore: use uv for package management; hatchling for builds

ahal:


src/taskgraph/run-task/run-task
```diff
@@ -1011,7 +1011,11 @@ def install_pip_requirements(repositories):
         return
 
     # TODO: Stop using system Python (#381)
-    cmd = [sys.executable, "-mpip", "install", "--break-system-packages"]
+    if shutil.which("uv"):
```
bhearsum, line 1014: This _feels_ like it could be risky, but I see that taskgraph's own tests are already running OK and end up here, so that's some level of confidence. (I suspect that we may see issues elsewhere for unpredictable reasons, but at least those ought to be caught in PRs as we start adding `uv` to various docker images and/or workers.)

bhearsum (approved):

src/taskgraph/run-task/run-task
ahal: Sorry, I'm not sure I'm understanding the risk. Is the concern that `uv pip install` will differ in some way from `pip install`? That's a valid concern, but a design goal of `uv pip install` is to be a drop-in replacement for `pip install`. So it's certainly possible there are implementation errors, but if it ever deviates from `pip install` they treat that as a bug. So far my confidence in this tool is very high.

ahal: It's also worth noting that we might not detect this via a patch to the repo if the repo is using an image from DockerHub and not pinning. But that's a tradeoff they decided to make, so I'm not too worried about it.

bhearsum: It's not a huge concern or anything, I'm just calling out that changing the way we install packages is an inherent risk and notable change, and that we should watch out for bustage when we start hitting this code path in other projects. Certainly not a blocker for doing this!

ahal: Thanks for clarifying! I do agree there is a level of risk here, but the perf boost is quite substantial. So I think it'll be worthwhile

PR #576, opened 2024-09-20
chore: version bump 11.1.0

bhearsum:
I could be convinced that #575 requires a major bump, but I don't think we typically do that when we change the Taskcluster service version we depend on?


CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
 # Change Log
 
+## [11.0.1] - 2024-09-20
```
jcristau, line 3:
```suggestion
## [11.1.0] - 2024-09-23
```

jcristau (changes requested):
This seems to be missing the actual version bump?

PR #585, opened 2024-10-07
Strip `build-` prefix from docker image task labels

ahal:


ahal:
There's no rush on this, just wanted to tee it up for the next time we're doing a breaking change anyway.

jcristau:
I guess this is to align with gecko?

ahal:
Yeah, it's blocking us consolidating the docker image logic as this prefix is hardcoded in a few places.

I guess we could make it a configuration option, but I don't like that Taskgraph defaults to `build-` as a prefix anyway, as this makes it easy to confuse them as the `build` kind if someone is parsing labels for the kind (which they shouldn't do, but happens in practice).

jcristau (approved):
Makes sense.

PR #588, opened 2024-10-09
Create built-in `matrix` transforms to facilitate splitting tasks by arbitrary values

ahal:
This was inspired by a comment someone made around how they much preferred Github Actions' `matrix` scheme to how you need to create purpose built transforms anytime you wanted to split a task by some set of values.

On reflection, I had to agree. I figured it wouldn't be terribly hard to implement a similar solution in Taskgraph.


docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+Will generate these tasks:
+
+- ``test-win-py312``
+- ``test-win-py311``
+- ``test-mac-py312``
+- ``test-mac-py311``
+- ``test-linux-py312``
+- ``test-linux-py311``
+
+Note that the name of the tasks will be built based on the order of rows in the
```
bhearsum, line 62: Is this guaranteed? It seems like this may differ based on what the yaml parser does and/or the `dict` (unless these are loaded into OrderedDicts).

docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+yaml substitution:
+
+.. code-block:: yaml
+
+   tasks:
+     test:
+       matrix:
+         os: ["win", "mac", "linux"]
+       description: Run {matrix[os]} tests
+       worker-type: {matrix[os]}-worker
```
bhearsum, line 84: This is an excellent default.

docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+Limiting Substitution
+'''''''''''''''''''''
+
+By default, all keys and values in the task definition will be checked for
+substitution parameters. But in some cases, it might be desirable to limit which
+keys get substituted, such as when using the ``matrix`` transforms alongside
+other transforms that perform substitution, such as the
+:mod:`~taskgraph.transforms.task_context` or
+:mod:`~taskgraph.transforms.chunking` transforms.
+
```
bhearsum, line 95: In translations, we've had cases where limiting to a particular field is not enough because we need to format the same value in multiple transforms. I ended up implementing a [partial substitution function](https://github.com/mozilla/firefox-translations-training/blob/d1d1efc441bfa9fc0f3e2c176f778197d16899df/taskcluster/translations_taskgraph/util/substitution.py) to handle these.

I'm not super eager to have that upstreamed, but as we add more and more very general transforms it may become necessary.

(The `task-context` part below may help reduce the need for this, but it can still come up elsewhere. Eg: in translations we have places where we have chunk numbers being substituted as well as other things.)

src/taskgraph/transforms/matrix.py
```diff
@@ -0,0 +1,130 @@
...
+
+        new_tasks = [task]
+        for key, values in matrix.items():
+            new_tasks = _resolve_matrix(new_tasks, key, values, exclude)
+
+        for new_task in new_tasks:
+            if set_name:
+                if "name" not in fields:
+                    fields.append("name")
+                new_task["name"] = set_name
```
bhearsum, line 116: I see that `name` is getting set both in `_resolve_matrix` and here. Can this be centralized in one place or the other? (I don't really see why `name` needs to be set when resolving the matrix, but maybe I'm missing something.)

src/taskgraph/transforms/matrix.py
```diff
@@ -0,0 +1,130 @@
...
+
+            if defer:
+                task_context = new_task.setdefault("task-context", {})
+                task_context["substitution-fields"] = list(fields)
+                task_context["from-object"] = {
```
bhearsum, line 121: If you are going to maintain support for this, you should amend `substitution-fields` and `from-object` rather than overwriting them, to ensure that they can be used as usual in a `kind` still.

bhearsum (commented):
This looks really fantastic overall! I"m leaving this unapproved for the moment because I'm interested in answers to some of the questions below.

docs/reference/transforms/matrix.rst
ahal: Yes, as of Python 3.7 the dict class preserves insertion order, so there's usually no need for OrderedDicts anymore. PyYaml should also keep the order because of this

docs/reference/transforms/matrix.rst
ahal: Yeah, I realized we have a lot of transforms that do some kind of substitution now..

By my count there's:

* task_context
* chunking
* from_deps
* run (substitutes `cwd: {checkout}`)
* and now matrix

Perhaps we need to consolidate this somehow. Maybe each one of these transforms could build up `**subs` in an attribute, then the `task` transforms could do a single substitution at the very end? I'm not sure if that would work for all cases or not though.

docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+notation can be used in ``substitution-fields`` to limit substitution to some
+sub configuration of the task definition.
+
+Deferring Substitution for the Task Context Transforms
+''''''''''''''''''''''''''''''''''''''''''''''''''''''
+
+In some cases where both the :mod:`~taskgraph.transforms.matrix` and
+:mod:`~taskgraph.transforms.task_context` transforms are being used, you may
+want to defer substitution entirely and let the ``task_context`` transforms
+handle it. To do this, you can set ``defer-substitution: true``:
```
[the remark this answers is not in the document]
ahal: Yes, and yes.

When this is true, it simply adds `"task_context": { "from-object": {"matrix": <matrix>}}` to the task definition, so the `task_context` transforms handle the substitution.

And yes, it's so you can use both transforms without interfering with one another.

docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+
+   tasks:
+     test:
+       matrix:
+         defer-substitution: true
+         os: ["win"]
+
+This will format the task definition to follow the ``task_context`` schema. If
+passing ``defer-substitution: true``, the ``task_context`` transforms **must** run
+afterwards, or else a task generation error will occur.
```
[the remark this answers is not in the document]
ahal: > Do we have other places where two entirely separate transforms are coupled like this? It strikes me as undesirable, even with it being optional, although I'm not going to say I'd reject it outright.

I don't think so.. In this case the only purpose of `defer-substitution` is to allow using this alongside `task_context`.. But maybe this isn't really necessary and we can leave it out for now?

> **(Value substitution has gotten complicated enough that I wonder if we need to take a new approach at some point. One idea might be to have it happen in the task transform, and other transforms would stick things into attributes instead. This is obviously well out of scope for this PR though!)**

Oh lol, I just commented the same idea to one of your earlier comments.. great minds :p

docs/reference/transforms/matrix.rst
```diff
@@ -0,0 +1,210 @@
...
+     test:
+       matrix:
+         os: ["win", "mac", "linux"]
+         python: ["py312", "py311"]
+         exclude:
+           - os: mac
+           - os: win
+             python: py311
+
+This will cause all combinations where ``os == mac`` to be skipped, as well as
```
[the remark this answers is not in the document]
ahal: True, this feature would mostly only be useful if you have at least three dimensions, and you want to exclude all combinations of two of them.

Do you think it's better to make the example simpler or more realistic?

src/taskgraph/transforms/matrix.py
ahal: We could centralize it here, it would just mean we need to iterate over the matrix values again. I don't have a strong preference, want me to move it?

docs/reference/transforms/matrix.rst
bhearsum: Yeah, that seems plausible for many cases. If, eg: we ever have a transform that cares about a value that an earlier transform may substitute it could be annoying. I don't think any of this needs to block this work though, for the record.

docs/reference/transforms/matrix.rst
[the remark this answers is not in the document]
bhearsum: I'd err on the side of realistic myself. (I also think it's fine if you only include the example that relies on os & python being excluded.)

src/taskgraph/transforms/matrix.py
bhearsum: Up to you!

docs/reference/transforms/matrix.rst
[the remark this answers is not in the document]
ahal: Yeah, maybe I should leave it out for now. It'll be easier to add it later if needed than remove it after the fact

src/taskgraph/transforms/matrix.py
ahal: I dropped this.

bhearsum (approved):

jcristau (approved):
Nice!

PR #590, opened 2024-10-15
fix(decision): ensure 'taskgraph' is on the $PATH in Decision image

ahal:
This was regressed by the switch to `uv` for packaging in the Decision image.


hneiva (approved):
LGTM

PR #592, opened 2024-10-16
fix(decision): add --no-cache to another uv command

jcristau:
We don't want to leave a root-owned cache in the worker user's HOME directory.

### DIFF
--- a/taskcluster/docker/decision/system-setup.sh
+++ b/taskcluster/docker/decision/system-setup.sh
@@ -9,7 +9,7 @@ apt-get install -y --force-yes --no-install-recommends \
     python3-pip
 
 pushd /setup/taskgraph
-uv export --no-dev > /setup/requirements.txt
+uv export --no-cache --no-dev > /setup/requirements.txt
 uv pip install --no-cache --system --break-system-packages -r /setup/requirements.txt
 uv pip install --no-cache --system --break-system-packages --no-deps .
 popd


jcristau:
Verified that the generated decision image doesn't have root-owned stuff in /builds/docker anymore:
```
root@53686a944a52:/# ls -la /builds/worker/
total 24
drwxr-xr-x 3 worker worker 4096 Oct  8 14:22 .
drwxr-xr-x 3 root   root   4096 Oct  8 14:22 ..
-rw-r--r-- 1 worker worker  220 Mar 29  2024 .bash_logout
-rw-r--r-- 1 worker worker 3526 Mar 29  2024 .bashrc
-rw-r--r-- 1 worker worker  807 Mar 29  2024 .profile
drwxr-xr-x 2 worker worker 4096 Oct  8 14:22 artifacts
```

jcristau:
Bah, this is not enough, see https://firefox-ci-tc.services.mozilla.com/tasks/PiXh8GIQR1-NJSIGeJw9xQ/runs/0/logs/public/logs/live_backing.log
```
[pip-install 2024-10-16T11:13:03.385Z] executing ['uv', 'pip', 'install', '--python', '/usr/bin/python3', '--break-system-packages', '--require-hashes', '-r', '/builds/worker/checkouts/vcs/taskcluster/requirements.txt']
[pip-install 2024-10-16T11:13:03.437Z] error: failed to create file `/tmp/uv-b17927aca3d28934.lock`
[pip-install 2024-10-16T11:13:03.438Z]   Caused by: Permission denied (os error 13)
```

ahal:
That looks like a separate issue, so I think we can still merge this PR

ahal (approved):
Thanks, I didn't think `export` would create a cache dir :/

ahal:
The codecov failure is a 503 on their end.

PR #593, opened 2024-10-16
fix(decision): make /tmp readable by worker user

ahal:


jcristau (changes requested):
This isn't the issue, /tmp is readable.  The problem is we leave a root-owned file in /tmp that then prevents uv from working as the worker user.

PR #594, opened 2024-10-16
fix(decision): cleanup temporary uv lockfiles

ahal:
This was causing problems, as the `worker` user was trying to create the same file, but it already existed and was owned by `root`.

Ensure we don't leave these lying around.


ahal:
I guess this is why we should be using multi-stage image builds ><

jcristau (approved):

PR #598, opened 2024-10-17
Fix readthedocs builds

ahal:
The fix was copy/pasted directly from the readthedocs documentation:
https://docs.readthedocs.io/en/stable/build-customization.html#install-dependencies-with-uv


hneiva (approved):
LGTM

PR #601, opened 2024-10-25
feat(cli): support --parameters=index:<index path>

ahal:
Currently we support finding the decision task id based on `project`, from the index, but sometimes we want to find other types of ids (e.g for cron decision tasks).


bhearsum (approved):
Does this warrant a doc update?

ahal:
Thanks for the call out.. yes it does.

PR #603, opened 2024-10-28
Don't assume the proxy is running on `http://taskcluster`

Eijebong:
This fixes the two places where taskgraph assumed the proxy would be running on `http://taskcluster` instead of looking at the `TASKCLUSTER_PROXY_URL` environment variable.

### DIFF
--- a/src/taskgraph/create.py
+++ b/src/taskgraph/create.py
@@ -124,7 +124,9 @@ def create_task(session, task_id, label, task_def):
         return
 
     logger.info(f"Creating task with taskId {task_id} for {label}")
-    proxy_url = os.environ.get("TASKCLUSTER_PROXY_URL", "http://taskcluster")
+    proxy_url = os.environ.get("TASKCLUSTER_PROXY_URL", "http://taskcluster").rstrip(
+        "/"
+    )
     res = session.put(
         f"{proxy_url}/queue/v1/task/{task_id}",
         json=task_def,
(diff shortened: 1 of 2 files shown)


src/taskgraph/create.py
```diff
@@ -123,7 +124,11 @@ def create_task(session, task_id, label, task_def):
         return
 
     logger.info(f"Creating task with taskId {task_id} for {label}")
-    res = session.put(f"http://taskcluster/queue/v1/task/{task_id}", json=task_def)
+    proxy_url = os.environ.get("TASKCLUSTER_PROXY_URL", "http://taskcluster")
```
ahal, line 127: Let's `rstrip("/")` the env first

src/taskgraph/run-task/run-task
```diff
@@ -39,7 +39,7 @@ from pathlib import Path
 from threading import Thread
 from typing import Optional
 
-SECRET_BASEURL_TPL = "http://taskcluster/secrets/v1/secret/{}"
+SECRET_BASEURL_TPL = "{}/secrets/v1/secret/{{}}".format(os.environ.get("TASKCLUSTER_PROXY_URL", "http://taskcluster"))
```
ahal, line 42: Ditto

ahal (changes requested):
Thanks! I'd like to strip trailing slashes first because someone is inevitably going to add a trailing slash

ahal (approved):
Thanks!

PR #604, opened 2024-11-01
Do not apply skip-unless-changed for cron triggers.

afranchuk:
If we want a little more flexibility, we could add an additional optional parameter to turn this behavior on or off (there _might_ be some utility in tasks that strictly only run if some files changed, so that you can trigger all tasks with cron and those will be optimized away).

### DIFF
--- a/src/taskgraph/optimize/strategies.py
+++ b/src/taskgraph/optimize/strategies.py
@@ -80,8 +80,9 @@ class SkipUnlessChanged(OptimizationStrategy):
         return False
 
     def should_remove_task(self, task, params, file_patterns):
-        # pushlog_id == -1 - this is the case when run from a cron.yml job or on a git repository
-        if params.get("repository_type") == "hg" and params.get("pushlog_id") == -1:
+        # skip-unless-changed should not apply when there is no commit delta,
+        # such as for cron and action tasks (there will never be file changes)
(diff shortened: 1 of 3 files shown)


src/taskgraph/optimize/strategies.py
```diff
@@ -80,8 +80,8 @@ def check(self, files_changed, patterns):
         return False
 
     def should_remove_task(self, task, params, file_patterns):
-        # pushlog_id == -1 - this is the case when run from a cron.yml job or on a git repository
-        if params.get("repository_type") == "hg" and params.get("pushlog_id") == -1:
+        # skip-unless-changed should not apply to cron tasks (there will never be file changes)
+        if params.get("tasks_for") == "cron":
```
jcristau, line 84: I suspect the same is true for action tasks, as far as I can tell they'll also have identical base-rev and head-rev, unlike push and pull-request events.  Could we check for that instead of explicitly looking for cron?

jcristau: Right, I'm hoping that would do the trick (but it depends on what .taskcluster.yml does, in case projects differ from the template that might complicate things).

afranchuk:
Oh, I need to update the other tests!

ahal (approved):
Thanks, lgtm!

PR #608, opened 2024-11-21
feat: support lists in 'resolve_keyed_by'

ahal:
I tested this in Gecko and it resulted in an identical full graph

### DIFF
--- a/docs/howto/resolve-keyed-by.rst
+++ b/docs/howto/resolve-keyed-by.rst
@@ -1,0 +1,251 @@
+Use Keyed By
+============
+
+Often fields in a task can depend on other values in the task. For example, a
+task's ``max-runtime`` may depend on the ``platform``. To handle this, you
+could re-define ``max-runtime`` in each task's definition like so:
+
+.. code-block:: yaml
+
+   tasks:
+     taskA:
+       platform: android
+       worker:
+         max-runtime: 7200
+
+     taskB:
+       platform: ios
+       worker:
+         max-runtime: 7200
+
+     taskC:
+       platform: windows
+       worker:
+         max-runtime: 3600
+
+     taskD:
+       platform: mac
+       worker:
+         max-runtime: 1800
+
+     ...
+
+This is simple, but if you have lots of tasks it's also tedious and makes
+updating the configuration a pain. To avoid this duplication you could use a
+:doc:`transform </concepts/transforms>`:
+
+.. code-block:: python
+
+   @transforms.add
+   def set_max_runtime(config, tasks):
+       for task in tasks:
+           if task["platform"] in ("android", "ios"):
+               task["worker"]["max-runtime"] = 7200
+           elif task["platform"] == "windows":
+               task["worker"]["max-runtime"] = 3600
+           else:
+               task["worker"]["max-runtime"] = 1800
+
+           yield task
+
+This works but now we've hardcoded constants into our code logic far away from
+the task's original definition! Besides this is pretty verbose and it can get
+complicated if you want to be able to change these constants per task.
+
+An Alternative Approach
+-----------------------
+
+Another way to accomplish the same thing is to use Taskgraph's "keyed by"
+feature. This can be used in combination with the ``task-defaults`` key to
+express the same logic directly in the ``kind.yml`` file:
+
+.. code-block:: yaml
+
+   task-defaults:
+     worker:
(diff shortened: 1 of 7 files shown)


jcristau (changes requested):
Looks like this breaks on python 3.8?

src/taskgraph/util/keyed_by.py
```diff
@@ -2,8 +2,40 @@
...
 
+from typing import Any, Dict, Generator, Tuple
 
-from .attributes import keymatch
+from taskgraph.util.attributes import keymatch
+
+
+def iter_dot_path(
+    container: Dict[str, Any], subfield: str
+) -> Generator[Tuple[dict[str, Any], str], None, None]:
```
hneiva, line 12: Python < 3.9 can't handle type annotation using built-in types 😢 Just need to replace `dict` with `Dict` (I think it can handle `str` tho). Very annoying.

bhearsum:
Moving this back to draft to get it out of review queues.

ahal:
Oops, I forgot about this PR.. I'll get it into shape soon

test/test_util_keyed_by.py
```diff
@@ -0,0 +1,29 @@
...
+        pytest.param(
+            {"a": {"b": {"c": 1}}, "d": 2}, "a.b.c", [({"c": 1}, "c")], id="simple case"
+        ),
+        pytest.param(
+            {"a": [{"b": 1}, {"b": 2}, {"b": 3}], "d": 2},
+            "a.b",
+            [({"b": 1}, "b"), ({"b": 2}, "b"), ({"b": 3}, "b")],
+            id="list case",
+        ),
+    ),
```
bhearsum, line 24: Some tests for error / corner cases would be welcome. Eg:
container: `{"a": [{"b": 1}, {"c": 2}]}`
subfield: `a.b`

docs/howto/resolve-keyed-by.rst
```diff
@@ -0,0 +1,251 @@
...
+                      default: bar.tar.gz
+
+With the associated transform:
+
+.. code-block:: python
+
+   @transforms.add
+   def resolve_artifact_paths(config, tasks):
+       for task in tasks:
+           resolve_keyed_by(task, "worker.artifacts.path", task["label"])
```
bhearsum, line 191: I don't want to bikeshed on this too much, so this is certainly not a blocker, but I have to admit that I find it a bit unintuitive that `.path` accesses multiple list items (in addition to access plain dict keys). I wonder if jq-style syntax would be preferable here? Eg: `worker.artifacts[].path`?

bhearsum (approved):
The doc improvements here are fantastic!

docs/howto/resolve-keyed-by.rst
```diff
@@ -0,0 +1,251 @@
...
+           for platform in task.pop("platforms"):
+               new_task = deepcopy(task)
+               # ...
+               resolve_keyed_by(
+                   new_task,
+                   "worker.max-runtime",
+                   task["label"],
+                   platform=platform,
+               )
+               yield task
```
jcristau, line 150:
```suggestion
               yield new_task
```

docs/howto/resolve-keyed-by.rst
ahal: I agree with you. I'll mark this as draft until I get a chance to implement a `[]` syntax.

ahal:
I didn't end up needing this for the thing I was working on at the time, so I've neglected it again.. but I'll try and get this landed at some point.

docs/howto/resolve-keyed-by.rst
```diff
@@ -0,0 +1,251 @@
...
+If there are more than one fields that another field may be keyed by, it
+can be specified like this:
+
+.. code-block:: python
+
+   Optional("max-run-time"): optionally_keyed_by("platform", "build-type", int)
+
+
+In this example either ``by-platform`` or ``by-build-type`` may be used. You
+may specify fields as you like this way, as long as the last argument to
```
ahal, line 249: s/specify fields/specify as many fields

docs/howto/resolve-keyed-by.rst
ahal: `worker.artifacts[].path` I presume?

bhearsum (approved):

PR #609, opened 2024-12-05
docs: improve the Reference -> Command Line section

gabrielBusta:
Found myself browsing the codebase after reading [reference/cli.html](https://taskcluster-taskgraph.readthedocs.io/en/5.5.0/reference/cli.html). Thought it'd be good to add more information about what the taskgraph sub-commands represent on this page.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -228,30 +228,30 @@ def generate_taskgraph(options, parameters, overrides, logdir):
 
 @command(
     "tasks",
-    help="Show all tasks in the taskgraph.",
+    help="Show the full task set in the taskgraph. The full task set includes all tasks defined by any kind, without edges (dependencies) between them.",
     defaults={"graph_attr": "full_task_set"},
 )
 @command(
-    "full", help="Show the full taskgraph.", defaults={"graph_attr": "full_task_graph"}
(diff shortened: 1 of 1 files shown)


src/taskgraph/main.py
```diff
@@ -228,30 +228,32 @@ def logfile(spec):
...
 @command(
     "tasks",
-    help="Show all tasks in the taskgraph.",
+    help="Show the full task set in the taskgraph. The full task set includes all tasks defined by any kind, without edges (dependencies) between them.",
     defaults={"graph_attr": "full_task_set"},
 )
 @command(
-    "full", help="Show the full taskgraph.", defaults={"graph_attr": "full_task_graph"}
+    "full",
+    help="Show the full taskgraph. The full taskgraph consists of the full task set, with edges (dependencies) between tasks.",
```
ahal, line 236: nit: s/taskgraph/task graph

While we're making things less confusing, let's make it clear that we're not talking about the package here. Note this comment applies to all the commands.

ahal (approved):
This change made me realize that there's probably no good reason for us to expose both the "set" and "graph" versions of each phase as top-level commands. IMO it only adds confusion.

ahal:
Filed https://github.com/taskcluster/taskgraph/issues/612 for my thought.

PR #610, opened 2024-12-06
fix: use base repo for project name in github .taskcluster.yml template

bhearsum:
project name is mainly used in routes, which require specific scopes to be set. Because of this, we should always use the `base` version of this. In cases where the `head` repository has a different name this ends up breaking things like actions in PRs.


jcristau (approved):
Consider making that same change in taskgraph's own .taskcluster.yml?

PR #611, opened 2024-12-06
feat: allow task group to be overridden in create_tasks/taskgraph_dec…

bhearsum:
…ision

I'm not sure yet if this is a good idea, but it's a possible path forward for https://github.com/mozilla-releng/fxci-config/pull/192, where I'm running an action task that I'd like to have status reported back into the task group that ran the action.

### DIFF
--- a/src/taskgraph/create.py
+++ b/src/taskgraph/create.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 testing = False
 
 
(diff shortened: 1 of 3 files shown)


bhearsum:
https://github.com/mozilla-releng/fxci-config/pull/192 is no longer an approach I'm taking (a different approach in https://github.com/mozilla-releng/fxci-config/pull/210 is being used instead). I'm closing this since it's no longer needed, and a fairly terrible hack in the first place.

PR #613, opened 2024-12-11
refactor: create new 'pytest-taskgraph' package to house test fixtures

ahal:
These fixtures are useful for any project that wants to test Taskgraph related transforms or scripts. By putting them in a package, we can share them to any other repos that may need them.


ahal:
I tried to use uv's workspaces here, but couldn't quite get it working. Using a path instead seems to work well though.

bhearsum (approved):
This is very cool, thank you!!

ahal:
There's a near 0% chance this is going to work perfectly on the first try, but I don't know how to test it all without creating a release. So I'll land and be prepared for a follow-up.

ahal:
Here is the inevitable follow-up PR I was talking about:
https://github.com/taskcluster/taskgraph/pull/614

Possibly more to come after that

PR #614, opened 2024-12-12
Use separate Github environment for releasing pytest-taskgraph

ahal:
This just ensures we don't accidentally publish the wrong package on the wrong release. The pypi-publish action should do this already, but doesn't hurt to have an extra safeguard.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -158,8 +158,12 @@ In order to release a new version of Taskgraph, you will need to:
 1. Update ``CHANGELOG.md``
 2. Update ``__version__`` in ``src/taskgraph/__init__.py``
 3. Commit, and land the above changes with a commit message like "chore: bump <version>"
-4. Create a release in Github pointing to the above commit. Be sure to also
-   create a new tag matching this version.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -174,9 +178,13 @@ release process for this package is:
 
 1. Update ``version`` in ``packages/pytest-taskgraph/pyproject.toml``
 2. Commit and land the changes with a commit message like "chore: bump pytest-taskgraph <version>"
-3. Create a release in Github pointing to the above commit. Be sure to also
-   create a new tag of the form ``pytest-taskgraph-v<version>``.
-4. Wait for the ``pypi-publish`` Github workflow and ``push-image-decision`` task to finish.
+3. Draft a release in Github pointing to the above commit.
+
+   a. Create a new tag of the form ``pytest-taskgraph-vX.Y.Z``
```
Eijebong, line 183: Bit weird to have `vX.Y.Z` on this tag but not on the other. Don't have a very strong opinion on this though

Eijebong (approved):

PR #615, opened 2024-12-12
chore: fix head_ref detection in 'release' target tasks method

ahal:
This ensures that the proper head ref is detected whether it is prefixed with `refs/tags/` or not. This also updates the test params to reflect reality.


ahal:
This bug caused the push-image tasks to run on a pytest-taskgraph release. Luckily those tasks were smart enough to detect the version already existed and bailed out (thanks @hneiva )

bhearsum (approved):

PR #617, opened 2024-12-24
run-task: update our copy of robustcheckout hg extension

jcristau:
This reverts a couple of local changes, back to what's in v-c-t.


jcristau:
More importantly for my purposes, this picks up https://hg.mozilla.org/hgcustom/version-control-tools/rev/f54fd9f1ad7538bb5d77c17ca1f802b66392efc4 instead of https://github.com/taskcluster/taskgraph/commit/5716426cd39b7bf5183e60fa6476a8bbfd83535c.  The latter breaks when hg uses python2.

PR #619, opened 2024-12-26
Add support for downloading a single file and putting it in an archive

glandium:


src/taskgraph/run-task/fetch-content
```diff
@@ -480,12 +480,31 @@ def should_repack_archive(
     return True
 
 
+EXECUTABLE_SIGNATURES = [
```
ahal, line 483: nit: might as well make this a set

src/taskgraph/run-task/fetch-content
```diff
@@ -960,6 +996,11 @@ def main():
         dest="headers",
         help="Header to send as part of the request, can be passed " "multiple times",
     )
+    url.add_argument(
+        "--single-file",
+        action="store_true",
+        help="The downloaded file is not an archive and should be put inside one",
```
ahal, line 1002: I was confused about the name "single_file" until I read the help text here. I think a better name would be `--force-archive`. Then if the url already points to an archive, the flag is a no-op, but if it points to a file, then the file is archived. Or maybe `--force-archive-single-file` if you want to keep the logic here the same without checking whether the url is an archive or not.

ahal (changes requested):

ahal (approved):
Thanks, lgtm

PR #620, opened 2024-12-30
Bug 1526028 - Support opting into or out of specific caches in `run` transforms

ahal:


src/taskgraph/transforms/run/common.py
```diff
@@ -31,7 +32,11 @@ def add_cache(task, taskdesc, name, mount_point, skip_untrusted=False):
         skip_untrusted (bool): Whether cache is used in untrusted environments
             (default: False). Only applies to docker-worker.
     """
-    if not task["run"].get("use-caches", True):
+    use_caches = task["run"].get("use-caches", True)
+    if isinstance(use_caches, list):
+        use_caches = any(re.match(r, name) for r in use_caches)
```
Eijebong, line 37: Nit: rename `r` to `pattern`? Not a huge fan of single letter variable especially when they're so close to another one in the same scope (`re`). I don't have super strong opinions though so feel free to ignore.

Eijebong (approved):

PR #623, opened 2025-01-07
Bug 1937801 - Implement mechanism to use caches for common tools in run transforms

ahal:
This is an overhaul of how we set up caches in the run transforms. At a high level, this is:

1. Creating a new mechanism to register caches for common tools (pip, uv, npm, cargo)
2. Refactoring the existing checkouts cache to use this mechanism.
3. Providing task and project wide configs to include or exclude caches
4. Removing the `cache-dotcache` key

I recommend reviewing commit by commit.

Making this a draft PR for now so I have a chance to test it out with real tasks.

### DIFF
--- a/docs/reference/migrations.rst
+++ b/docs/reference/migrations.rst
@@ -3,6 +3,23 @@ Migration Guide
 
 This page can help when migrating Taskgraph across major versions.
 
+12.x -> 13.x
+------------
+
+* Remove all ``run.cache-dotcache`` keys. If it was set to ``true``, replace it
+  with:
+
+  .. code-block:: yaml
+
+     run:
+       use-caches: [checkout, <caches>]
+
+  Where caches can be any of ``cargo``, ``pip``, ``uv`` or ``npm``. If the task
+  was setting up ``.cache`` for another tool, a mount will need to be created
+  for it manually. If ``use-caches`` was previously set to ``false``, omit
+  ``checkout`` in the example above. If ``use-caches`` was previously set to
+  ``true``, replace ``true`` with the value above (including ``checkout``).
+
 11.x -> 12.x
 ------------
 
@@ -110,7 +127,7 @@ This page can help when migrating Taskgraph across major versions.
 5.x -> 6.x
 ----------
 
-* Replace all uses of ``command-context` with the more generalized ``task-context``
+* Replace all uses of ``command-context`` with the more generalized ``task-context``
 
 4.x -> 5.x
 ----------
--- a/src/taskgraph/transforms/run/common.py
+++ b/src/taskgraph/transforms/run/common.py
@@ -13,6 +13,13 @@ import json
 from taskgraph.util import path
 from taskgraph.util.taskcluster import get_artifact_prefix
 
+CACHES = {
+    "cargo": {"env": "CARGO_HOME"},
+    "npm": {"env": "npm_config_cache"},
+    "pip": {"env": "PIP_CACHE_DIR"},
+    "uv": {"env": "UV_CACHE_DIR"},
+}
+
 
 def get_vcsdir_name(os):
     if os == "windows":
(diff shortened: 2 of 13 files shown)


ahal:
This PR was tested in a try push in Gecko:
https://treeherder.mozilla.org/jobs?repo=try&revision=dcbe341f010d75c14a2b8f8983545ece6fbf2049

It should be ready for review

src/taskgraph/transforms/run/common.py
```diff
@@ -32,10 +39,13 @@ def add_cache(task, taskdesc, name, mount_point, skip_untrusted=False):
         skip_untrusted (bool): Whether cache is used in untrusted environments
             (default: False). Only applies to docker-worker.
     """
-    if not task["run"].get("use-caches", True):
+    worker = task["worker"]
+    if worker["implementation"] not in ("docker-worker", "generic-worker"):
+        # caches support not implemented
```
bhearsum, line 39: It seems a bit odd to me that `add_cache` silently does nothing in certain cases. This is clearly existing behaviour though, and I imagine it would break a lot of things if it changed, so it's certainly not a blocker here.

src/taskgraph/transforms/run/common.py
```diff
@@ -166,3 +172,24 @@ def support_vcs_checkout(config, task, taskdesc, repo_configs, sparse=False):
         taskdesc["worker"]["taskcluster-proxy"] = True
 
     return vcsdir
+
+
+def support_caches(task, taskdesc):
+    """Add caches for common tools."""
+    worker = task["worker"]
+    workdir = task["run"].get("workdir")
+    base_cache_dir = ".task-cache"
```
bhearsum, line 166: I think it's _probably_ fine to put these caches outside of their default place, and obviously it worked fine with all your testing with Gecko. I anticipate that some edge case tasks or issues will come up though (eg: scripts or other things that are trying to either pull things from a cache, or inject them, and not paying attention to the env vars when doing so). That's not something that ought to stop this work, but it may cause unexpected bustage when this is picked up in various places.

docs/reference/migrations.rst
```diff
@@ -3,6 +3,23 @@ Migration Guide
...
+  with:
+
+  .. code-block:: yaml
+
+     run:
+       use-caches: [checkout, <caches>]
+
+  Where caches can be any of ``cargo``, ``pip``, ``uv`` or ``npm``. If the task
+  was setting up ``.cache`` for another tool, a mount will need to be created
+  for it manually. If ``use-caches`` was previously set to ``false``, omit
```
bhearsum, line 19: Including a block (here here or in a linked doc) about exactly what to add to a task to set up a new cache would be useful (I don't think it's obvious to all readers).

This also makes me wonder if it would be beneficial to allow additional caches to be added in the project-wide or task configuration (so that the mounts would be set up automatically). This is firmly in the category of "future enhancement", though.

bhearsum (changes requested):
Requesting changes due to the failing tests (which look to be silly things, not indicative of major issues); this looks fine though.

src/taskgraph/transforms/run/common.py
ahal: Good call out!

The reason I couldn't use `.cache` is that when setting e.g `VOLUME /builds/worker/.cache/pip` in the the Dockerfile, `image-builder` actually creates that directory with root. The `run-task` script then chown's all VOLUMEs to the worker user, but it doesn't handle parent directories, so `~/.cache/pip` ends up owned by worker, but `~/.cache` ends up owned by root. This prevented Gecko from writing to `~/.cache` and broke some tests in a subtle way ><. We could fix `run-task` to chown the whole path to a VOLUME, but I thought it would be cleanest to just use an entirely separate directory.

Another issue is that the default place is different for every platform. So we would need logic in Taskgraph that detects which platform the task is running and tries to set the default place accordingly per tool. This is further complicated by the fact that under Windows the default cache dir is under `%LOCALAPPDATA%`, which I'm not sure `generic-worker` would be capable of resolving? We could probably implement this, but being explicit with the envs seemed simpler.

docs/reference/migrations.rst
ahal: Yeah, I think that makes sense. It should be possible by monkeypatching the CACHES dict, but we might want some kind of registry / decorator thing like we have for other things.

ahal:
Sorry for all the churn.. I reverted the problematic test and I'm going to re-land in a follow-up PR.

bhearsum (approved):

PR #624, opened 2025-01-09
Avoid computing expensive default values when the value is overridden…

### LINKED ISSUE #616: Filling default Parameters should avoid expensive operations
The default for files_changed is to do a somewhat expensive `repo.get_changed_files` call. It's aggravated by `gecko_taskgraph` having its own default that uses `repo.get_locally_changed_files`, which can be even more expensive.
But out-of-tree extension of the parameters schema (per `extend_parameters_schema`) is not the only way the value may be uselessly computed: one may call `Parameters(files_changed=...)`, and both values that were computed above end up thrown away.

glandium:
… anyways

The most expensive call at the moment is repo.get_changed_files, which does down the drain if:
- another default_fn overrides the value
- an explicit override is given when creating the `Parameters`

With this change, the default function can return a function as a value, which is not evaluated unless necessary.

Fixes #616

### DIFF
--- a/src/taskgraph/parameters.py
+++ b/src/taskgraph/parameters.py
@@ -110,7 +110,7 @@ def _get_defaults(repo_root=None):
         "do_not_optimize": [],
         "enable_always_target": True,
         "existing_tasks": {},
-        "files_changed": repo.get_changed_files("AM"),
+        "files_changed": lambda: repo.get_changed_files("AM"),
         "filters": ["target_tasks_method"],
         "head_ref": repo.branch or repo.head_rev,
         "head_repository": repo_url,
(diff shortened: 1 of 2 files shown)


src/taskgraph/parameters.py
```diff
@@ -210,7 +211,7 @@ def _fill_defaults(repo_root=None, **kwargs):
 
         for name, default in defaults.items():
             if name not in kwargs:
-                kwargs[name] = default
+                kwargs[name] = default() if inspect.isfunction(default) else default
```
jcristau, line 214:
```suggestion
                kwargs[name] = default() if callable(default) else default
```
?

ahal (approved):

PR #627, opened 2025-01-15
Back out "feat(caches): support opting into or out of specific name c…

ahal:
…aches"

This backs out commit a70a57c72c794d992075548eefc03755e387c9f1.

### DIFF
--- a/src/taskgraph/transforms/run/common.py
+++ b/src/taskgraph/transforms/run/common.py
@@ -9,7 +9,6 @@ consistency.
 
 import hashlib
 import json
-import re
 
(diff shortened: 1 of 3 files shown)


ahal:
For context I should have held off landing this as I ended up deciding that it was better to have well defined caches that need to be opted into explicitly.

I want to avoid having a release with this regex feature to ensure we don't have users accidentally using it and then needing to stop using it on the next release.

jcristau (approved):
nit: would love some explanation for the revert as part of the commit message :)

PR #630, opened 2025-01-15
[run-task] Support interpolating `$TASK_WORKDIR` in environment variables

ahal:
We currently hard code some environment variables that should be turned into absolute paths. But it would be great if we had the ability to do this for any environment variable.

My current use case is setting `PIP_CACHE_DIR` to an absolute path under the task's workdir. I could add it to the list of hard coded envs, but it would be better to do something like `PIP_CACHE_DIR=$TASK_WORKDIR/.cache/pip` and have run-task do the interpolation automatically.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1091,10 +1091,10 @@ def _display_python_version():
 
 
 def main(args):
-    os.environ["TASK_WORKDIR"] = os.getcwd()
+    task_workdir = os.environ["TASK_WORKDIR"] = os.getcwd()
     print_line(
         b"setup",
-        b"run-task started in %s\n" % os.environ["TASK_WORKDIR"].encode("utf-8"),
+        b"run-task started in %s\n" % task_workdir.encode("utf-8"),
     )
     print_line(
         b"setup",
@@ -1313,20 +1313,27 @@ def main(args):
     for repo in repositories:
         vcs_checkout_from_args(repo)
 
-    resource_process = None
-
-    try:
-        for k in ["MOZ_FETCHES_DIR", "UPLOAD_DIR"] + [
-            "{}_PATH".format(repository["project"].upper())
-            for repository in repositories
-        ]:
-            if k in os.environ:
-                os.environ[k] = os.path.abspath(os.environ[k])
+    # Interpolate environment variables with defined substitution patterns. For
+    # example, the variable `CACHE_DIR={task_workdir}/.cache` will be
+    # interpolated with the task's working directory.
+    env_subs = {
+        "task_workdir": task_workdir,
+    }
+    for k, v in os.environ.items():
+        for subname, subval in env_subs.items():
+            # We check for an exact match rather than using a format string to
+            # avoid accidentally trying to interpolate environment variables
+            # that happen to contain brackets for some other reason.
+            pattern = f"{{{subname}}}"
+            if pattern in v:
+                os.environ[k] = v.replace(pattern, subval)
                 print_line(
                     b"setup",
                     b"%s is %s\n" % (k.encode("utf-8"), os.environ[k].encode("utf-8")),
                 )
 
+    resource_process = None
+    try:
         if "MOZ_FETCHES" in os.environ:
             fetch_artifacts()
 
--- a/src/taskgraph/transforms/run/__init__.py
+++ b/src/taskgraph/transforms/run/__init__.py
@@ -367,7 +367,7 @@ def use_fetches(config, tasks):
             "task-reference": json.dumps(task_fetches, sort_keys=True)
         }
 
-        env.setdefault("MOZ_FETCHES_DIR", "fetches")
+        env.setdefault("MOZ_FETCHES_DIR", "{task_workdir}/fetches")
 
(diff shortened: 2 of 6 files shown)


src/taskgraph/run-task/run-task
```diff
@@ -1316,12 +1316,9 @@ def main(args):
...
     try:
-        for k in ["MOZ_FETCHES_DIR", "UPLOAD_DIR"] + [
-            "{}_PATH".format(repository["project"].upper())
-            for repository in repositories
-        ]:
-            if k in os.environ:
-                os.environ[k] = os.path.abspath(os.environ[k])
+        for k, v in os.environ.items():
+            if "$TASK_WORKDIR" in v:
+                os.environ[k] = v.replace("$TASK_WORKDIR", task_workdir)
```
Eijebong, line 1321: This feels a bit wrong, what about `${TASK_WORKDIR}`? Might be worth generalizing this and using `os.path.expandvar`?

Eijebong: Yeah, it's a valid concern. As a user I would definitely expect both to work though.

ahal: Hmm..
```
>>> os.path.expandvars("$HOME/$FOO/$BAR")
'/home/ahal/$FOO/$BAR'
```

so it does leave things unchanged if the env doesn't exist... Maybe that's ok then?

ahal: Alright, I don't want to get too distracted here, and this change would potentially be a big distraction. I'll update the PR to support both `$TASK_WORKDIR` and `${TASK_WORKDIR}`, then file an issue to explore making it more generic.

src/taskgraph/run-task/run-task
```diff
@@ -1313,20 +1313,27 @@ def main(args):
...
-    try:
-        for k in ["MOZ_FETCHES_DIR", "UPLOAD_DIR"] + [
-            "{}_PATH".format(repository["project"].upper())
-            for repository in repositories
-        ]:
-            if k in os.environ:
-                os.environ[k] = os.path.abspath(os.environ[k])
+    # Interpolate environment variables with defined substitution patterns. For
+    # example, the variable `CACHE_DIR={task_workdir}/.cache` will be
+    # interpolated with the task's working directory.
```
bhearsum, line 1318: It would be great to get some minimal mentions of this in the docs.

src/taskgraph/run-task/run-task
```diff
@@ -1313,20 +1313,27 @@ def main(args):
...
+    env_subs = {
+        "task_workdir": task_workdir,
```
bhearsum, line 1320: I assume there's a reason that we can't use the upper-cased `$TASK_WORKDIR` here? It's a bit weird to have both versions of it. If we can't use the upper cased version, I guess this is the best alternative. (Having a completely differently named variable that is actually the same value would be worse...)

bhearsum (approved):
> My current use case is setting PIP_CACHE_DIR to an absolute path under the task's workdir. I could add it to the list of hard coded envs, but it would be better to do something like PIP_CACHE_DIR=$TASK_WORKDIR/.cache/pip and have run-task do the interpolation automatically.

Just to make sure I understand fully: this is necessary when running payloads that don't execute through bash? (We already have places in translations where we [use $TASK_WORKDIR directly](https://github.com/mozilla/translations/blob/2eb36983958147fec3ab223827aea2084379bc15/taskcluster/kinds/train-student/kind.yml#L110), but that runs through bash, which I suppose is why that works.

ahal:
> Just to make sure I understand fully: this is necessary when running payloads that don't execute through bash? (We already have places in translations where we [use $TASK_WORKDIR directly](https://github.com/mozilla/translations/blob/2eb36983958147fec3ab223827aea2084379bc15/taskcluster/kinds/train-student/kind.yml#L110), but that runs through bash, which I suppose is why that works.

No, this is necessary to be able to specify absolute paths in the task definition with `generic-worker`. Normally we just use relative paths like `./artifacts` or `./checkouts` because the task user isn't known. But that breaks if the task ever changes the working directory.

Specifically for my use case, I want to be able to set:
```
PIP_CACHE_DIR=/home/task_123456/.task-cache/pip
```

directly in the task definition. Since `task_123456` can't be known, I'll set it to `{work_dir}/.task-cache/pip` instead and `run-task` will interpolate it.

src/taskgraph/run-task/run-task
ahal: I agree, but `run-task` and the `run` transforms are entirely undocumented atm. Documenting this should be pretty high priority, but I don't want to scope bloat just to have a place to add this.

src/taskgraph/run-task/run-task
ahal: (see my explanation in the main thread, I think it'll make more sense then)

Eijebong (approved):

src/taskgraph/run-task/run-task
bhearsum: Sorry, I missed this earlier. Thanks for the explanation - sorry I didn't notice it in the other thread!

PR #631, opened 2025-01-22
docs: add to docker howto

jcristau:
Describe the `parent`, `definition` and `args` entries supported by the docker_image transforms.

### DIFF
--- a/docs/howto/docker.rst
+++ b/docs/howto/docker.rst
@@ -201,6 +201,65 @@ process the ``Dockerfile`` and handle the special syntax. Whereas the
 only generated once and then re-used by all subsequent pushes until the image
 is modified.
 
+Layering images on top of each other
+....................................
+
+The ``parent`` value on a docker-image task can refer to another image's ``name``.
+The parent image can then be referenced from the dependent image's Dockerfile
+as ``$DOCKER_IMAGE_PARENT``.  Whenever the base image's definition (or
+something it includes) changes, that automatically also triggers a
+rebuild of the dependent image.
+
+For example, ``kind.yml`` could include:
+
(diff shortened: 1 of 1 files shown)


JohanLorenzo (approved):
Great! Thanks for documenting this taskgraph-ism 🙂

To future readers: an external contributor asked a question about `DOCKER_IMAGE_PARENT` to @jcristau and me. That's what uncovered this lack of documentation.

docs/howto/docker.rst
```diff
@@ -201,6 +201,36 @@ process the ``Dockerfile`` and handle the special syntax. Whereas the
 only generated once and then re-used by all subsequent pushes until the image
 is modified.
 
+Layering images on top of each other
+....................................
+
+The `parent` value on a docker-image task can refer to another image's `name`.
+The parent image can then be referenced from the dependent image's Dockerfile
+as `$DOCKER_IMAGE_PARENT`.
```
ahal, line 209: An example showing both the task definitions + a Dockerfile snippet would be useful here.

ahal: Might also be nice to link to `build-decision` and maybe add a warning about `taskgraph build-image` not being able to process this.

ahal (approved):
Thanks, this is clearly an improvement so r+.

I do have some suggestions to make it even more useful though :)

docs/howto/docker.rst
jcristau: Updated, thanks for the suggestions.

A rendered version is at https://taskcluster-taskgraph--631.org.readthedocs.build/en/631/howto/docker.html FWIW.

PR #643, opened 2025-02-10
fix(attrmatch): use None to match missing attributes

ahal:


jcristau (commented):
This feels like a slightly too subtle incompatible change.  `None` used to only match `None`, now it also matches a missing attribute. That feels a bit scary, in terms of unintended consequences, plus it means there's no longer a way to match an attribute that's present but null?

ahal:
Yes, valid points. I did consider introducing some kind of other syntax, we could set it to a constant value like `missing`, or we could prefix the key with `!` or something like that. One complication here is that I want to be able to define the kwargs in yaml, so that precludes a sentinel object or the like.

One reason I thought this would be a good approach, is that the only place this is used (in Gecko and this repo at least), is in `from-deps.with-attributes`. And afaict we aren't setting anything to `None` there.

But we could implement something else. Do you have any preferences for handling the distinction?

ahal:
Had a chat with jcristau, I'm going to try to solve my use case by ensuring the attribute I'm working with is always set (to `None` instead of not existing). This would make this feature unnecessary for my use case.

Something like this might still be useful for other use cases, but we couldn't think of a good way to implement it that doesn't involve either tricky edge cases or writing a DSL.

PR #645, opened 2025-02-13
feat: port `taskgraph action` from gecko_taskgraph

Eijebong:
The documentation already says `You can generate actions.json locally by running taskgraph actions.` however that command was never ported over from gecko's `./mach taskgraph actions`

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -738,7 +738,7 @@ def actions(args):
         logging.root.setLevel(logging.DEBUG)
 
     try:
-        parameters = parameters_loader(args["parameters"])
+        parameters = parameters_loader(args["parameters"], strict=False)
         tgg = TaskGraphGenerator(root_dir=args.get("root"), parameters=parameters)
 
         actions = render_actions_json(tgg.parameters, tgg.graph_config, "DECISION-TASK")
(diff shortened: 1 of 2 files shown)


ahal (approved):
Thanks!

Eijebong:
Marking this back as a draft as after talking to @hneiva  about it, I discovered that there were indeed a few tests in `test_main` where the CLI was tested so I started writing some.

Eijebong:
Ah great, something is leaving the test env dirty.

`uv run -v pytest ./test/test_main.py` passes but `uv run -v pytest` doesn't

Eijebong:
Ah yeah, I see

https://github.com/taskcluster/taskgraph/blob/5200e65c5be059e77b927378a9c06c403a5e9f5e/test/test_actions_registry.py#L101 -> https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/actions/registry.py#L278 -> https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/actions/registry.py#L361 -> https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/actions/registry.py#L349

The modules declaring the actions get imported, the actions registered. Later when the second test runs, it mocks the `actions` list from the registry, then calls things to reload the actions. The modules being already imported by the previous test, it's a no-op, actions don't get re-registered, test fails. Fun

hneiva (approved):
LGTM!

PR #648, opened 2025-02-18
refactor!: revert {task_workdir} interpolation in run-task environment

ahal:
In hindsight, this ended up being a bad idea and caused lots of subtle issues in Gecko. As an alternative, I landed a change to get generic-worker itself to set $TASK_WORKDIR.

This will take awhile to propagate to all pools, but in the meantime it's cleanest to revert back to the abspath method.

### DIFF
--- a/docs/reference/migrations.rst
+++ b/docs/reference/migrations.rst
@@ -3,6 +3,14 @@ Migration Guide
 
 This page can help when migrating Taskgraph across major versions.
 
+13.x -> 14.x
+------------
+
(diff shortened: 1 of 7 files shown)


bhearsum (approved):

ahal:
Tested here:
https://treeherder.mozilla.org/jobs?repo=try&revision=e0091a203323d8df628b16b97604e3a94e138851

I see it working with both docker-worker and Windows generic-worker tasks.

PR #650, opened 2025-02-18
ci: add pre-commit check to keep uv.lock updated

ahal:


ahal:
So the pre-commit check is working fine.. but for some reason running `uv lock` is removing the version for "taskcluster-taskgraph", which presumably is causing this failure?

Not sure if this is a uv regression or if our CI isn't setup properly.

Eijebong:
Updating uv should do the trick. I'm not entirely sure what the issue is but the fact that `uv python install 3.13` installs a RC when there are 2 dot releases already is not great and most likely related.

image

ahal:
Well that worked, but looks like it uncovered a latent issue in the decision image task.

ahal:
Thanks confirmed. Specifically this commit:
https://github.com/astral-sh/uv/pull/7544/files

Looks like this might be a misconfiguration on our end that was previously failing silently, but is now loud.

Eijebong (approved):

PR #651, opened 2025-02-19
ci: create a 'complete-pr' task to block PRs from landing

ahal:


ahal:
Once this lands I'll update the required tasks to just point to `complete-pr`.

taskcluster/config.yml
```diff
@@ -38,3 +38,8 @@ workers:
             implementation: docker-worker
             os: linux
             worker-type: linux-gcp
+        succeed:
```
Eijebong, line 41: I'm curious why this is necessary? I didn't have to do this in https://github.com/mozilla-releng/balrog/commit/b593d465d3291aee3572b90cfd0b150922754c0d

ahal: Oh, maybe it's not... I copied it from Gecko.

ahal: I was looking for where you did that, but I couldn't find it

ahal: Yeah, looks like it's not necessary.

Eijebong (approved):

PR #652, opened 2025-02-21
fix: Replace deprecated datetime.utcnow() with timezone-aware alterna…

AnsahMohammad:
Replaced instances of datetime.datetime.utcnow() with datetime.datetime.now(datetime.UTC) to address deprecation warnings while running pytest.


jcristau (changes requested):
Thanks for the patch.

You'll need to use datetime.timezone.utc, as datetime.UTC was only added in python 3.11.

AnsahMohammad:
> Thanks for the patch.
> 
> You'll need to use datetime.timezone.utc, as datetime.UTC was only added in python 3.11.

Just did that, please check now

AnsahMohammad:
@jcristau Hey, I noticed that the `firefoxci-taskcluster / Decision Task` failed due to some time-formatting issues. However, the tests didn't catch any errors. Would you suggest extending the test coverage to include this as well?

jcristau:
> @jcristau Hey, I noticed that the `firefoxci-taskcluster / Decision Task` failed due to some time-formatting issues. However, the tests didn't catch any errors. Would you suggest extending the test coverage to include this as well?

Sounds like a good idea, yes :)

AnsahMohammad:
> > @jcristau Hey, I noticed that the `firefoxci-taskcluster / Decision Task` failed due to some time-formatting issues. However, the tests didn't catch any errors. Would you suggest extending the test coverage to include this as well?
> 
> Sounds like a good idea, yes :)

I went through the codebase and found this : 
```
            except KeyError:
                # go on to the next index path
                pass
```
https://github.com/taskcluster/taskgraph/blob/63ec7322aac955f80808b39c4d463052c384c476/src/taskgraph/optimize/strategies.py#L66


This made the test case miss the time-format issue on tests. I am not sure if it'll be a good idea to replace the "pass" statement here as any error might break the flow. I am confused on how to proceed.

jcristau:
@AnsahMohammad I've pushed a commit to your branch to add a test and fix the issue that was breaking the decision task.

AnsahMohammad:
> @AnsahMohammad I've pushed a commit to your branch to add a test and fix the issue that was breaking the decision task.

Thanks for the commit :+1:.
I thought it was another file causing the error.

jcristau:
Updated to pull in #659 and trigger the required `complete-pr` task.

test/test_optimize.py
```diff
@@ -28,9 +28,16 @@ def should_remove_task(self, task, params, arg):
...
     def should_replace_task(self, task, params, deadline, taskid):
-        expires = datetime.utcnow() + timedelta(days=1)
-        if deadline and expires < datetime.strptime(deadline, "%Y-%m-%dT%H:%M:%S.%fZ"):
-            return False
+        expires = datetime.datetime.now(datetime.timezone.utc) + datetime.timedelta(
+            days=1
+        )
+        if deadline:
+            deadline_dt = datetime.datetime.strptime(
+                deadline.replace("+00:00Z", "Z"), "%Y-%m-%dT%H:%M:%S.%fZ"
```
jcristau, line 36: We shouldn't expect `+00:00Z` as a suffix in `deadline` here, that's invalid.  It's not clear to me what this is trying to achieve.

AnsahMohammad: Looks like i over-complicated things, this can be easily done instead using `tzinfo=None`.

jcristau:
Added the "breaking change" label because both `json_time_from_now` and `current_json_time` with `datetime_format=True` now return a timezone-aware object.

jcristau (approved):

PR #653, opened 2025-02-22
feat: Allow task-priority to be keyed by level

### LINKED ISSUE #252: Allow `task-priority` to be keyed by level
Currently the `task-priority` key in the graph config can be keyed by project, which is mostly only useful for Gecko. For non-Gecko projects, it would be useful to be able to key this by level (and possibly other values) as well.

AnsahMohammad:
Allow 'level' as a key in task-priority config.
Extend the test_optionally_keyed_by function to include validation for multiple keys (foo, bar).

closes #252


ahal (approved):
Thanks for the test!

PR #660, opened 2025-03-04
feat: support git shallow clones

### LINKED ISSUE #503: [run-task] Support Git shallow clones
Supporting shallow clones in run-task would be a quick and easy win. This should probably even be the default behaviour.

AnsahMohammad:
Support the git shallow clones using --depth=1 flag.

This PR fixes #503

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -600,6 +600,7 @@ def git_checkout(
     commit: Optional[str],
     ssh_key_file: Optional[Path],
     ssh_known_hosts_file: Optional[Path],
+    shallow: bool = True,
 ):
     env = {
         # abort if transfer speed is lower than 1kB/s for 1 minute
@@ -639,6 +640,9 @@ def git_checkout(
             base_repo if base_repo else head_repo,
             destination_path,
         ]
+
+        if shallow:
+            args.append("--depth=1")
 
         retry_required_command(b"vcs", args, extra_env=env)
 


AnsahMohammad:
@ahal  Can you take a look at this change?

ahal (changes requested):
Thanks for the PR!

This looks like a good start, but the `git_checkout` function is internal to `run-task`, so there's no way for a user to request a shallow clone here. You'll need to trace backwards, forwarding the argument all the way until the `main` function, where you'll need to create a new command line argument for this.

It would also be much appreciated if you could update a test to cover this new flag somewhere in `test/test_scripts_run_task.py` :)

AnsahMohammad:
> This looks like a good start, but the `git_checkout` function is internal to `run-task`, so there's no way for a user to request a shallow clone here. You'll need to trace backwards, forwarding the argument all the way until the `main` function, where you'll need to create a new command line argument for this.
> 
> It would also be much appreciated if you could update a test to cover this new flag somewhere in `test/test_scripts_run_task.py` :)

Oh yeah, good point—I didn’t think about that.
Sure i'll work on that !

src/taskgraph/run-task/run-task
```diff
@@ -978,6 +985,7 @@ def vcs_checkout_from_args(options):
                 revision,
                 ssh_key_file,
                 ssh_known_hosts_file,
+                options.get("shallow", True),
```
ahal, line 988: Let's default to `False` to start. This could potentially cause a lot of backwards compatibility issues, and I'd rather prove out the feature in a handful of tasks that opt into it first. We can file a follow-up to make shallow clones the default later.

src/taskgraph/run-task/run-task
```diff
@@ -1061,6 +1069,9 @@ def main(args):
     parser.add_argument("--user", default="worker", help="user to run as")
     parser.add_argument("--group", default="worker", help="group to run as")
     parser.add_argument("--task-cwd", help="directory to run the provided command in")
+    parser.add_argument("--shallow", type=lambda x: x.lower() == "true",
```
ahal, line 1072: Instead of using a lambda with `type`, you should specify `action="store_true"`. This will allow you to pass `--shallow` without any argument at all.

test/test_scripts_run_task.py
```diff
@@ -354,13 +356,14 @@ def _commit_file(message, filename):
...
-        (None, "main", ["mainfile"], "main"),
-        (None, "mybranch", ["mainfile", "branchfile"], "branch"),
-        ("main", "main", ["mainfile"], "main"),
-        ("main", "mybranch", ["mainfile", "branchfile"], "branch"),
+        (None, None, ["mainfile"], "main", True),
+        (None, "main", ["mainfile"], "main", False),
+        (None, "mybranch", ["mainfile", "branchfile"], "branch", True),
+        ("main", "main", ["mainfile"], "main", False),
+        ("main", "mybranch", ["mainfile", "branchfile"], "branch", True),
+        ("main", "mybranch", ["mainfile", "branchfile"], "branch", False),
```
ahal, line 366: Let's keep the parameter set to `False` for all the existing tests, and then add new one where it is `True`. I wouldn't want to accidentally change the intent of one of the original tests.

ahal (changes requested):
Thanks! This is close, have a few more minor requests though.

AnsahMohammad:
> Thanks! This is close, have a few more minor requests though.

I've addressed the comments, please review again.

src/taskgraph/run-task/run-task
```diff
@@ -1061,6 +1069,9 @@ def main(args):
     parser.add_argument("--user", default="worker", help="user to run as")
     parser.add_argument("--group", default="worker", help="group to run as")
     parser.add_argument("--task-cwd", help="directory to run the provided command in")
+    parser.add_argument("--shallow", action="store_true",
+        default=False, help="perform a shallow clone (default: false)",
+    )
```
ahal, line 1074: I'm really sorry, I told you to add an argument here. But on closer look, I was wrong and instead we should add this in `collect_vcs_options` using an environment variable like most of the other vcs arguments. You should add something like:
```
shallow = os.environ.get("%s_SHALLOW_CLONE" % env_prefix)
```

and make sure to return it (like you already are).

ahal (changes requested):
Apologies for the delayed review!

Issue #67, opened 2022-06-22
Improve performance of `test_util_vcs.py`


ahal:
This test takes awhile, and since we run it across multiple Python versions, is the main reason why the `unit` task takes so long to run. It would be nice to figure out why it's so slow and see if we can speed it up at all.

miladHakimi:
Hi @ahal,

I ran a profiler (pytest-profiling) on the tests and figured out that the `subprocess.check_output()` function is the main contributer to the runtime. 

https://github.com/taskcluster/taskgraph/blob/311cf5255ff8785beb9b4c3a9ccec4a50d3a7ba8/src/taskgraph/util/vcs.py#L40-L51

It is called by the `Repository.run()` method to run different commands like `add`, `commit`, `reset`, etc. Here is the output of the profiler:
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       52    0.001    0.000   22.405    0.431 runner.py:109(pytest_runtest_protocol)
       52    0.002    0.000   22.392    0.431 runner.py:117(runtestprotocol)
      156    0.001    0.000   22.389    0.144 runner.py:217(call_and_report)
 2123/572    0.005    0.000   22.381    0.039 _hooks.py:244(__call__)
 2123/572    0.002    0.000   22.379    0.039 _manager.py:77(_hookexec)
 2123/572    0.015    0.000   22.378    0.039 _callers.py:9(_multicall)
      156    0.001    0.000   22.357    0.143 runner.py:245(call_runtest_hook)
      156    0.002    0.000   22.355    0.143 runner.py:316(from_call)
      156    0.001    0.000   22.352    0.143 runner.py:260(<lambda>)
      321    0.023    0.000   21.337    0.066 subprocess.py:417(check_output)
      321    0.030    0.000   21.313    0.066 subprocess.py:506(run)
      313    0.004    0.000   20.940    0.067 vcs.py:40(run)
      321    0.021    0.000   20.673    0.064 subprocess.py:1163(communicate)
      623   19.903    0.032   19.909    0.032 {method 'read' of '_io.TextIOWrapper' objects}
       52    0.000    0.000   19.403    0.373 runner.py:158(pytest_runtest_call)
       52    0.000    0.000   19.401    0.373 python.py:1787(runtest)
       52    0.001    0.000   19.398    0.373 python.py:188(pytest_pyfunc_call)
       32    0.001    0.000    4.690    0.147 vcs.py:254(get_changed_files)
[...]

PR #662, opened 2025-03-09
perf: Improve test performance by parallelizing pytest

AnsahMohammad:
While working on #67 , discovered that `pytest -n X` enables parallel test execution, where X is the number of cores. This can significantly reduce test execution time.

Though it didn’t resolve the issue in #67, it could substantially speed-up automated tests. 

image
On my 8-core machine, runtime dropped from 43.67s to 20.07s.

### DIFF
--- a/taskcluster/kinds/test/kind.yml
+++ b/taskcluster/kinds/test/kind.yml
@@ -54,4 +54,4 @@ tasks:
             symbol: unit(py{matrix[python]})
         run:
             command: >-
-                uv run coverage run --data-file /builds/worker/artifacts/coverage --context=py{matrix[python]} -m pytest -vv -n auto
+                uv run coverage run --data-file /builds/worker/artifacts/coverage --context=py{matrix[python]} -m pytest -vv -n auto --dist=loadfile


ahal:
Thanks, this seems reasonable to me!

However I'm not sure why those tests are failing.. Normally I'd dismiss the failures as not related to a change like this, but it's weird that only the py310 tests pass. So that leads me to think that it *is* something to do with the parallelization. Very strange.

I think we'll need to sort this before landing, this failure isn't happening on the main branch.

ahal (changes requested):
Requesting changes to investigate the test issue. If you get stuck, I can poke around when I have some time, though not sure when that will be

AnsahMohammad:
Thanks for reviewing! I’ll look into the test failures and see if the parallelization is causing issues. 

> Requesting changes to investigate the test issue. If you get stuck, I can poke around when I have some time, though not sure when that will be

Thanks, I'll reach out if i get stuck

AnsahMohammad:
Looks like the issue is definitely coming from parallelization since the failures changed when I tweaked the distribution strategy :/ . I'll try to dig deep and see if its fixable.

PR #664, opened 2025-03-25
Only consider tracked files when building toolchain hashes

### LINKED ISSUE #663: _get_all_files in util/hash.py very slow on large repos
`_get_all_files` in `util/hash.py` takes 35 seconds to call once in a local checkout of the [mozilla/translations](https://github.com/mozilla/translations) repo. My vcs_root here is full of lots of stuff. This eagerly enumerates all of the .git directory, my .gitignore files like data files, and my git submodules, and my python virtual environments. This is currently 251,192 files.

screenshot of snakeviz showing _get_all_files taking 35 seconds.

I'm not sure the requirements for other projects for this fucntion, but when I swapped out the implementation with `git ls-files` the result took 3.5 seconds.

```py
@functools.lru_cache(maxsize=None)
def _get_all_files(base_path: str) -> list[str]:
    try:
        output = subprocess.check_output(
            ["git", "ls-files",
                # Include tracked files.
                "--cached",
                # Include untracked files.
                "--others",
                # Respect the .gitignore
                "--exclude-standard"],
            cwd=Path(base_path).resolve(),
            text=True,
        )
        return [
            mozpath.normsep(line.strip())
            for line in output.strip().splitlines()
            if line.strip()
        ]
    except subprocess.CalledProcessError:
        # Fall back to rglob if the git command fails.
        pass

    return [
        mozpath.normsep(str(path))
        for path in Path(base_path).rglob("*")
        if path.is_file()
    ]
```

ahal:
Previously we were walking the file system and were e.g scanning files in the `.git` directory. This can be quite slow on larger repos.

Fixes #663

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -477,6 +477,7 @@ def show_taskgraph(options):
             print(f"Generating {options['graph_attr']} @ {base_rev}", file=sys.stderr)
             ret |= generate_taskgraph(options, parameters, overrides, logdir)
         finally:
+            assert cur_rev
             repo.update(cur_rev)
 
         # Generate diff(s)
--- a/src/taskgraph/util/hash.py
+++ b/src/taskgraph/util/hash.py
@@ -4,9 +4,10 @@
 
 import functools
 import hashlib
-from pathlib import Path
+import os
 
 from taskgraph.util import path as mozpath
+from taskgraph.util.vcs import get_repository
 
 
 @functools.lru_cache(maxsize=None)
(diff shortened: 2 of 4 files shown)


jcristau:
@glandium fyi

ahal:
Note that `gecko_taskgraph` already does this, but it uses `FileFinder` so I couldn't port the fix exactly.

bhearsum (approved):

gregtatum:
Thanks for the work on this, this will be a big quality of life improvement for translations!

PR #666, opened 2025-04-02
Let generic-worker verify run-task/fetch-content integrity

jcristau:
The decision task knows the contents of these scripts so it can pass their checksums to downstream tasks.


jcristau:
AIUI this should work with generic-worker >= 10.8.0 (which included https://bugzilla.mozilla.org/show_bug.cgi?id=1459376)

ahal (approved):
Sorry this slipped, lgtm!

PR #669, opened 2025-04-03
fix(vcs): remove empty string from command in repo.get_tracked_files()

ahal:
The test should be exercising this code path, so I'm not really sure why it passes in Taskgraph but failed in another case in CI. Maybe it's related to the version of Git on the worker.


jcristau (commented):
I don't understand why this would make a difference?

jcristau:
That part I can understand, but I don't see how this patch changes that.

src/taskgraph/util/vcs.py
```diff
@@ -477,7 +480,10 @@ def get_commit_message(self, revision=None):
 
     def get_tracked_files(self, *paths, rev=None):
         rev = rev or "HEAD"
-        return self.run("ls-tree", "-r", "--name-only", rev, *paths).splitlines()
+        cmd = ["ls-tree", "-r", "--name-only", rev]
+        if paths:
+            cmd.extend(paths)
+        return self.run(*cmd).splitlines()
```
hneiva, line 486: To Julien's point, this isn't removing any empty strings:
```
>>> a = []
>>> [1,2,3,*a]
[1, 2, 3]

>>> a = ['']
>>> [1,2,3,*a]
[1, 2, 3, '']
```
If the idea is to filter out, you'd probably need something like `[p for p in paths if p]`

jcristau: Indeed.  (Also, it'd be good to understand/document why/how an empty string would come in here, naively that seems like it'd be a bug elsewhere.)

ahal: I decided to strip out empty strings rather than make `graph_config.vcs_path` absolute to avoid causing issues elsewhere.

jcristau (approved):
I'm not sure I like the semantics here (vs fixing the caller to not pass the empty string) but it's probably fine.

bhearsum:
I dug into this a bit more. AFAICT the only call to `get_tracked_files` is from [hash.py](https://github.com/taskcluster/taskgraph/blob/d3829d219c0c78571f2b9affd96b65faf9282f40/src/taskgraph/util/hash.py#L57). The only call to _that_ is from [toolchain.py's `get_digest_data`](https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/transforms/run/toolchain.py#L65), which uses `config.graph_config.vcs_root` for the `base_path` that eventually makes it to `get_tracked_files`.

For typical cases [`vcs_root` defaults to `GraphConfig.root_dir`](https://github.com/taskcluster/taskgraph/blob/d3829d219c0c78571f2b9affd96b65faf9282f40/src/taskgraph/config.py#L148). This is usually [set correctly when load_graph_config is used](https://github.com/taskcluster/taskgraph/blob/d3829d219c0c78571f2b9affd96b65faf9282f40/src/taskgraph/config.py#L159-L168), but unfortunately some code paths (such as [decision tasks](https://github.com/taskcluster/taskgraph/blob/d3829d219c0c78571f2b9affd96b65faf9282f40/src/taskgraph/decision.py#L106-L108)) end up setting `root_dir` as None. (This doesn't break this repo, because there are no toolchain tasks...but it will break Gecko and other repos.)

It's somewhat tangential to this specific issue, I guess, but it seems like we ought to make sure `root_dir` is set in all of the common cases; I'll work up a patch for this.

PR #673, opened 2025-04-07
fix: ensure GraphConfig.root_dir is an absolute path

bhearsum:


src/taskgraph/config.py
```diff
@@ -157,6 +157,9 @@ def validate_graph_config(config):
 
 
 def load_graph_config(root_dir):
+    # ensure we have an absolute path; this is required for assumptions
+    # made later, such as the `vcs_root` being a directory above `root_dir`
+    root_dir = os.path.abspath(root_dir)
```
ahal, line 162: In theory, I think this should go in the `__post_init__` method of `GraphConfig`. In practice, I don't think it really matters. Given GraphConfig is frozen, it would mean we'd need to use `object.__setattr__` to set it, so this is probably fine.

ahal (approved):

src/taskgraph/config.py
bhearsum: ah; but if there's ever anything else that creates a `GraphConfig` that could be a problem, I think? This seems like an easy change to make.

ahal: Thanks!
https://stackoverflow.com/a/73718478

bhearsum:
failures are codecov (meh) and pre-commit (issue on their side).

PR #675, opened 2025-04-09
fix: ensure GraphConfig.vcs_root is set when root_dir is a relative path

bhearsum:
Related/alternative to https://github.com/taskcluster/taskgraph/pull/669 and https://github.com/taskcluster/taskgraph/pull/673.


ahal (approved):

bhearsum:
No longer necessary because of https://github.com/taskcluster/taskgraph/pull/673

PR #678, opened 2025-04-11
Add a workflow to run auto updates for pre-commit

Eijebong:
This replaces the pre-commit.ci app's autoupdate.

The one caveat with this is that it doesn't check for PRs that are already opened on that branch. I did that to keep it as simple as possible. It's not a real issue as github will detect that and close the previous PR and then open a new one but it might be a bit spammy if we don't merge those PRs

### DIFF
--- a/.github/workflows/pre-commit-autoupdate.yml
+++ b/.github/workflows/pre-commit-autoupdate.yml
@@ -1,4 +1,4 @@
-name: Autoupdate
+name: Auto update pre-commit hooks every month
 on:
   schedule:
     - cron: '0 0 1 * *'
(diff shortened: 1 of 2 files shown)


Eijebong:
example of it working.

I have this repo where I ran the job on push instead of cron. 

https://github.com/Eijebong/test-pr-gha

.github/workflows/pre-commit-autoupdate.yml
```diff
@@ -0,0 +1,36 @@
...
+          git config --global user.email "releng+pre-commit-autoupdate@mozilla.com"
+      - uses: actions/setup-python@v5
+      - run: python -m pip install pre-commit
+        shell: bash
+      - name: Update files
+        run: |
+          git checkout -B pre-commit-update
+          pre-commit autoupdate -j4
+          git add .pre-commit-config.yaml
+          git commit -am "Update pre-commit"
```
ahal, line 26: Would it be too annoying to parse:
https://github.com/taskcluster/taskgraph/blob/main/.pre-commit-config.yaml#L6

instead? Or at the very least remove that config (and I guess the "schedule" one too). Fwiw, I like the configured commit message better.

ahal:
I somewhat think it wouldn't be much harder to do this as a Taskcluster cron, but also that it's not worth switching now

.github/workflows/pre-commit-autoupdate.yml
Eijebong: Changed it to use `ci.autoupdate_commit_msg` if it exists and removed keys that were only used by the app

Eijebong: I'll use the same for the PR title actually since we squash things

ahal:
> The advantage of this is that we don't have to worry about the token as github provides it for us in the workflow.

True!

One trick we could use is that because `taskcluster-github` is guaranteed to be installed and it already has read&write PR access, we could authenticate as that app. We'd have to be comfortable storing its private key in a TC secret though. And it's definitely more complex than "it just works out of the box" ><

Eijebong:
> One trick we could use is that because `taskcluster-github` is guaranteed to be installed and it already has read&write PR access

I think it's something I have mentioned previously at a community meeting. We already have a way of [creating comments](https://firefox-ci-tc.services.mozilla.com/docs/reference/integrations/github/api#createComment) through said app, it would be nice to expand on that capability

ahal (approved):

PR #680, opened 2025-04-29
feat: implement new 'load-task' command to debug tasks locally

ahal:
This command is an extension of the `load-image` command. Given a taskId, it will:

1. Find and download the docker image from the parent task.
2. Spin up a docker container using this image.
3. Ensure environment variables in the task's definition are set.
4. Run the "setup" portion of `run-task` (checkouts, fetches, etc)
5. Drop the user into a bash shell.

This will make it easier to debug tasks locally. Note for now only tasks using run-task with a docker-worker (or D2G) payload are supported.

### DIFF
--- a/src/taskgraph/docker.py
+++ b/src/taskgraph/docker.py
@@ -5,10 +5,13 @@
 
 import json
 import os
+import shlex
 import subprocess
 import tarfile
+import tempfile
 from io import BytesIO
 from textwrap import dedent
+from typing import List, Optional
 
 try:
     import zstandard as zstd
@@ -19,6 +22,7 @@ from taskgraph.util import docker
 from taskgraph.util.taskcluster import (
     get_artifact_url,
     get_session,
+    get_task_definition,
 )
 
 DEPLOY_WARNING = """
@@ -90,7 +94,7 @@ def load_image_by_task_id(task_id, tag=None):
     else:
         tag = "{}:{}".format(result["image"], result["tag"])
     print(f"Try: docker run -ti --rm {tag} bash")
-    return True
+    return tag
 
 
 def build_context(name, outputFile, args=None):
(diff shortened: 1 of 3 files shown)


src/taskgraph/docker.py
```diff
@@ -237,3 +239,59 @@ def download_and_modify_image():
...
+    if not command or not command[0].endswith("run-task"):
+        print("Only tasks using `run-task` are supported!")
+        return 1
+
+    # Remove the payload section of the task's command. This way run-task will
+    # set up the task (clone repos, download fetches, etc) but won't actually
+    # start the core of the task. Instead we'll drop the user into a command
+    # shell.
+    if (index := command.index("--")) > -1:
+        command[index + 1 :] = ["echo 'Task setup complete!'"]
```
jcristau, line 263: Would it be useful to stash the rest of the command in an env var or something?  So that you could e.g. `eval $ORIGINAL_COMMAND` from the shell.

src/taskgraph/docker.py
```diff
@@ -237,3 +239,59 @@ def download_and_modify_image():
...
+
+    image_task_id = task_def["payload"]["image"]["taskId"]
+    image_tag = load_image_by_task_id(image_task_id, None)
```
jcristau, line 266: nit: None is the default

src/taskgraph/docker.py
```diff
@@ -237,3 +239,59 @@ def download_and_modify_image():
...
+    tmpfile = None
+    try:
+        command = [
+            "docker",
+            "run",
+            "-it",
+            image_tag,
+            "bash",
+            "-c",
+            f"{' '.join(command)} && bash",
```
jcristau, line 279: Initially on reading this I was worried about quoting corner cases here, but I guess since we only keep the run-task part of the command there's not actually much arbitrary input to take into account...  Consider using `shlex.join(command)` though?

src/taskgraph/main.py
```diff
@@ -652,6 +652,31 @@ def image_digest(args):
...
+    "load-task",
+    help="Loads a pre-built Docker image and drops you into a container with "
+    "the same environment variables and run-task setup as the specified task. "
+    "The task's payload.command will be replaced with 'bash'. You need to have "
+    "docker installed and running for this to work.",
+)
+@argument(
+    "task_id",
+    help="Load the image at public/image.tar.zst in this task, "
+    "rather than searching the index",
```
jcristau, line 665: "searching the index" doesn't seem to be an option here, so maybe remove that part?

src/taskgraph/main.py
```diff
@@ -652,6 +652,31 @@ def image_digest(args):
...
+)
+@argument(
+    "--rm",
+    action="store_true",
+    default=False,
```
jcristau, line 670: I'd probably argue `--rm` should be the default?

jcristau (commented):
Very cool, thank you!

src/taskgraph/docker.py
ahal: Good idea!

src/taskgraph/main.py
ahal: Oops, this was copy/pasted from the `load-image` command and I neglected to update the description.

PR #683, opened 2025-05-06
Fix a couple issues with `taskgraph load-task`

ahal:


ahal:
Hm, sudo isn't going to exist in all images.. will need to re-think this..

bhearsum (approved):

PR #692, opened 2025-05-13
test: validate the output of run-task transforms against the schema

jcristau:


Eijebong (approved):
Nice, even better. Thank you

PR #695, opened 2025-05-23
docs: improvements to the getting-started tutorial

ahal:
This does two things:

1. Removes mentions of the communitytc instance. We don't want to direct new projects over to the community instance at this time.

2. Re-orders the steps. Previously the tutorial said to connect your repo to Taskcluster *and then* add tasks, but this is the wrong order because the repo needs to have a `tc.yml` file in order for fxci-config generation to succeed. So the order is now:

   a. Create tasks b. Request Scopes c. Connect to Github


hneiva (approved):
LGTM!

PR #698, opened 2025-05-29
chore: version bump 14.4.0

bhearsum:


Eijebong:
Would be nice to wait a bit and get https://github.com/taskcluster/taskgraph/pull/700 in there too if possible so we can update it in firefox-ios because I feel like that issue is going to be back as I'm assuming those commit messages hyphens are automatically converted by macos.

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Change Log
 
+## [14.4.0] - 2025-05-29
```
Eijebong, line 3: Nit: date is wrong

Eijebong (approved):

PR #703, opened 2025-06-04
Update file_url() handling of Thunderbird comm paths

coreycb:
In commit df291c6e31b69daa2c9cb85879a2872f2d7480ca, root_dir was changed from a relative path to an absolute path, resulting in file_url() no longer detecting the path correctly as it didn't start with "comm/". This change allows file_url() to correctly detect and handle an absolute path that includes "comm/".


ahal (approved):
Thanks!

PR #706, opened 2025-06-12
Set the version in pyproject.toml instead of having it dynamic

Eijebong:
This reduces complexity of the build and should allow us to use dependabot with uv. Also standardizes releases a bit more as the version isn't hidden anymore in the __init__.py file


ahal (approved):
Thanks!

PR #709, opened 2025-06-18
Convert schema comments to descriptions

ahal:
This ensures they show up in the docs.


jcristau (approved):
More of a rubberstamp than anything else, but this looks plausible.

PR #722, opened 2025-07-17
Check for slash in route

### LINKED ISSUE #254: taskgraph should raise and exception if index routes have a `/` in them
According to the Taskcluster documentation, [slashes are forbidden in index routes](https://docs.taskcluster.net/docs/reference/core/index#indexing-routes):
> Note that slashes (/) aren’t allowed in index paths.

Taskgraph will happily create them though, and when these routes are used as part of caches, it ends up with cache routes that never work (and tasks that never get optimized by replacement).

abhishekmadan30:
Makes an exception be thrown if a route contains a "/" and added tests for it. Closes #254

### DIFF
--- a/src/taskgraph/util/verify.py
+++ b/src/taskgraph/util/verify.py
@@ -172,11 +172,6 @@ def verify_trust_domain_v2_routes(
 
     for route in routes:
         if route.startswith(route_prefix):
-            # Check for invalid / in the route
-            if "/" in route:
-                raise Exception(
-                    f"{task.label} has invalid route with forward slash: {route}"
-                )
             if route in scratch_pad:
                 raise Exception(
                     f"conflict between {task.label}:{scratch_pad[route]} for route: {route}"
@@ -214,11 +209,6 @@ def verify_routes_notification_filters(
 
     for route in routes:
         if route.startswith(route_prefix):
-            # Check for invalid / in the route
-            if "/" in route:
-                raise Exception(
-                    f"{task.label} has invalid route with forward slash: {route}"
-                )
             # Get the filter of the route
             route_filter = route.split(".")[-1]
             if route_filter not in valid_filters:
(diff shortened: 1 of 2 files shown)


src/taskgraph/util/verify.py
```diff
@@ -206,9 +206,17 @@ def verify_routes_notification_filters(
...
     task_dict = task.task
     routes = task_dict.get("routes", [])
+    print(f"DEBUG: Task {task.label} routes: {routes}")  
 
     for route in routes:
+        print(f"DEBUG: Checking route: {route}")  
         if route.startswith(route_prefix):
+            # Check for invalid characters in the route
+            if "/" in route:
+                print(f"DEBUG: Found slash in route: {route}")
```
hneiva, line 216: use `logger.debug` or `logger.error` instead of `print`

hneiva (approved):
Just a minor change, but LGTM otherwise.

hneiva (changes requested):
Oops, didn't notice tests are failing

bhearsum:
Ah; it looks like the checks I had to approve to run caught a linting issue. Once that's fixed we can get this merged.

src/taskgraph/util/verify.py
```diff
@@ -223,6 +223,25 @@ def verify_routes_notification_filters(
...
+    This function ensures that routes do not contain forward slashes.
+    """
+    if task is None:
+        return
+    task_dict = task.task
+    routes = task_dict.get("routes", [])
+
+    for route in routes:
+        # Check for invalid / in the route
+        if "/" in route:
```
ahal, line 240: Sorry for the late request.. but it just occurred to me that only *index* routes can't have a slash. Routes in general are allowed to have a slash.

So we should change this to `if route.startswith("index.") and "/" in route:`. Also updating the function name to `verify_index_routes..` or similar.

ahal (changes requested):

ahal (approved):
Thanks!

PR #723, opened 2025-07-22
fix: make `hash_paths` work when `base_path` is set

bhearsum:
Since #664 we started returning paths relative to the repository root from `_find_matching_files`. This meant that when we join `base_path` with `path` in `hash_paths`, we end up with `base_path` in there twice.

Fixing this without breaking `mozpath.match` means we need to join together the repository root and matched files after `mozpatch.match`. This, in turn, requires that some tests are able to call `get_repository`, which requires faking a repository being present.

As far as I can tell, this hasn't worked since #664, and possibly before that. We have very little usage of this function though, and I couldn't find any that used directories.


ahal (approved):
Thanks, lgtm

Issue #21, opened 2022-04-25
Fix false assumptions in 'build_docker_worker_payload'


ahal:
We call `util.docker.parse_volumes` in the `build_docker_worker_payload` function
https://github.com/taskcluster/taskgraph/blob/4f589d8ff995f90becb53d3841ac701da1afc662/src/taskgraph/transforms/task.py#L578

This in turn makes several false assumptions, namely that the root dir is `taskcluster/ci` and that there exists a kind called `docker-image`:
https://github.com/taskcluster/taskgraph/blob/4f589d8ff995f90becb53d3841ac701da1afc662/src/taskgraph/util/docker.py#L15
https://github.com/taskcluster/taskgraph/blob/4f589d8ff995f90becb53d3841ac701da1afc662/src/taskgraph/util/docker.py#L211

Neither of these things are necessarily true. Rather than re-parsing the docker image's `kind.yml` file, we should store the path to the Dockerfile as an attribute (or tag) on the docker image task. Then we should move the check from the first link to a "verification" in `util/verify.py`. This way we'll be able to access the path to the Dockerfile directly from the dependency.

ahal:
Oops, I didn't use permalinks and the urls above were all wrong. Just fixed them.

To re-phrase this bug a bit, the problem is in the last two links. For example, in that last one we have `load_yaml("taskcluster", "kinds", "docker-image", "kind.yml")`. The root directory (`taskcluster`) is just a convention, so the fact that we're hardcoding it here, means any repo that has it configured in this non-standard place is going to break.

The kind name (`docker-image`) is also just a convention, so if a repo uses a different kind name for their docker-images, this will also break.

PR #728, opened 2025-07-28
fix: removed hardcoding from docker-image dir and kinds

abhishekmadan30:
Closes https://github.com/taskcluster/taskgraph/issues/21

### DIFF
--- a/taskcluster/config.yml
+++ b/taskcluster/config.yml
@@ -14,6 +14,8 @@ index:
 
 task-priority: low
 
+docker-image-kind: docker-image
+
 taskgraph:
     register: self_taskgraph:register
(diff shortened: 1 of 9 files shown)


taskcluster/config.yml
```diff
@@ -14,6 +14,8 @@ index:
 
 task-priority: low
 
+docker-image-kind: docker-image
```
ahal, line 17: I would lean towards not defining this unless we actually want to use a non-default name, but doesn't really matter I guess.

abhishekmadan30:
I updated the where the default - docker-image-kind is decide as the optional + default combinations doesn't work as schema(obj) is not called on the graph config schema so the fix I had was to add a try catch field in images_path as that is only place it would be called

PR #729, opened 2025-07-29
fix: removed exception for vcs root property 

### LINKED ISSUE #522: Taskgraph fails with non-standard root dir
Taskgraph supports a `--root` flag which supposedly allows you to place the Taskgraph files into an arbitrary directory. However, due to this check:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/config.py#L130

Doing so results in a guaranteed exception. I'd be shocked if this were the only issue. We should either support this use case or remove the flag. This came up in https://github.com/mozilla/neqo because @larseggert would like to keep the CI files in the `.taskcluster/` dir to help keep them hidden from the rest of the repo files.

abhishekmadan30:
Removing this exception should all for non-standard root directories to work. Closes #522. I am not sure exactly what this #522 is asking for, please correct me if interpretation was wrong

### DIFF
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -11,6 +11,7 @@ from typing import Dict
 
 from voluptuous import ALLOW_EXTRA, All, Any, Extra, Length, Optional, Required
 
+from .util import path
 from .util.caches import CACHES
 from .util.python_path import find_object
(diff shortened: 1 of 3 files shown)


ahal (approved):
Thanks!

PR #730, opened 2025-07-31
fix: dump subprocesses python versions

### LINKED ISSUE #7: Dump Python and other runtime info in run-task

abhishekmadan30:
Closes #7

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -1033,6 +1033,23 @@ def _display_python_version():
         b"setup", b"Python version: %s\n" % platform.python_version().encode("utf-8")
     )
 
+def _display_subprocess_python_version():
+
+    python_versions = ["python","python3"]
+
+    for python_executable in python_versions:
+        try:
+            version = subprocess.check_output(
+                [python_executable, "-c", "import platform; print(platform.python_version())"],
+            ).strip()
+
+            print_line(
+                b"setup", b"Subprocess %s version: %s\n" % (python_executable.encode(), version)
+            )
+
+        except FileNotFoundError:
+            pass
+
 
 def main(args):
     task_workdir = os.environ["TASK_WORKDIR"] = os.getcwd()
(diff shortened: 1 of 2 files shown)


src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,24 @@ def _display_python_version():
...
+
+    for python_executable in python_versions:
+        try:
+            version = subprocess.check_output(
+                [python_executable, "-c", "import platform; print(platform.python_version())"],
+                text=True
+            ).strip()
+
+            print_line(
+                b"setup", b"Subprocess %s version: %s\n" % (python_executable.encode(), version.encode("utf-8"))
```
ahal, line 1048: nit: If you remove the `text=True` then `version` will be a byte string already and you won't need to re-encode it.

src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,24 @@ def _display_python_version():
         b"setup", b"Python version: %s\n" % platform.python_version().encode("utf-8")
     )
 
+def _display_subprocess_python_version():
+
+    python_versions = ["python","python2","python3"]
```
ahal, line 1038: Let's not dump `python2`, I sure hope it wouldn't be used, so I think logging about it will just add confusion.

ahal (changes requested):
Nice, I like this line of thinking!

Initially I was going to suggest an output format that includes the Python binary name, file path and version.. but then I realized we can use `uv` in the `run-task` script.

So instead of messing with all this, you can run `uv python list --only-installed`. This command will output exactly what I was thinking and more. Plus it'll find Pythons that aren't even on the $PATH!

It's possible for `run-task` to run in a context where `uv` isn't installed, so you'll have to have a check for that (see the other place in `run-task` that already uses `uv` to see how), I think in this case it's fine to just omit the extra logging.

abhishekmadan30:
@ahal ,  I also create another pr with the UV command as the main check, let me know which one would be better : https://github.com/taskcluster/taskgraph/pull/733/files

ahal (changes requested):
I think we should go with the `uv` version

PR #731, opened 2025-07-31
feat: add a commandline flag to disable verifications

ahal:
Bug: 1980436


src/taskgraph/generator.py
```diff
@@ -437,9 +439,11 @@ def _run_until(self, name):
...
         return self._run_results[name]
 
-    def verify(self, name, obj, *args, **kwargs):
-        verifications(name, obj, *args, **kwargs)
-        return name, obj
+    def verify(self, name, *args, **kwargs):
+        if self._enable_verifications:
+            verifications(name, *args, **kwargs)
+        if args:
+            return name, args[0]
```
hneiva, line 446: Why does it return anything?

PR #733, opened 2025-08-01
used uv command to check python version

### LINKED ISSUE #7: Dump Python and other runtime info in run-task

abhishekmadan30:
Also would close #7


src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,33 @@ def _display_python_version():
...
 
+def _display_subprocess_python_version():
+    try:
+        output = subprocess.check_output(
+            ["uv", "python", "list", "--only-installed"],
+        ).strip()    
+
+        if output:
+            for line in output.splitlines():
+                print_line(b"setup", b"Subprocess python version: " + line + b"\n")
```
ahal, line 1044: I don't see any value in prefixing the uv output here. Was thinking you could do a pre-amble log, maybe something like:
```
print_line(b"setup", b"Detected the following Python versions on the system:")
subprocess.run(["uv", ...])
```

(i.e, don't capture the output and just let it pass through to stdout)

src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,33 @@ def _display_python_version():
...
+
+    except FileNotFoundError:
```
ahal, line 1046: Does `FileNotFoundError` get raised? I thought it would be `subprocess.CalledProcessError`. You might need to change `FileNotFoundError` to that. Beware if you switch to `subprocess.run` above, you can pass `check=True` to make it raise an exception when the command fails.

src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,33 @@ def _display_python_version():
...
+            try:
+                version = subprocess.check_output(
+                    [python_executable, "-c", "import platform; print(platform.python_version())"],
+                ).strip()
+
+                print_line(
+                    b"setup", b"Subprocess %s version: %s\n" % (python_executable.encode(), version)
+                )
+
+            except FileNotFoundError:
```
ahal, line 1060: Same comment as above.

src/taskgraph/run-task/run-task
```diff
@@ -1033,6 +1033,33 @@ def _display_python_version():
...
+            for line in output.splitlines():
+                print_line(b"setup", b"Subprocess python version: " + line + b"\n")
+
+    except FileNotFoundError:
+
+        python_versions = ["python","python3"]
+
+        for python_executable in python_versions:
+            try:
+                version = subprocess.check_output(
```
ahal, line 1052: This isn't wrong, but just fyi, you should try and use `subprocess.run` instead of all the various `call` / `check_call` / `check_output` / `Popen` / whatever methods.

`subprocess.run` is designed to replace all of the above and is the officially recommended method unless you're doing something advanced (in which case using `Popen` is needed).

In this case the equivalend to `check_output` would be:
```
subprocess.run([...], check=True, capture_output=True)
```

test/test_scripts_run_task.py
```diff
@@ -438,6 +438,15 @@ def test_display_python_version_should_output_python_versions(run_task_mod, caps
     assert ("Python version: 3." in output) or ("Python version: 2." in output) is True
 
 
+def test_display_subprocess_python_version_should_output_python_versions_title(
```
ahal, line 441: Thanks for the test!

ahal (changes requested):

PR #736, opened 2025-08-07
feat: use `primary-dependency` attribute to find primary dependency of a task

bhearsum:
Storing this on `from_deps` tasks allows us to avoid iterating over potentially all the task dependencies when finding it later, which can add up to a lot of time in larger graphs.

In a profile I ran for gecko taskgraph, `get_dependencies` (via `get_primary_dependency`) was the second hottest function. With this patch it moves down to 10th place, and takes 1/4 of the amount of time it used to.

### DIFF
--- a/src/taskgraph/transforms/from_deps.py
+++ b/src/taskgraph/transforms/from_deps.py
@@ -206,6 +206,7 @@ def from_deps(config, tasks):
             )
 
             primary_dep = [dep for dep in group if dep.kind == primary_kind][0]
+            new_task["attributes"]["primary-dependency-label"] = primary_dep.label
 
             if set_name:
                 func = SET_NAME_MAP[set_name]
--- a/src/taskgraph/util/dependencies.py
+++ b/src/taskgraph/util/dependencies.py
@@ -68,10 +68,9 @@ def get_dependencies(config: TransformConfig, task: Dict) -> Iterator[Task]:
 
 
 def get_primary_dependency(config: TransformConfig, task: Dict) -> Optional[Task]:
-    """Return the ``Task`` object associated with the primary dependency.
-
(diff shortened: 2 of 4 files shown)


src/taskgraph/transforms/from_deps.py
```diff
@@ -206,6 +206,7 @@ def from_deps(config, tasks):
             )
 
             primary_dep = [dep for dep in group if dep.kind == primary_kind][0]
+            new_task["attributes"]["primary-dependency"] = primary_dep
```
ahal, line 209: Should we anticipate issues from storing an object as an attribute rather than a string? Is there precedent for this?

ahal (approved):
Thanks makes sense!

I think Abhishek might also have a backwards incompatible change coming soon, so it might be worth waiting for his patch and releasing them together.

src/taskgraph/transforms/from_deps.py
bhearsum: There was precedent for it until it was removed in https://phabricator.services.mozilla.com/D193079 :laughing: 

I admit it's a bit weird. Maybe we can just store the primary dependency label, and look it up by that instead, come to think of it?

abhishekmadan30:
I am not working on the duplicate dependency patch currently so it would fine to land this

src/taskgraph/transforms/from_deps.py
bhearsum: Yes, that works, and is much less weird and memory consuming! It arguably makes this a non-breaking change now, too? (I guess there's other ways that `primary-kind-dependency` might be set other than `from_deps` though...)

bhearsum:
Do you want another look at this, @ahal ?

PR #741, opened 2025-08-12
fix: replaced taskcluster rest api calls with package

### LINKED ISSUE #221: Use the Taskcluster Python client rather than direct http requests
Currently we just make manually crafted API requests to the TC endpoint rather than going through the Taskcluster Python client. This has been working fine for a long time, but we are likely missing out on some robustness and could push some complexity out of Taskgraph.

But the bigger reason to consider switching soonish, is that we want to start using the object service in Firefox CI. The official clients all already support artifact verification out of the box. If Taskgraph were using the Python client, it would start getting the benefit of artifact integrity as soon as things start switching over to the object service.

abhishekmadan30:
fixes #221 . Created a bunch of helpers to use this package in taskcluster.py . I also didnt completely depcreate the url creation as that is still used. I bypassed it and replaced all uses when to get request is made.


src/taskgraph/run-task/fetch-content
```diff
@@ -39,6 +39,11 @@ try:
 except ImportError:
     certifi = None
 
+try:
+    import taskcluster
+except ImportError:
+    taskcluster = None
```
ahal, line 45: Nice idea, but in this case I think the benefits of using the TC client probably don't outweigh the complexity of supporting two ways of interacting with the API. So I think we should leave this file alone

src/taskgraph/util/taskcluster.py
```diff
@@ -75,6 +76,66 @@ def get_root_url(use_proxy):
...
+
+
+@functools.lru_cache(maxsize=None)
+def get_taskcluster_secret():
+    if "TASKCLUSTER_PROXY_URL" in os.environ:
+        secret_options = {"rootUrl": os.environ["TASKCLUSTER_PROXY_URL"]}
+    else:
+        secret_options = taskcluster.optionsFromEnvironment()
+
+    return taskcluster.Secrets(secret_options)
```
ahal, line 136: Might be nicer to do something like:
https://github.com/mozilla-releng/fxci-config/blob/50f95de0057cb77b87d5f0cee12560c335338a40/taskcluster/fxci_config_taskgraph/util/integration.py#L38

At the very least factoring out the options into some kind of `_get_options` helper would help cut down some duplication.

src/taskgraph/util/taskcluster.py
```diff
@@ -164,13 +225,17 @@ def get_artifact(task_id, path, use_proxy=False):
...
 
 
 def list_artifacts(task_id, use_proxy=False):
-    response = _do_request(get_artifact_url(task_id, "", use_proxy).rstrip("/"))
-    return response.json()["artifacts"]
+    queue = get_taskcluster_queue()
+    task = queue.task(task_id)
+    if task:
+        return task["artifacts"]
+    raise KeyError(f"task {task_id} not found")
```
ahal, line 238: We probably don't want `KeyError` here, that's usually reserved for attempting to access a non-existent key in a dictionary.

That said, I think this might be dead code anyway? Is it possible for `queue.task()` to return `None` without raising an HTTPError already?

ahal (changes requested):
This is shaping up nicely, thanks!

I have some changes to request. Once you put up a new revision I'll pull this down locally and see if it causes any differences in the graph (both here and in Gecko).

src/taskgraph/util/taskcluster.py
abhishekmadan30: It would raise a resourcenotfound error so I can remove the key error but I think it would be better to keep the none check because removes pyright issues

ahal:
Also I think I changed my mind about leaving `use_proxy` in the function signature just for backwards compatibility.. We should remove it rather than confuse consumers who are attempting to set it to `False` or whatnot and then wondering why it isn't working.

We can do a major version bump for this.. no big deal.

src/taskgraph/util/taskcluster.py
```diff
@@ -494,16 +453,14 @@ def get_ancestors(
...
-        except requests.HTTPError as e:
+        task_def = get_task_definition(task_id)
+        if not task_def:
             # Task has most likely expired, which means it's no longer a
             # dependency for the purposes of this function.
-            if e.response.status_code == 404:
-                continue
-
-            raise e
+            continue
```
ahal, line 460: Are you sure the taskcluster client would return `None` on a 404? I would have thought it would raise an exception still.. Though it would probably be from aiohttp rather than requests.

test/test_optimize_strategies.py
```diff
@@ -57,29 +58,31 @@ def params():
         ),
     ),
 )
-def test_index_search(caplog, responses, params, state, expires, expected, logs):
```
ahal, line 60: This is fine, but fyi there exists an aioresponses mock library designed to be used with aiohttp that we could make use of in a similar way.

ahal (changes requested):
Thank you!

This is mostly cleanup, but there are two issues I neglected to catch in my first review:
1. The `_handle_artifact` for yaml
2. The `.capitalize()` issue

src/taskgraph/util/taskcluster.py
```diff
@@ -115,47 +125,14 @@ def get_session():
...
-def _handle_artifact(path, response):
-    if path.endswith(".json"):
-        return response.json()
-    if path.endswith(".yml"):
-        return yaml.load_stream(response.content)
-    response.raw.read = functools.partial(response.raw.read, decode_content=True)
-    return response.raw
-
-
 def get_artifact_url(task_id, path, use_proxy=False):
```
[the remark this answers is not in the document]
abhishekmadan30: I tried doing this but buildUrl causes the url to be encoded so I think it be better to make a separate bug as we need to not only modify this plus all place that input buildUrl

PR #745, opened 2025-08-14
fix: use importlib.import_module instead of `__import__`

bhearsum:
`import_module` is newer, and provides a couple of advantages over a bare `__import__`:
> The import_module() function acts as a simplifying wrapper around importlib.__import__(). This means all semantics of the function are derived from importlib.__import__(). The most important difference between these two functions is that import_module() returns the specified package or module (e.g. pkg.mod), while __import__() returns the top-level package or module (e.g. pkg).

It's unclear if this will actually fix any problems in the real world, but there's no recent to use `__import__` at this point AFAICT.

### DIFF
--- a/src/taskgraph/config.py
+++ b/src/taskgraph/config.py
@@ -148,7 +148,9 @@ class GraphConfig:
         sys.path.insert(0, self.root_dir)
         register_path = self["taskgraph"].get("register")
         if register_path:
-            find_object(register_path)(self)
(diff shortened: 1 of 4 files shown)


bhearsum:
The first version of this PR added a `Callable` return type to `find_object`. This is how it's used in practice, but broke tests, which highlights that it would be technically a breaking change to do so. I've removed that part.

PR #749, opened 2025-08-20
Add a function to get tasks from multiple kinds in one pass

glandium:
This is especially useful when some of the kinds the caller is interested in is a dependency of another kind.


jcristau:
There seem to be other places where we assume the name matches the label e.g. https://github.com/taskcluster/taskgraph/blob/63172076144dbaf8589ea548d1dc6c517b0aaaf6/src/taskgraph/actions/retrigger.py#L150; so changing the test fixture to set name to the same thing as label seems like a reasonable change, and would match https://github.com/taskcluster/taskgraph/blob/63172076144dbaf8589ea548d1dc6c517b0aaaf6/src/taskgraph/transforms/task.py#L1331

glandium:
Can you merge this?

Issue #476, opened 2024-03-21
Fix remaining type checking errors


ahal:
https://github.com/taskcluster/taskgraph/pull/475 refactored our Pyright type checking excludes to be per line rather than ignoring entire files. Fixing these remaining errors is a good first issue!

First ensure you can run `pyright` locally:
```
$ pip install -r requirements/dev.txt
$ pyright
```
You should see no errors. Next:

1. Search for the string `# type: ignore` in the code base.
2. Remove it.
3. Run `pyright` and verify you see the error.
4. Figure out how to fix the error.
5. Submit a PR

Feel free to batch many fixes into one PR if you like. And no need to fix everything! This issue can be tackled a little bit at a time, possibly even by multiple contributors.

ahal:
The change to switch to ignoring type errors line by line finally landed:
https://github.com/taskcluster/taskgraph/commit/ae78c110445c749f189489ea9dd88ec04ae94930

So the instructions in this issue make sense now.

PR #751, opened 2025-08-24
Fix some more type checking errors

YaySushi:
This fixes (another very small) part of issue https://github.com/taskcluster/taskgraph/issues/476.

Notes:
pytest reported the same stats before and after my changes (looks like orjson gives a warning)
Some removals of # type: ignore did not result in an error from pyright (some status code comparisons)


ahal (approved):
Thanks very much for the contribution!

PR #752, opened 2025-08-26
Convert voluptuous schema to msgspec 

abhishekmadan30:


src/taskgraph/config.py
```diff
@@ -8,107 +8,89 @@
...
-graph_config_schema = Schema(
-    {
-        # The trust-domain for this graph.
-        # (See https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/taskgraph.html#taskgraph-trust-domain)  # noqa
-        Required("trust-domain"): str,
-        Optional(
-            "docker-image-kind",
-            description="Name of the docker image kind (default: docker-image)",
-        ): str,
-        Required("task-priority"): optionally_keyed_by(
```
ahal, line 34: I think we'll need to preserve the `optionally_keyed_by` function.

I can help you with this later, but at a high level, I think we should be able to dynamically return the proper Python type annotation for it.

ahal: To illustrate what `optionally_keyed_by` is doing, it means in this case the following inputs would all be valid:

```
task-priority:
    by-project:
        autoland: high
        default: low
```
```
task-priority:
    by-level:
        3: high
        1: low
```
```  
task-priority: high
```

So we need to come up with a standard Python type annotation that captures this.. It might be a bit ugly, but I think it should be do-able.

docs/concepts/transforms.rst
```diff
@@ -105,17 +105,17 @@ about the state of the tasks at given points. Here is an example:
...
-       Required("foo"): str,
-       Optional("bar"): bool,
-   })
+   class MySchema(msgspec.Struct, kw_only=True):
+       foo: str                   # Required field
+       bar: Optional[bool] = None # Optional field
 
-   transforms.add_validate(my_schema)
+   transforms = TransformSequence()
+   transforms.add_validate(Schema(MySchema))
```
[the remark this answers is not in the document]
abhishekmadan30: Yes I did

src/taskgraph/transforms/matrix.py
```diff
@@ -8,59 +8,49 @@
...
+    # Useful for cases where the default of joining matrix values by
+    # a dash is not desired.
+    set_name: Optional[str] = None
+    # List of fields in the task definition to substitute matrix values into.
+    #
+    # If not specified, all fields in the task definition will be
+    # substituted.
+    substitution_fields: Optional[List[str]] = None
+    # Allow extra fields for matrix dimensions
+    __extras__: Dict[str, List[str]] = msgspec.field(default_factory=dict)
```
ahal, line 43: Is this necessary? According to the msgspec docs, unknown fields are ignored by default:
https://jcristharif.com/msgspec/structs.html#forbidding-unknown-fields

Though, does that mean we need to somehow pass `forbid_unknown_fields=True` for every schema where we *don't* want this? This might be a bit annoying :/

What do you think about trying to invert the behaviour? E.g, in the `Schema` subclass, we set `forbid_unknown_fields` to `True` by default, and then if the user does something like:
```
class Foo(Schema, allow_extra=True):
    ...
```

Then in that case we pass `forbid_unknown_fields=False`.

I'm not going to lie, subclassing the `msgspec.Struct` is tricky due to it being a dataclass, so I'm not entirely sure how to implement this off-hand, or even if it's possible. Hopefully Claude can assist

ahal: Oh, I put my comment in the wrong place... In this case we do want to explicitly list `__extras__` because any extra keys should absolutely follow that schema.. so my bad.

But for example, lower down on line 49, it's probably not necessary because the "extras" could literally be anything, the matrix schema has no idea.

docs/concepts/transforms.rst
```diff
@@ -105,17 +105,16 @@ about the state of the tasks at given points. Here is an example:
...
    from taskgraph.transforms.base import TransformSequence
    from taskgraph.util.schema import Schema
 
-   my_schema = Schema({
-       Required("foo"): str,
-       Optional("bar"): bool,
-   })
+   class MySchema(Schema):
+       foo: str                   # Required field
+       bar: Optional[bool] = None # Optional field
```
ahal, line 114: nit: Python standard is to have two spaces before comment

src/taskgraph/config.py
```diff
@@ -177,7 +161,9 @@ def kinds_dir(self):
 
 
 def validate_graph_config(config):
-    validate_schema(graph_config_schema, config, "Invalid graph configuration:")
+    """Validate graph configuration using msgspec."""
+    # With rename="kebab", msgspec handles the conversion automatically
```
ahal, line 161: nit: This comment doesn't make sense in this context.

src/taskgraph/parameters.py
```diff
@@ -143,15 +162,22 @@ def extend_parameters_schema(schema, defaults_fn=None):
...
             dict mapping parameter name to default value in the
             event strict=False (optional).
     """
     global base_schema
     global defaults_functions
-    base_schema = base_schema.extend(schema)
+    global _schema_extensions
+
+    # Store the extension schema for use during validation
+    _schema_extensions.append(schema)
```
ahal, line 172: Is this unused? I don't see where this is being validated

src/taskgraph/transforms/fetch.py
```diff
@@ -9,91 +9,76 @@
...
+FETCH_SCHEMA = FetchSchema
 
 # define a collection of payload builders, depending on the worker implementation
 fetch_builders = {}
 
 
 @dataclass(frozen=True)
 class FetchBuilder:
-    schema: Schema
+    schema: Any  # Either msgspec.Struct type or validation function
```
ahal, line 73: Should this be `Union[Schema, Callable]` then?

src/taskgraph/transforms/fetch.py
```diff
@@ -9,91 +9,76 @@
...
-    }
-)
+class FetchConfig(Schema, rename=None, omit_defaults=False):
+    """Configuration for a fetch task type."""
+
+    type: str
+    # Additional fields handled dynamically by fetch builders
+
+
+class FetchSchema(Schema):
```
ahal, line 35: I've noticed you sometimes use `FooSchema`, sometimes `FooConfig` and sometime neither. It would be good to be consistent across all schema class names.

Personally I think I like `FooSchema` the best (even for things that are sub schemas). Though happy to bikeshed with you if you have another preference.

src/taskgraph/transforms/fetch.py
```diff
@@ -9,91 +9,76 @@
...
+    attributes: Optional[Dict[str, Any]] = None
+
+    def __post_init__(self):
+        # Validate that fetch has a 'type' field
+        if not isinstance(self.fetch, dict) or "type" not in self.fetch:
+            raise msgspec.ValidationError("fetch must be a dict with a 'type' field")
+
+
+# Backward compatibility
+FETCH_SCHEMA = FetchSchema
```
ahal: We're already breaking backwards compatibility, so I don't think we should worry about preserving the old variable names. I'll be easy enough to update.

src/taskgraph/transforms/fetch.py
```diff
@@ -204,45 +197,46 @@ def make_task(config, tasks):
...
+    """GPG signature verification configuration."""
+
+    # URL where GPG signature document can be obtained. Can contain the
+    # value ``{url}``, which will be substituted with the value from ``url``.
+    sig_url: str
+    # Path to file containing GPG public key(s) used to validate download.
+    key_path: str
+
+
+class StaticUrlFetchConfig(Schema, rename="kebab"):
```
ahal, line 207: nit: the `rename="kebab"` isn't necessary (right?)

src/taskgraph/transforms/from_deps.py
```diff
@@ -13,108 +13,93 @@
...
+        # Validate group_by
+        if self.group_by is not None:
+            if isinstance(self.group_by, str) and self.group_by not in GROUP_BY_MAP:
+                raise msgspec.ValidationError(f"Invalid group-by: {self.group_by}")
+            elif isinstance(self.group_by, dict):
+                keys = list(self.group_by.keys())
+                if len(keys) != 1 or keys[0] not in GROUP_BY_MAP:
+                    raise msgspec.ValidationError(
+                        f"Invalid group-by dict: {self.group_by}"
+                    )
```
ahal, line 88: I think you can get rid of this `__post_init__` function by creating a new type alias like you did for `priority` in `config.py`.

src/taskgraph/transforms/from_deps.py
```diff
@@ -13,108 +13,93 @@
...
+
+
 #: Schema for from_deps transforms
```
ahal, line 91: These `#:` comments aren't necessary for classes, I only needed them for autodoc to process module level variables. For classes, as long as it has a docstring, you can remove these comments.

src/taskgraph/transforms/from_deps.py
```diff
@@ -13,108 +13,93 @@
...
-            ): {str: [fetches_schema]},
-        },
-    },
-    extra=ALLOW_EXTRA,
-)
+    from_deps: FromDepsConfig
+
+
+# Backward compatibility
+FROM_DEPS_SCHEMA = FromDepsSchema
```
ahal: nit: ditto, can remove this imo

src/taskgraph/transforms/notify.py
```diff
@@ -54,46 +70,114 @@
...
+
+    notify: Optional[NotifyConfig] = None
+    notifications: Optional[LegacyNotificationsConfig] = None
+
+    def __post_init__(self):
+        # Ensure only one of notify or notifications is present
+        if self.notify and self.notifications:
+            raise msgspec.ValidationError(
+                "Cannot specify both 'notify' and 'notifications'"
+            )
```
ahal, line 146: Oh nice, this is a good way of handling mutual exclusion!

src/taskgraph/transforms/run/__init__.py
```diff
@@ -11,13 +11,13 @@
 
 import copy
 import logging
-from textwrap import dedent
+from typing import Any, Dict, List, Union
+from typing import Optional as TOptional
```
ahal, line 15: nit: can you revert `TOptional` back to `Optional`?

src/taskgraph/util/dependencies.py
```diff
@@ -35,7 +34,7 @@ def group_by_all(config, tasks):
     return [[task for task in tasks]]
 
 
-@group_by("attribute", schema=Schema(str))
+@group_by("attribute")
```
ahal, line 37: I guess there's no easy way to validate a schema with no keys with msgspec? Seems fair enough to just drop the schema here then

src/taskgraph/util/schema.py
```diff
@@ -17,52 +16,39 @@ def validate_schema(schema, obj, msg_prefix):
...
-        raise Exception("\n".join(msg) + "\n" + pprint.pformat(obj))
 
-
-def optionally_keyed_by(*arguments):
+    try:
+        if isinstance(schema, type) and issubclass(schema, Schema):
+            # Use the validate class method for Schema subclasses
+            schema.validate(obj)
+        elif isinstance(schema, type) and issubclass(schema, msgspec.Struct):
+            msgspec.convert(obj, schema)
```
ahal, line 33: Do we need to support this use case? I'm not opposed, but if we never call `validate_schema` with a Struct object, then I'd mildly prefer we remove this code path. If we do call it, then this is totally fine.

src/taskgraph/util/schema.py
```diff
@@ -150,99 +136,130 @@ def resolve_keyed_by(
...
+    """Reference to a task artifact."""
+
+    artifact_reference: str
+
+
+# Create a custom validator
+class OptimizationValidator:
+    """A validator that can validate optimization schemas."""
+
+    def __call__(self, value):
```
ahal, line 208: nit: Might as well just make these functions rather than classes with a `__call__`.

template/{{cookiecutter.project_name}}/taskcluster/{{cookiecutter.project_slug}}_taskgraph/transforms/hello.py
```diff
@@ -1,14 +1,12 @@
...
-        Required("noun"): str,
-    },
-    extra=ALLOW_EXTRA,
-)
+
+class HelloSchema(Schema):
+    noun: str  # Required field
+
+
+HELLO_SCHEMA = HelloSchema
```
ahal, line 9: nit: Let's drop this intermediate variable.

ahal (changes requested):
Thanks, this is shaping up nicely!

I don't think I have any other major concerns, it's mostly all nits and a minor changes. So I think you can go ahead and start converting Gecko without fear of a major new request (from me at least :p)

docs/tutorials/creating-a-task-graph.rst
```diff
@@ -136,16 +136,16 @@ comments for explanations):
...
 .. code-block:: python
 
-   from voluptuous import Optional, Required
-
-   from taskgraph.transforms.base import TransformSequence
+   from typing import Optional
    from taskgraph.util.schema import Schema
+   from taskgraph.transforms.base import TransformSequence
+
+   # Define the schema using Schema base class for better type checking and performance.
```
ahal, line 143: nit: I'd cut out `for better type checking and performance` here. The user reading this won't know what you're comparing against

src/taskgraph/config.py
```diff
@@ -21,94 +17,78 @@
...
+
+    This schema allows extra fields for repository-specific configuration.
+    """
+
+    # Required fields first
+    name: str
+
+    # Optional fields
+    project_regex: Optional[str] = None  # Maps from "project-regex"
+    ssh_secret_name: Optional[str] = None  # Maps from "ssh-secret-name"
```
ahal, line 52: I guess this maps from works because in the base class we pass in the "rename" arg. Out of curiosity, what does it look like if we *wanted* to use underscores rather than dashes?

src/taskgraph/config.py
```diff
@@ -21,94 +17,78 @@
...
+
+
+class TaskGraphSchema(Schema):
+    """Taskgraph specific configuration."""
+
+    # Required fields first
+    repositories: Dict[str, Repository]
+
+    # Optional fields
+    register: Optional[str] = None
```
ahal, line 68: I feel bad for not commenting this earlier, but do you think we could preserve the old descriptions as comments? The schema isn't really documented other than these descriptions, so I feel it's important that we preserve them somehow. Even if it's just a comment

ahal: I think Claude should be able to do this if you get it to examine the diff and teach it about the mapping from dash to underscore.

src/taskgraph/config.py
```diff
@@ -21,94 +17,78 @@
...
+
+    # Optional fields
+    project_regex: Optional[str] = None  # Maps from "project-regex"
+    ssh_secret_name: Optional[str] = None  # Maps from "ssh-secret-name"
+
+
+class RunConfig(Schema):
+    """Run transforms configuration."""
+
+    use_caches: Optional[Union[bool, List[str]]] = None  # Maps from "use-caches"
```
ahal, line 58: This shouldn't just be `List[str]`, there are only specific values allowed here (similar to `TaskPriority`)

ahal (changes requested):
I didn't get a chance to review the whole thing, but I'll leave the comments I had so far for now and resume tomorrow. Figured you might want to get a head start

src/taskgraph/config.py
abhishekmadan30: I couldn't do Literal because it doesn't support dynamics values so I added a post_init function to make sure the values are valid

src/taskgraph/transforms/task.py
```diff
@@ -49,343 +45,96 @@ def _run_task_suffix():
...
+    # Description of the task (for metadata)
+    description: str
+    # The provisioner-id/worker-type for the task
+    worker_type: str
+    # Attributes for this task
+    attributes: Dict[str, Any] = msgspec.field(default_factory=dict)
+    # Relative path (from config.path) to the file task was defined in
+    task_from: Optional[str] = None
+    # Dependencies of this task, keyed by name
+    dependencies: Dict[str, Any] = msgspec.field(default_factory=dict)
```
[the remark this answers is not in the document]
abhishekmadan30: Sorry, I am confused what you mean this ?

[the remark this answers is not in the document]
ahal: In the old schema, if you named a dependency edge "self" or "decision", it would raise an exception. That won't happen anymore.

ahal:
Thanks, looks like you have some merge conflicts too. Please rebase and push your latest iteration and hopefully I can do one final review (of just the interdiff this time)

src/taskgraph/transforms/run/__init__.py
```diff
@@ -27,146 +27,98 @@
...
-                specified. If 'label' is not provided, it will be generated from
-                the 'name' by prepending the kind.
-                """
-            ),
-        ): str,
-        # the following fields are passed directly through to the task description,
-        # possibly modified by the run implementation. See
-        # taskcluster/taskgraph/transforms/task.py for the schema details.
-        Required("description"): task_description_schema["description"],
-        Optional("priority"): task_description_schema["priority"],
```
ahal, line 67: All these keys that previously referenced the `task_description_schema` should continue referencing it moving forward. I think this is where I was saying we'd need to use that `__annotations__` special member variable. Feel free to reach out.

So:
```
Required("description"): task_description_schema["description"]
```

Would become:
```
description: TaskSchema.__annotations__["description"]
```

I think.. some testing / experimentation needed.

src/taskgraph/transforms/docker_image.py
```diff
@@ -5,18 +5,15 @@
...
+from typing import Dict, List, Optional
 
 import taskgraph
 from taskgraph.transforms.base import TransformSequence
+from taskgraph.transforms.task import TaskDescriptionIndexSchema
 from taskgraph.util import json
 from taskgraph.util.docker import create_context_tar, generate_context_hash
 from taskgraph.util.schema import Schema
 
-from .task import task_description_schema
```
ahal, line 18: Same here, anytime we see a schema being referenced in another file, we should continue referencing the new msgstruct style schema from the same file.

src/taskgraph/transforms/fetch.py
```diff
@@ -9,91 +9,72 @@
...
+
+# Base class for fetch configurations - ensures type field exists
+class BaseFetchSchema(Schema, forbid_unknown_fields=False):
+    """Base schema for fetch configurations.
+
+    This allows any additional fields beyond 'type' to support
+    different fetch types (static-url, git, etc).
+    """
+
+    type: str
```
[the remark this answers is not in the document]
abhishekmadan30: in the voluptuous schema, it was also a string not a literal

jcristau:
Is there any chance we can do this in a backwards-compatible way, e.g. by leaving existing util/schema.py APIs alone and adding the msgspec schemas as an alternative, so we can convert things piecemeal and avoid massive changes and likely regressions/back and forth?  Or is that impossible for some reason?

src/taskgraph/util/schema.py
```diff
@@ -2,67 +2,109 @@
...
+
+def UnionTypes(*types):
+    """Use `functools.reduce` to simulate `Union[*allowed_types]` on older
+    Python versions.
+    """
+    return reduce(lambda a, b: Union[a, b], types)
 
 
-def optionally_keyed_by(*arguments):
+def optionally_keyed_by(*arguments, use_msgspec=False):
```
ahal: Oh, I see what you meant now by having separate functions for this.. For some reason I was thinking there'd be an object passed in here we could test, but that obviously isn't the case. I like this approach though.

src/taskgraph/transforms/run/__init__.py
```diff
@@ -11,162 +11,134 @@
...
-                This object will be passed through to the task description, with additions
-                provided by the task's run-using function.
-                """
-            ),
-        ): dict,
-    }
-)
+
+# Fetches schema using msgspec
+class FetchesStruct(Struct):
```
ahal: Would you mind splitting all the actual schema conversion work into a separate commit from supporting msgspec in validate_schema? This way it will be easy to prove that the generation works before and after the change.

PR #765, opened 2025-09-05
feat: generate kinds in parallel across multiple processes

bhearsum:
This is a slightly updated version of #738 + #744, which were backed out a few weeks ago due to not working correctly on Windows. In a conversation I had today it was suggested that we should just not support multiprocess on Windows so we can at least take the win elsewhere. This patch does that, and surely no other issues will arise...

### DIFF
--- a/src/taskgraph/generator.py
+++ b/src/taskgraph/generator.py
@@ -5,6 +5,12 @@
 import copy
 import logging
 import os
+import platform
+from concurrent.futures import (
+    FIRST_COMPLETED,
+    ProcessPoolExecutor,
+    wait,
+)
 from dataclasses import dataclass
 from typing import Callable, Dict, Optional, Union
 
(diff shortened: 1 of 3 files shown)


bhearsum:
This still needs more testing in Gecko; I'm opening it as a draft for now to sanity check CI on the taskgraph side.

src/taskgraph/generator.py
```diff
@@ -253,6 +264,101 @@ def _load_kinds(self, graph_config, target_kinds=None):
...
+
+        return all_tasks
+
+    def _load_tasks_parallel(self, kinds, kind_graph, parameters):
+        all_tasks = {}
+        futures_to_kind = {}
+        futures = set()
+        edges = set(kind_graph.edges)
+
+        with ProcessPoolExecutor() as executor:
```
jcristau, line 305: AIUI we rely on the use of `fork` here, so per the documentation note at https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ProcessPoolExecutor should pass
`mp_context=multiprocessing.get_context("fork")` to ensure we get the right thing (and instead of platform == windows we can test for `"fork" in multiprocessing.get_all_start_methods()`?).

I might be confused though because apparently `spawn` is the default on macos already.  Did we test this there?

bhearsum: Yup; good callout. Even with forking though, it looks like this crashes on mac repeatedly. https://github.com/python/cpython/issues/77906 suggests that this is a common problem when forking on macOS :(. (`forkserver` also does not work; it fails the same way that `spawn` does.)

bhearsum: I ended up turning off support for everything but Linux, to be safe.

jcristau, line 305: still need to explicitly request the `fork` thing here

src/taskgraph/generator.py
```diff
@@ -307,31 +413,18 @@ def _run(self):
...
-                    message += f' - "{k}"\n'
-                raise Exception(message)
+        # Current parallel generation relies on multiprocessing, and forking.
+        # This causes problems on Windows and macOS due to how new processes
+        # are created there, and how doing so reinitializes global variables
+        # that are modified earlier in graph generation, that doesn't get
+        # redone in the new processes. Ideally this would be fixed, or we
+        # would take another approach to parallel kind generation. In the
+        # meantime, it's not supported outside of Linux.
+        if platform.system() != "Linux":
```
jcristau, line 423: Consider something like
```suggestion
        if platform.system() != "Linux" or os.environ.get("TASKGRAPH_SERIAL"):
```
so that the serial case can be tested on linux still.

jcristau (approved):
r+wc

src/taskgraph/generator.py
bhearsum: erm, sorry...that's a major oversight

PR #768, opened 2025-09-08
Add ability to use custom images in `taskgraph load-task`

ahal:
This makes it really nice to iterate on a docker image and test it against a live task, by using e.g:
```
taskgraph load-task <task-id> --image python
```

This will re-build the `taskcluster/docker/python` image, launch it in a container, and load `<task-id>` into it.

### DIFF
--- a/src/taskgraph/docker.py
+++ b/src/taskgraph/docker.py
@@ -10,13 +10,14 @@ import tarfile
 import tempfile
 from io import BytesIO
 from textwrap import dedent
-from typing import List, Optional
+from typing import Dict, Generator, List, Mapping, Optional, Union
 
 try:
     import zstandard as zstd
 except ImportError as e:
     zstd = e
 
+from taskgraph.config import GraphConfig
 from taskgraph.util import docker, json
 from taskgraph.util.taskcluster import (
     find_task_id,
@@ -43,7 +44,15 @@ The VERSION file contains the version of the image.
 """
 
 
-def get_image_digest(image_name):
+def get_image_digest(image_name: str) -> str:
+    """Get the digest of a docker image by its name.
+
+    Args:
+        image_name: The name of the docker image to get the digest for.
+
+    Returns:
+        str: The digest string of the cached docker image task.
+    """
     from taskgraph.generator import load_tasks_for_kind  # noqa: PLC0415
     from taskgraph.parameters import Parameters  # noqa: PLC0415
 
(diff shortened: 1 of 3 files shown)


src/taskgraph/main.py
```diff
@@ -683,11 +683,34 @@ def image_digest(args):
     help="Keep the docker container after exiting.",
 )
 @argument("--user", default=None, help="Container user to start shell with.")
+@argument(
+    "--image",
+    default=None,
+    help="Use a custom image instead of the task's image. Can be the name of "
+    "an image under `taskcluster/docker`, a reference to a local docker image, "
```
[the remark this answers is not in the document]
ahal: Oops, I implemented that in a future commit that I haven't yet put up for review. I think it got absorbed here.

> Also, for the "pass the name of an image" case, it seems like ideally this should take the name of the task from the docker-image kind, not the name of the directory (with the latter you'll miss things like parent - although I guess that doesn't work with build_image anyway - and args)?

Heh, I'm spoiling the surprise for my demo, but I got `taskgraph build-image` to use `image_builder`. It'll now recursively build parent images (if their digest changed). I have a long stack of WIP commits I need to untangle and get shaped up for review though, so wanted to just get this landed first.

src/taskgraph/docker.py
```diff
@@ -387,6 +430,13 @@ def load_task(task_id: str, remove: bool = True, user: Optional[str] = None) ->
         print("Only tasks using `run-task` are supported!")
         return 1
 
+    try:
+        image = custom_image or image
+        image_tag = _resolve_image(image, graph_config)
+    except Exception as e:
+        print(e)
```
jcristau, line 437: nit: `file=sys.stderr`

src/taskgraph/docker.py
```diff
@@ -354,7 +355,47 @@ def _index(l: List, s: str) -> Optional[int]:
         pass
 
 
-def load_task(task_id: str, remove: bool = True, user: Optional[str] = None) -> int:
+def _resolve_image(image: Union[str, Dict[str, str]], graph_config: GraphConfig) -> str:
+    image_task_id = None
+
+    # Standard case, image comes from the task definition.
+    if isinstance(image, dict):
```
jcristau, line 362: Should we attempt to support the case where `payload.image` is just a string, like decision tasks?  It looks like currently that fails with `TypeError: string indices must be integers, not 'str'` because we assume it's a dict, and with this change it'd clash with the "refer to an in-tree image" case.  Maybe move that to `in-tree=foo` to leave the option open?

ahal: That's https://github.com/taskcluster/taskgraph/pull/772/commits/39008139879e6a9fc298640702bb3e05be355cf4

But I'll just fold that into here as it's so trivial (just ended up separate due to how I happened to write my commits).

In this implementation the behaviour will be to first check for a local image, and then use a registry if there's no match, which I think I prefer to `in-tree=`, but don't feel super strongly about.

src/taskgraph/main.py
```diff
@@ -683,11 +683,34 @@ def image_digest(args):
     help="Keep the docker container after exiting.",
 )
 @argument("--user", default=None, help="Container user to start shell with.")
+@argument(
+    "--image",
+    default=None,
+    help="Use a custom image instead of the task's image. Can be the name of "
+    "an image under `taskcluster/docker`, `task-id=<task id>`, or "
+    "`index=<index path>`.",
```
jcristau, line 691: or the name of an image to pull from a registry

jcristau (approved):

PR #772, opened 2025-09-11
Support shallow clones with Git

ahal:


src/taskgraph/run-task/run-task
```diff
@@ -611,8 +611,8 @@ def git_checkout(
     base_repo: Optional[str],
     base_ref: Optional[str],
     base_rev: Optional[str],
-    ref: Optional[str],
-    commit: Optional[str],
+    head_ref: Optional[str],
+    head_rev: Optional[str],
```
jcristau: :100: thanks!

src/taskgraph/run-task/run-task
```diff
@@ -669,29 +685,18 @@ def git_checkout(
...
-        args.extend(["--tags", "--force"])
-
-    else:
-        args.append("--no-tags")
+    # If a head_ref was provided, it might be tag, so we need to make sure we fetch
+    # those. This is explicitly only done when base and head repo match,
+    # because it is the only scenario where tags could be present. (PRs, for
+    # example, always include an explicit rev.) Failure to do this could result
+    # in not having a tag, or worse: having an outdated version of one.
+    tags = True if head_ref and base_repo == head_repo else False
```
jcristau: nit: `tags = head_ref and base_repo == head_repo` (although this gets modified in a following commit, so doesn't matter)

src/taskgraph/run-task/run-task
```diff
@@ -698,16 +724,29 @@ def git_checkout(
...
 
     args = [
         "git",
         "checkout",
         "-f",
     ]
 
-    if head_ref:
-        args.extend(["-B", head_ref])
+    if head_ref and head_ref != head_rev:
```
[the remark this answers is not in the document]
ahal: I believe I ran into a case where `head_ref` was being set to `head_rev` in the .tc.yml, though now I'm having trouble finding it, so I'm wondering if I'm just being overly cautious.

Maybe instead we should error out of `run-task` if `head_ref` follows the pattern of a commit sha to help cut down on edge cases.

PR #779, opened 2025-09-18
feat: pass along `write_artifacts` to loaders

bhearsum:
Taskgraph already passes this information along to transforms, and over in https://bugzilla.mozilla.org/show_bug.cgi?id=1989038 we have a use case to have loaders know about it too: we need to write out an artifact whose contents are only known after all tasks for a specific kind have been loaded and transformed.

(Technically we could avoid passing this along, but it would mean we have no reasonable way to switch off the writing of this artifact outside of decision tasks...)

### DIFF
--- a/src/taskgraph/generator.py
+++ b/src/taskgraph/generator.py
@@ -66,6 +66,7 @@ class Kind:
             config,
             parameters,
             list(kind_dependencies_tasks.values()),
+            write_artifacts,
         )
 
         transforms = TransformSequence()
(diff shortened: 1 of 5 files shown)


src/taskgraph/generator.py
```diff
@@ -66,6 +66,7 @@ def load_tasks(self, parameters, kind_dependencies_tasks, write_artifacts):
             config,
             parameters,
             list(kind_dependencies_tasks.values()),
+            write_artifacts,
```
jcristau, line 69: I guess it's not worth trying to avoid breaking backwards compat here?

bhearsum: If there's a way to avoid the compat break without jumping through major hoops I'd love it...I couldn't find a way myself.

bhearsum: Clever! Yes, let's do that!

PR #782, opened 2025-09-19
fix(from_deps): use appropriate dependency name when unique-kinds is …

ahal:
…true


ahal:
I'm somewhat surprised we never ran into this bug before now, given that `unique-kinds` defaults to True.

bhearsum (approved):
This should be fine; it's a breaking change though AFAICT?

PR #783, opened 2025-09-22
fix(deps): update to image_builder 6.0.0

ahal:


src/taskgraph/transforms/docker_image.py
```diff
@@ -198,7 +198,7 @@ def fill_template(config, tasks):
                 "artifacts": [
                     {
                         "type": "file",
-                        "path": "/workspace/image.tar.zst",
+                        "path": "/workspace/out/image.tar.zst",
```
hneiva, line 201: is this dir change because of a change in image_builder from 5 to 6?

hneiva (approved):
LGTM

PR #785, opened 2025-09-24
Use image_builder in `build-image` and `load-task` commands

ahal:


src/taskgraph/main.py
```diff
@@ -582,19 +579,16 @@ def show_taskgraph(options):
...
-    if args["context_only"] is None:
-        build_image(args["image_name"], args["tag"], graph_config, os.environ)
-    else:
-        build_context(
-            args["image_name"], args["context_only"], graph_config, os.environ
-        )
+    return build_image(
+        graph_config,
+        args["image_name"],
+        args["context_only"],
```
jcristau, line 590: The new save_image isn't really the same thing as the old context_only?

src/taskgraph/docker.py
```diff
@@ -183,19 +161,82 @@ def build_image(
...
+            if e.response.status_code != 404:
+                raise
+
+            # Parent id doesn't exist, needs to be re-built as well.
+            parent = task.dependencies["parent"][len("docker-image-") :]
+            parent_tar = temp_dir / "parent.tar"
+            build_image(graph_config, parent, save_image=str(parent_tar))
+            volumes[str(parent_tar)] = "/workspace/parent.tar"
+
+    task_def["payload"]["env"]["CHOWN_OUTPUT"] = "1000:1000"
```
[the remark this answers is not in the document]
ahal: What happens if the uid/gid doesn't exist in image_builder though? I think only 1000:1000 is there.. That said, I guess if the user has something different they are out of luck anyway so it doesn't hurt to set it to the current uid:gid?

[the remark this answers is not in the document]
jcristau: They're just numbers, chown doesn't care?  (I'm not actually sure I understand what breaks without this though, so I might well be misunderstanding the issue)

src/taskgraph/main.py
ahal: Unless you meant it as an actual question, in which case, no context_only just saves the docker-context as generated by the decision task, it doesn't actually build the image.

I kind of question the usefulness of this, but it was easy enough to preserve *shrug*

jcristau: My mistake, I misread the arguments to taskgraph.docker.build_image

jcristau (approved):

PR #802, opened 2025-10-06
generator: add support for "graph_config" verifications

jcristau:


src/taskgraph/generator.py
```diff
@@ -378,6 +378,7 @@ def _run(self):
 
         # Initial verifications that don't depend on any generation state.
         self.verify("initial")
+        self.verify("graph_config", graph_config)
```
ahal: Should we just get rid of initial verifications or pass `graph_config` into "initial" and not have a separate "graph_config" verification? I'm not sure if it's worth breaking backwards compat over it or not, though I doubt anyone but Gecko is using this.

I'm ok landing this as is for now though.

PR #807, opened 2025-10-07
fix: added use_proxy flag to get_artifact_url

abhishekmadan30:
I thought adding a block_proxy flag to get_root_url, would be the best option for get_rool_url so we could avoid having to duplicate the code from get_root_url if use_proxy is not used in get_artifact_url without it accidently picking up the proxy url even if that is set to false

### DIFF
--- a/src/taskgraph/util/taskcluster.py
+++ b/src/taskgraph/util/taskcluster.py
@@ -33,8 +33,8 @@ CONCURRENCY = 50
 
 
 @functools.lru_cache(maxsize=None)
-def get_root_url():
-    if "TASKCLUSTER_PROXY_URL" in os.environ:
+def get_root_url(block_proxy=False):
+    if "TASKCLUSTER_PROXY_URL" in os.environ and not block_proxy:
         logger.debug(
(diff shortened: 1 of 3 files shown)


src/taskgraph/util/taskcluster.py
```diff
@@ -140,8 +140,26 @@ def get_session():
     return requests_retry_session(retries=5)
 
 
-def get_artifact_url(task_id, path):
-    artifact_tmpl = liburls.api(get_root_url(), "queue", "v1", "task/{}/artifacts/{}")
+def get_artifact_url(task_id, path, use_proxy=False):
+    if use_proxy:
```
hneiva, line 144: I wonder if we should bypass this flag if `TASKCLUSTER_PROXY_URL` is not in os.environ

jcristau: -1.  IMO if use_proxy is requested and we don't have a proxy, we should fail rather than silently not use it.

abhishekmadan30: I think it would be better to raise the error as instead if they are expecting a proxy to be use but the environment variable isn't set I think we should throw an error

PR #821, opened 2025-10-16
Drop support for Python 3.8

ahal:
We inadvertently added a dependency that doesn't support 3.8, so this is more like reflecting reality than causing any changes.

### DIFF
--- a/src/taskgraph/actions/registry.py
+++ b/src/taskgraph/actions/registry.py
@@ -30,13 +30,13 @@ def is_json(data):
     return True
 
 
-@functools.lru_cache(maxsize=None)
+@functools.cache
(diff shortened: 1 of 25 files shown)


ahal:
I can be convinced to hold off landing this until firefox-main stops supporting 3.8 as well. It's possible that consumers using Python 3.8 might not be exercising any 3.8 only code paths in the `taskcluster` package. On the other hand, 3.8 is no longer supported and we're probably being overly cautious..

jcristau (approved):

jcristau:
Something seems to be going wrong with the 3.14 test task, looks like it segfaults?  Maybe move that commit out to a separate PR?

PR #824, opened 2025-10-17
Use `links_and_reverse_links_dict` in `replace_tasks`

Eijebong:
In b7134975846a0bc2266213bc64cd4e2b0c953067 we added a function to get both `links_dict` and `reverse_links_dict` in one go, avoiding iterating over the graph twice to build those. That function also benefits from being cached. I'm not expecting a big change in performance here but we might as well.


jcristau (approved):
With #827 this should be fine.

PR #825, opened 2025-10-17
Make sure we return read only dicts from `links_and_reverse_links_dict`

Eijebong:
Because we cache the return value, we don't want to be returning objects with inner mutability as a consumer modifying those would end up modifying the cached value, impacting all future accesses.


jcristau (approved):

jcristau (changes requested):
On second thought.. the values in these ReadOnlyDicts are still mutable...

PR #834, opened 2025-10-23
feat: add support for outputting a mermaid flowchart of the kind graph

bhearsum:
Especially in repositories with large graphs, it can be useful to visualize the relationship between the various kinds.

I looked at directly outputting svg files of the graphs here, but decided against it because rendering mermaid diagrams requires a browser or puppeteer AFAICT; and I don't want to add that dependency.

I also looked at adding support for visualizing entire graphs, but those quickly get very unwieldy and I decided against it. (This might be useful in conjunction with a new `--target-task` option, but that's more than I'm willing to take on at this time.)

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -64,6 +64,31 @@ def format_taskgraph_yaml(taskgraph):
     return yaml.safe_dump(taskgraph.to_json(), default_flow_style=False)
 
 
+def format_kind_graph_mermaid(kind_graph):
+    """
+    Convert a kind dependency graph to Mermaid flowchart format.
+
+    @param kind_graph: Graph object containing kind nodes and dependencies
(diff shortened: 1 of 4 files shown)


src/taskgraph/main.py
```diff
@@ -225,6 +250,70 @@ def logfile(spec):
...
+        overrides={"target-kinds": target_kinds},
+    )
+
+    tgg = get_taskgraph_generator(options.get("root"), parameters)
+    kind_graph = tgg.kind_graph
+
+    output = format_kind_graph_mermaid(kind_graph)
+
+    output_file = options.get("output_file")
+    if output_file:
```
ahal: nit: walrus operator would be apropos here

ahal (approved):
Thanks, looks nifty!

PR #849, opened 2025-10-30
Convert some "check" style transforms into verifications

ahal:


bhearsum (approved):
Nice! I imagine this probably isn't a huge perf win in the decision task for Gecko, but it's probably something, and it is certainly a correctness win.

(I imagine there's probably similar things we should do in Gecko; can we get something on file for that if so?)

PR #850, opened 2025-10-30
fix: support windows in `run-task`

bhearsum:
Taskgraph's version of `run-task` does not currently support Windows. After some investigation, it looks like much of https://bugzilla.mozilla.org/show_bug.cgi?id=1459737 was already ported to it - the only thing that we need to change is to relocate the POSIX-specific imports into a POSIX-specific function.


Eijebong (approved):
LGTM, ruff isn't pleased with this though, you'll have to annotate these to appease it.

PR #857, opened 2025-11-04
fix: allow run-task-hg suffix

coreycb:
It seems that run-tasks can have an -hg suffix.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1998132


ahal:
This is happening because a verification that is meant to verify the results of `taskgraph.transforms.task` is instead verifying the results of `gecko_taskgraph.transforms.task`.

The fix is to not use this verification. Here's how we disabled them in `gecko_taskgraph`:
https://searchfox.org/firefox-main/source/taskcluster/gecko_taskgraph/__init__.py#82

So you can do something similar in comm. You could alternatively iterate through the verifications and only delete the one with the failure if you wanted to keep running all the other ones.

I think once Gecko is migrated to Github we can think about consolidating these verifications.. but for now I'm inclined to not deal with it.

PR #859, opened 2025-11-05
fix infinite loop in cached_tasks transforms

jcristau:
Detect dependency cycle instead of looping forever.

### DIFF
--- a/src/taskgraph/transforms/cached_tasks.py
+++ b/src/taskgraph/transforms/cached_tasks.py
@@ -3,9 +3,8 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
-from collections import deque
-
 import taskgraph
+from taskgraph.graph import Graph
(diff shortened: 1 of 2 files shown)


src/taskgraph/graph.py
Eijebong: Should probably change this? TBH it was already wrong since it wouldn't hang anymore but now it is well defined what happens

src/taskgraph/transforms/cached_tasks.py
```diff
@@ -15,25 +14,20 @@
...
+        (label, dep, "")
+        for label in pending
+        for dep in pending[label].get("dependencies", {}).values()
+        if dep.startswith(kind_prefix)
+    }
+    graph = Graph(nodes, edges)
+
+    for label in graph.visit_postorder():
+        yield pending.pop(label)
+    if pending:
```
Eijebong: visit_postorder would raise though if there was a cycle?

PR #860, opened 2025-11-05
feat: add --develop flag to `taskgraph load-task`

ahal:
This flag will setup the current local checkout as a volume to the task's checkout and avoid all VCS operations.

This allows developers to make changes to the source and immediately see it in action in the task.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -767,9 +767,18 @@ def image_digest(args):
     action="store_true",
     default=False,
     help="Setup the task but pause execution before executing its command. "
-    "Repositories will be cloned, environment variables will be set and an"
+    "Repositories will be cloned, environment variables will be set and an "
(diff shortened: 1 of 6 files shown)


src/taskgraph/main.py
```diff
@@ -767,10 +767,17 @@ def image_digest(args):
     action="store_true",
     default=False,
     help="Setup the task but pause execution before executing its command. "
-    "Repositories will be cloned, environment variables will be set and an"
+    "Repositories will be cloned, environment variables will be set and an "
     "executable script named `exec-task` will be provided to resume task "
     "execution. Only supported for `run-task` based tasks.",
 )
+@argument(
+    "--develop",
```
bhearsum: nit: `--develop` doesn't really describe what this does. I can imagine some users preferring `--develop`, but adding an alias like `--use-local-checkout` might also be nice.

bhearsum (approved):

PR #861, opened 2025-11-06
fix(load-task): consider -hg and -git suffixes in run-task check

ahal:
This supports using `load-task` in Gecko, where `run-task`, `run-task-git` and `run-task-hg` are all valid script names.


jcristau (approved):
Thanks.  Do you mind adding a reference to the bug to the commit message?

ahal:
I knew you were going to suggest that, couldn't update in time :p

PR #874, opened 2025-12-08
fix: get_artifact_from_index now uses artifact_path instead of index_path(bug 1999039)

abhishekmadan30:


abhishekmadan30:
I updated the bug number attached to this as they split the bugs. This is a fix for TypeError: '_io.BytesIO' object is not subscriptable.  I will create a separate patch for auth failure

PR #878, opened 2025-12-10
ci: layer the python docker image on top of run-task

jcristau:


taskcluster/kinds/docker-image/kind.yml
ahal: Maybe just define this in-line now?

ahal (approved):

PR #920, opened 2026-03-13
Optimize: Update transitive closure algorithm

hneiva:
The previous implementation scanned all edges on every iteration of a fixed-point loop, resulting in O(depth * edges) complexity. For deep dependency chains this is catastrophic - a 5000-node chain required 5000 full passes over all edges.

Replace with a standard Breadth-First Search using an adjacency list built once upfront, reducing complexity to O(vertices + edges). Benchmarks on large graphs show 22x-1333x speedups depending on graph shape, with identical results verified against the original algorithm.


bhearsum (approved):
I'm -0 on this...it adds complexity for no real world win. As long as it doesn't hurt perf in our real world cases I guess it's fine to take.

ahal (approved):
Seems fine to me. It's a few more lines of code, but I can't really say it's much more complex than the old algorithm

ahal:
This is superseded by https://github.com/taskcluster/taskgraph/pull/1014 (sorry I forgot this existed). I think we should take the other one because it includes some new tests

PR #922, opened 2026-03-16
feat: add CodSpeed benchmark tests for graph operations

hneiva:
Add pytest-codspeed benchmarks testing core graph operations (transitive closure, topological sort, link dicts, for_each_task, JSON round-trip, and transform sequences) across 4 graph geometries with 1000 tasks each: linear chain, fan-out/fan-in, binary tree, and diamond layers.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -1,0 +1,30 @@
+---
+name: CodSpeed
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  # `workflow_dispatch` allows CodSpeed to trigger backtest
+  # performance analysis in order to generate initial data.
+  workflow_dispatch:
+
+jobs:
+  benchmarks:
+    name: Run benchmarks
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read  # required for actions/checkout
+      id-token: write  # required for OIDC authentication with CodSpeed
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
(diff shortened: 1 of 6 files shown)


.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+      - uses: astral-sh/setup-uv@fac544c07dec837d0ccb6301d7b5580bf5edae39  # v8.2.0
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: uv sync
+      - name: Run benchmarks
+        uses: CodSpeedHQ/action@v4
+        with:
+          run: uv run pytest test/ --codspeed
```
hneiva, line 30: tbh I'm unsure if we should test everything or just `test/test_graph_perf.py`

bhearsum: I can't imagine why we'd want to run it on the non-perf tests? (In fact, we should probably exclude the new perf test from running as a unit test as well?)

.github/workflows/codspeed.yml
```diff
@@ -0,0 +1,30 @@
+---
```
bhearsum, line 1: What's the advantage to running this as a GHA instead of in taskcluster? (I presume there is one; it should probably be documented as a top level comment.)

ahal (approved):
Thanks!

.github/workflows/codspeed.yml
hneiva: Mostly for ease of use and maintainability.
If we use TC we'd also need to set it up with a "legacy Token" auth method on Codspeed.
GH has OIDC authentication built-in and works out of the box with the service.

PR #925, opened 2026-03-31
Bug 2028208 - Support updating parameters via Git note

ahal:


jcristau (approved):
lgtm, just one small comment (which won't matter much in practice if there's a single extra ref)

PR #928, opened 2026-04-07
fix: ensure `taskgraph full -J --tasks <regex>` shows all dependencies

ahal:
Overwriting `task.dependencies` with the graph edges doesn't play nicely with `get_filtered_taskgraph` (e.g `--tasks <regex>`) because in this case the full graph has already been trimmed. The result is that when using `--tasks`, the dependencies displayed will be a subset of the real dependencies. This is very misleading and can easily lead one to false conclusions while debugging. Recently this caused me to waste an hour of my time.

I traced this comment back through history and it's been like this since it was originally implemented. Likely the intent was that if we a have a subset of the graph, then we should modify the task definition to not display any dependencies that aren't part of it.

But I disagree with this. I'd argue that `Task.dependencies` is a very important part of the task definition, and not at all relevant to the higher level DAG that the task happens to be a part of. I believe mutating the task definition just because we did some filtering is a bug.

Testing the full graph from the Firefox repo, I can confirm this change results in identical graphs (when not using `--tasks`).

### DIFF
--- a/src/taskgraph/taskgraph.py
+++ b/src/taskgraph/taskgraph.py
@@ -44,13 +44,10 @@ class TaskGraph:
 
     def to_json(self):
         "Return a JSON-able object representing the task graph, as documented"
-        named_links_dict = self.graph.named_links_dict()
         # this dictionary may be keyed by label or by taskid, so let's just call it 'key'
         tasks = {}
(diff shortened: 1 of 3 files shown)


hneiva (approved):
LGTM!

PR #934, opened 2026-04-15
Prevent tests from using the user and system git config

Eijebong:
This prevents tests from asking to sign dummy commits on that repo if
there's a global gpgsign config set. This also has the nice side effect
of having the same tests not pollute the user git config with random
safe directories (since run_task calls `git config --global` which tests
exercise).

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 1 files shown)


jcristau (approved):
Alternatively should we set GIT_CONFIG_GLOBAL=/dev/null (and possibly GIT_CONFIG_SYSTEM) for the tests?

test/conftest.py
```diff
@@ -13,9 +14,12 @@
 # interference with tests.
 # This ignores ~/.hgrc
 os.environ["HGRCPATH"] = ""
-# This ignores system-level git config (eg: in /etc). There apperas to be
-# no way to ignore ~/.gitconfig other than overriding $HOME, which is overkill.
+# This ignores system-level git config (eg: in /etc).
 os.environ["GIT_CONFIG_NOSYSTEM"] = "1"
+# run-task writes to the global git config, so point it at a real
+# (writable) empty file instead of ~/.gitconfig.
+git_global = tempfile.NamedTemporaryFile(prefix="taskgraph-gitconfig-")
```
jcristau: Might want to move that to a (autouse/session-scope?) fixture so you can delete the file at the end?

Eijebong: AFAIK it'll get deleted at the end when it gets GC'ed on interpreter exit.

jcristau: Oh, right.

jcristau (approved):

PR #935, opened 2026-04-15
Move git notes fetching from run-task to the decision task and make them optional

Eijebong:
This removes the concept of extra refs fetching as it was introduced for this feature alone as it turned out to be quite complex to support if we wanted notes to be optional. The main issue it's fixing is being able to run decision tasks off of repos that don't have that notes branch at all.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -31,7 +31,7 @@ import time
 import urllib.error
 import urllib.request
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, Optional
 
 SECRET_BASEURL_TPL = "{}/secrets/v1/secret/{{}}".format(
     os.environ.get("TASKCLUSTER_PROXY_URL", "http://taskcluster").rstrip("/")
@@ -640,7 +640,6 @@ def git_checkout(
     ssh_key_file: Optional[Path],
     ssh_known_hosts_file: Optional[Path],
     shallow: bool = False,
-    extra_refs: Optional[List[str]] = None,
 ):
     assert head_ref or head_rev
 
@@ -734,14 +733,6 @@ def git_checkout(
         shallow=shallow,
         env=env,
     )
-
-    if extra_refs:
-        git_fetch(
-            destination_path,
-            *[f"{ref}:{ref}" for ref in extra_refs],
-            remote=head_repo,
-            env=env,
-        )
 
     args = [
         "git",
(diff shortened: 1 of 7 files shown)


src/taskgraph/run-task/run-task
```diff
@@ -584,7 +585,22 @@ def git_fetch(
         args.append("--depth=1")
 
     args.extend([remote, *set(targets)])
-    retry_required_command(b"vcs", args, cwd=destination_path, extra_env=env)
+    try:
+        retry_required_command(b"vcs", args, cwd=destination_path, extra_env=env)
```
jcristau: If it's optional then we might not want to retry before we've checked ls-remote?

Eijebong: Right... This is so annoying...

src/taskgraph/run-task/run-task
```diff
@@ -735,11 +760,15 @@ def git_checkout(
         env=env,
     )
 
-    if extra_refs:
+    for ref in extra_refs or []:
+        optional = ref.startswith("?")
+        if optional:
+            ref = ref[1:]
         git_fetch(
```
jcristau: Can we avoid calling git_fetch once per ref?

Eijebong: I don't think so? Cause we cannot extract what ref failed to fetch, so we'd have to check them all after the fact

Eijebong:
(I dropped the taskcluster.yml change and pushed empty notes to my fork to get CI green, I'll re-add the `?` change in a followup and delete notes from my fork to test properly)

src/taskgraph/run-task/run-task
```diff
@@ -735,11 +760,15 @@ def git_checkout(
         env=env,
     )
 
-    if extra_refs:
+    for ref in extra_refs or []:
+        optional = ref.startswith("?")
```
ahal: Tbh, I think I'd rather these just never fail, and if there's some use case where we absolutely *need* the extra ref fetched, then we can deal with it then

src/taskgraph/util/vcs.py
```diff
@@ -594,13 +602,30 @@ def does_revision_exist_locally(self, revision):
...
+        remote: str,
+        revision: Optional[str] = None,
+    ) -> Optional[str]:
         if not note.startswith("refs/notes/"):
             note = f"refs/notes/{note}"
 
-        revision = revision or "HEAD"
         try:
-            return self.run("notes", f"--ref={note}", "show", revision).strip()
+            self.run("fetch", remote, f"{note}:{note}")
```
Eijebong, line 615: One downside of doing it here is that we lose the retry path completely

jcristau: And we call fetch on each `get_note` instead of just once?  (Granted we call get_note just once anyway)

src/taskgraph/util/vcs.py
```diff
@@ -594,13 +602,30 @@ def does_revision_exist_locally(self, revision):
...
+        except subprocess.CalledProcessError:
+            ls = subprocess.run(
+                ["git", "ls-remote", "--exit-code", remote, note],
+                cwd=self.path,
+                capture_output=True,
+            )
+            if ls.returncode != 2:
```
jcristau: could we return here if if `ls.returncode == 2`?

jcristau (approved):

src/taskgraph/util/vcs.py
Eijebong: Sure, that matches the pattern just below

PR #938, opened 2026-04-22
Add a way to prevent some actions from getting generated

Eijebong:
Right now taskgraph always generates some generic actions that are used on fxci on some projects. Projects that do not need to use said actions are forced to add `rm actions.json` at the end of their decision task to not get basic taskcluster features overridden by taskgraph (rerun/retrigger mainly).

This adds a `disabled_actions` graph configuration that allows giving a list of actions that should not be emitted in the actions.json.

### DIFF
--- a/src/taskgraph/actions/registry.py
+++ b/src/taskgraph/actions/registry.py
@@ -273,12 +273,27 @@ def render_actions_json(parameters, graph_config, decision_task_id):
             artifact.
     """
     assert isinstance(parameters, Parameters), "requires instance of Parameters"
+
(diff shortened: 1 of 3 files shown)


Eijebong:
https://github.com/Bananium/staging-archipelago-index/commit/f83f91fd96052ba955ea39e6452b34ce23ee63fe -> https://taskcluster.bananium.fr/tasks/e2f_Va_mQQm5-6aEQ1q6cw (no actions emitted \o/)

ahal (approved):
Thanks, lgtm.

PR #939, opened 2026-04-23
Support defining Schemas as dicts

ahal:
Msgspec schemas are much faster, but I find the struct based definitions to look clunky, especially with many nested structs. This uses `msgspec.defstruct` to support defining schemas as a dictionary, somewhat similar to how the voluptuous schemas used to look.

I converted one file (`from_deps.py`) as an example of what the new format looks like. Will update the rest in a follow-up PR.

### DIFF
--- a/docs/howto/define-schemas.rst
+++ b/docs/howto/define-schemas.rst
@@ -1,0 +1,156 @@
+Define Schemas
+==============
+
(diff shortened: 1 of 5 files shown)


hneiva (approved):
Not super familiar with how MsgSpec Schemas work, but LGTM

PR #940, opened 2026-04-29
feat!: convert parameter schema from voluptuous to msgspec

abhishekmadan30:


src/taskgraph/parameters.py
```diff
@@ -28,44 +29,52 @@ class ParameterMismatch(Exception):
...
-        # used at run-time
-        Required("target_tasks_method"): str,
-        Required("tasks_for"): str,
-        Required("version"): Any(str, None),
-        Optional("code-review"): {
-            Required("phabricator-build-target"): str,
-        },
-    }
-)
+class BaseParametersSchema(
```
ahal: Could you make this use `Schema.from_dict` instead?

taskcluster/self_taskgraph/custom_parameters.py
```diff
@@ -15,12 +15,18 @@ def get_defaults(repo_root):
...
-)
+class CustomParametersSchema(Schema, kw_only=True, rename=None):
+    pull_request_number: Optional[int]
+
+    def __post_init__(self):
+        super().__post_init__()
+        if self.pull_request_number is not None and self.pull_request_number < 1:
+            raise ValueError(
+                f"pull_request_number must be >= 1, got {self.pull_request_number}"
+            )
```
ahal: Might as well be consistant and use `from_dict` here too. You can use https://jcristharif.com/msgspec/api.html#msgspec.Meta to support the > 0 constraint.

ahal (changes requested):
Thanks, looks good!

I'd like to standardize on the `from_dict` format though, so maybe we can land new code in that format.

ahal (approved):
Thanks, lgtm!

ahal:
One minor nit, but if we're following conventional commits, the prefix should be `feat!` (exclamation mark means breaking). Also there's supposed to be somehting like:
```
BREAKING CHANGE: <reason>
```

in the commit body. No big deal because we don't enforce it here, so more of an fyi

PR #947, opened 2026-05-05
ci: use shared pre-commit hook

ahal:


ahal:
Oh right, I only granted scopes in mozilla-releng. Will need to add this to taskgraph. Also the large number of comments is due to https://github.com/mozilla-releng/fxci-config/pull/974 not being landed yet ><

bhearsum (approved):

bhearsum:
superceded by #964 probably?

ahal:
oops

PR #953, opened 2026-05-14
fix: skip external task references when rebuilding edges in from_json

abhishekmadan30:


bhearsum (dismissed):
This needs some more explanation and context. I think it's probably a regression, and if so, pointing at what caused it, and a more detailed breakdown of the issue and why this fixes it would be useful.

abhishekmadan30:
Regression from #928, which stopped to_json rewriting task["dependencies"] from graph edges. I think in optimized/morphed graphs Task.dependencies contains taskIds of replaced/cached tasks that aren't nodes here. to_json now writes them through, and from_json adds them as edges, so the reconstructed Graph has edges to non-nodes and links_and_reverse_links_dict raises KeyError.

bhearsum (commented):
Thanks for the additional context. A pointer to where this is actually causing problems would still be useful.

I'm a bit iffy and whether or not this is the right fix. As I understand it, some of the JSON we're getting fed contains tasks that reference dependencies that don't exist. Arguably, that makes that JSON an invalid taskgraph that shouldn't be accepted.

Perhaps we should adjust whatever use case is feeding in such taskgraphs instead? Maybe by having them feed a different graph, or massaging the graph before feeding it in? Or if we still want taskgraph to be able to accept these, maybe it should require an extra flag that explicitly tells it to skip edges if the upstream task doesn't exist (although that still feels like a future footgun to me).

@ahal - flagging you since you authored #928 in case you want to weigh in.

src/taskgraph/taskgraph.py
```diff
@@ -63,6 +63,7 @@ def from_json(cls, tasks_dict):
             if "task_id" in value:
                 tasks[key].task_id = value["task_id"]
             for depname, dep in value["dependencies"].items():
-                edges.add((key, dep, depname))
+                if dep in tasks_dict:
```
ahal, line 67: Can you add a comment along the lines of "Task filtering can cause dependencies to be removed from the graph" or something like that?

ahal (approved):
Thanks, this makes sense to me

ahal:
> I'm a bit iffy and whether or not this is the right fix. As I understand it, some of the JSON we're getting fed contains tasks that reference dependencies that don't exist. Arguably, that makes that JSON an invalid taskgraph that shouldn't be accepted.

We had a conversation prior to #928 where IIRC we agreed that flags like `--tasks <regex>` should modify the structure of the graph, but not the contents of the tasks in the graph. This meant that we stopped re-writing dependencies in the Task objects when calling `to_json`.

But I neglected to update `from_json` which can now attempt to create edges to tasks that were filtered out by the aforementioned `--tasks` flag. This fix makes sense to me

PR #958, opened 2026-05-21
feat(tests): Add unit-serial to test TASKGRAPH_SERIAL

hneiva:


taskcluster/kinds/test/linux.yml
```diff
@@ -48,7 +48,7 @@ unit:
         symbol: unit(py{matrix[python]})
     run:
         command: >-
-            uv run coverage run --data-file /builds/worker/artifacts/coverage --context=py{matrix[python]} -m pytest -vv
+            uv run --python={matrix[python]} coverage run --data-file /builds/worker/artifacts/coverage --context=py{matrix[python]} -m pytest -vv
```
bhearsum: This isn't necessary actually; it's set in the `UV_PYTHON` env var further up.

taskcluster/kinds/test/linux.yml
```diff
@@ -68,7 +68,27 @@ unit-multithread:
...
+
+unit-serial:
+    description: "Run unit tests with py{matrix[python]} on Linux in serial mode"
+    matrix:
+        set-name: "unit-serial-py{matrix[python]}"
+        substitution-fields: [description, run.command, treeherder, worker, attributes]
+        python: ["314t"]
```
bhearsum: Let's run this against 3.14 -- 3.14t is specifically meant for testing free threaded Python.

bhearsum (changes requested):
It looks like there's a test failure with `TASKGRAPH_SERIAL` currently? 😬. All the more reason to run these...

taskcluster/kinds/test/linux.yml
hneiva: TIL

hneiva:
Updated PR title + commit message to reflect changes

test/test_generator.py
```diff
@@ -23,6 +23,9 @@
     not os.environ.get("TASKGRAPH_USE_THREADS"),
     reason="requires multithreading to be enabled",
 )
+asynconly = pytest.mark.skipif(
```
bhearsum:
```suggestion
multiprocessonly = pytest.mark.skipif(
```

(async and multiprocess are not the same thing.)

test/test_generator.py
```diff
@@ -23,6 +23,9 @@
...
+    os.environ.get("TASKGRAPH_SERIAL"), reason="requires TASKGRAPH_SERIAL to be 0"
```
bhearsum:
```suggestion
    os.environ.get("TASKGRAPH_SERIAL"), reason="requires multiprocessing to be enabled"
```

(This is checking whether or not `TASKGRAPH_SERIAL` is not at all, not looking for whether or not it is 0.)

test/test_generator.py
```diff
@@ -57,6 +61,7 @@ def test_kind_ordering_multiprocess(mocker, maketgg):
 
 
 @threadsonly
+@asynconly
```
bhearsum: I think it would be better to fix the `threadsonly` marker to account for `TASKGRAPH_SERIAL` rather than stack these in this case, but this works too.

bhearsum (approved):
Approving, but please consider applying the suggestions.

PR #961, opened 2026-05-25
fix(tests): preserve safe directories in git config fixture

bhearsum:
AFAICT https://github.com/taskcluster/taskgraph/commit/7e30651ef3a43ca6010b4883529db0042b22686b introduced this new intermittent failure.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,5 +1,6 @@
 import os
 import platform
+import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


ahal (approved):
Maybe we could just skip overwriting the config when in CI and only do it for local runs? But this seems fine too

Eijebong (approved):

PR #968, opened 2026-06-01
feat: allow resolving keys in task context

bhearsum:
This extends task-context to allow for calling `resolve_keyed_by` is a generic way. Combined with `task-context's` existing support for pulling context from parameters it removes the need for transforms for many usages of `resolve_keyed_by` (I counted at least a dozen or so in Gecko alone, and there's probably many more.)

Mostly for simplicitly I chose to re-use `substitution-fields` as a source for what `resolve_keyed_by` should be called on. I could be convinced to maintain a separate list for that (although it might introduce repetition in certain cases). I'm honestly not sure which is preferable.

### DIFF
--- a/docs/reference/transforms/task_context.rst
+++ b/docs/reference/transforms/task_context.rst
@@ -4,8 +4,8 @@ Task Context Transforms
 =======================
 
 The :mod:`taskgraph.transforms.task_context` transform can be used to
-substitute values into any field in a task with data that is not known
-until ``taskgraph`` runs.
+resolve keys and/or substitute values into any field in a task with data
+that is not known until ``taskgraph`` runs.
 
 This data can be provided in a few ways, as described below.
 
@@ -37,7 +37,10 @@ Then create a ``task-context`` section in your task definition, e.g:
 
    tasks:
      build:
-       description: my description {foo}
+       description:
+        by-foo:
+          bar: super special description for bar tasks
+          default: my description of {foo}
        task-context:
          from-parameters:
            foo: foo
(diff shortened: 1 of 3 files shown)


docs/reference/transforms/task_context.rst
```diff
@@ -37,7 +37,10 @@ Then create a ``task-context`` section in your task definition, e.g:
 
    tasks:
      build:
-       description: my description {foo}
+       description:
+        by-foo:
+          FOO: super special description for FOO tasks
+          default: my description of {foo}
```
ahal, line 43: You're just extending the existing example so feel free to ignore, but I think a more real-world example would make this easier to understand.

docs/reference/transforms/task_context.rst
```diff
@@ -50,9 +53,12 @@ are used:
 
 .. code-block:: yaml
 
-   foo: with some extra
+   foo: FOO
```
ahal: If we don't make the example more real-world, let's at least set the value to `bar` here instead of another `FOO`

src/taskgraph/transforms/task_context.py
```diff
@@ -89,6 +100,12 @@ def render_task(config, tasks):
         if "name" in task:
             subs.setdefault("name", task["name"])
 
-        # Now that we have our combined context, we can substitute.
+        # Now that we have our combined context, we can resolve keys and substitute.
+        task_name = task.get("name", task.get("label", "?"))
+        resolve_keyed_by_options = {}
+        if opts := sub_config.get("resolve-keyed-by-options"):
+            resolve_keyed_by_options.update(opts)
```
ahal: nit: `resolve_keyed_by_options = sub_config.get("resolve-keyed-by-options") or {}`

ahal (approved):
Thanks, lgtm modulo some nits

ahal:
Oh and you'll want to rebase on main to get the new pre-commit-v1 task to unblock landing

docs/reference/transforms/task_context.rst
bhearsum: I think it's worthwhile to improve the example; will do.

bhearsum: I ended up keeping the existing example but adding a new one that's much more real world.

PR #969, opened 2026-06-02
fix: replace uses of Parameters.is_try() with level check

ahal:
In the Github world, what is considered "try" or not is much less defined than it is on hgmo. Currently we have "enterprise-firefox-try", "firefox-dev" and "staging-firefox" that are all "try-like" in that they will run tasks on any arbitrary branch push.

We can no longer rely on the magic string "try" being present in the repo name to determine this, and imo nor should we ever have been.

At least in this repo, all uses of `is_try()` are probably better off just using `level == 1` instead.

Bug: 2044330

### DIFF
--- a/docs/reference/migrations.rst
+++ b/docs/reference/migrations.rst
@@ -8,6 +8,11 @@ This page can help when migrating Taskgraph across major versions.
 
 * You must now pass in `--allow-parameter-override` into Decision tasks that
   need to load the `try_task_config.json` file at the root of the repo.
+* `taskgraph.parameters.Parameters.is_try()` is removed. Either in-line the old
+  method to a utility or use alternative logic.
+* Level 1 artifacts' default expiry changed from "1 year" to "28 days". If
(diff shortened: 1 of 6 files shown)


ahal:
I debated if this should be considered breaking or not.. I don't think it materially changes graphs, just might reduce artifact expiry in some cases, so I opted for not

ahal:
Actually, I just realized that gecko_taskgraph has its own `is_try` logic.. so I think I'll just remove `Parameters.is_try()` and make this breaking.

bhearsum (approved):
This is a good improvement. I think we should get rid of the very Mozilla-specific `try_task_config_file` in the future as well, but that's out of scope here.

ahal:
> This is a good improvement. I think we should get rid of the very Mozilla-specific `try_task_config_file` in the future as well, but that's out of scope here.

Once the Git note is working well and Gecko is migrated to it we can remove support for the file. Though `v2` of the file itself isn't actually Mozilla specific, it's just a way to allow users to override parameters :)

PR #971, opened 2026-06-03
feat: add support for custom context to task-context

bhearsum:
This is somewhat of a continuation to #968 in that it would allow for more `resolve_keyed_by` calls to be replaced by `task-context`. As a concrete example, we use [`by-release-level` in many places in gecko](https://searchfox.org/firefox-main/search?q=by-release-level&path=&case=false&regexp=false). Its value is [derived from parameters and some constants](https://searchfox.org/firefox-main/source/taskcluster/gecko_taskgraph/util/attributes.py#157). With this patch, we could register it as custom context and replace a whole bunch of `resolve_keyed_by` calls in transforms, eg:

In some python code that's called at taskgraph registration:
```
@custom_context("release-level")
def release_level_context(config, task):
    """Provide ``release-level`` ("production" or "staging"), derived from
    parameters, so ``by-release-level`` fields can be resolved declaratively."""
    return {"release-level": release_level(config.params)}
```

And in various kinds we'd end up with things like:
```
transforms:
    - taskgraph.transforms.task_context

task-defaults:
    task-context:
        from-custom:
            - release-level
        substitution-fields:
            - scopes

    scopes:
        by-release-level:
            production:
                - project:releng:ship-it:server:production
                - project:releng:ship-it:action:create-new-release
            staging:
                - project:releng:ship-it:server:staging
                - project:releng:ship-it:action:create-new-release
```

### DIFF
--- a/docs/reference/transforms/task_context.rst
+++ b/docs/reference/transforms/task_context.rst
@@ -98,35 +98,6 @@ with the this ``kind``:
 ...the description will bring in the value ``foo`` from the parameters if
 present, or ``default`` otherwise.
 
-from-file
-~~~~~~~~~
-
-Context may also be provided from a defined yaml file. The provided file
-should usually only contain top level keys and values (eg: nested objects
-will not be interpolated - they will be substituted as text representations
-of the object).
-
-For example, with this kind definition:
-
-.. code-block:: yaml
-
-   tasks:
-     build:
-       description: my description {foo}
-       task-context:
-         from-file: some_file.yaml
-         substitution-fields:
-           - description
-
-And this in ``some_file.yaml``:
-
-.. code-block:: yaml
-
-   foo: from a file
-
-...description will end up with "my description from a file".
-
-
 from-object
 ~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


jcristau (commented):
Consider adding to task_context transform docs?

bhearsum:
> Consider adding to task_context transform docs?

er, yes! I meant to do that before moving it out of draft.

src/taskgraph/transforms/task_context.py
```diff
@@ -89,10 +97,15 @@ def render_task(config, tasks):
         if from_file:
             file_context = load_yaml(from_file)
 
+        custom_context = {}
+        for name in sub_config.pop("from-custom", None) or []:
+            custom_context.update(CUSTOM_CONTEXT_MAP[name](config, task))
```
jcristau, line 105: Should we check that `name` is in `CUSTOM_CONTEXT_MAP` and return something more friendly than `KeyError`?

docs/reference/transforms/task_context.rst
```diff
@@ -185,7 +236,7 @@ Keys will be resolved on ``substitution-fields`` first, then substitution
 will be performed on the resolved value.
 
 If the same key is found in multiple places the order of precedence is as
-follows: ``from-parameters``, ``from-object`` keys, ``from-file`` and finally
+follows: ``from-parameters``, ``from-object`` keys, ``from-custom`` providers, ``from-file`` and finally
```
jcristau, line 239: Do the tests cover the order of precedence?

jcristau (approved):

src/taskgraph/transforms/task_context.py
bhearsum: Yeah, that's a good idea.

docs/reference/transforms/task_context.rst
bhearsum: They do now!

PR #980, opened 2026-06-09
revert: hash pinning on pypi publish workflow

bhearsum:


ahal (approved):
Will renovate try and change this back to using a hash?

I think as long as we're reasonably confident that renovate is pulling updates from release branches, supporting the hash format in the trusted publisher seems fine..

But this is fine too. We just may need to tweak the renovate config to not use hashes for this one.

PR #981, opened 2026-06-09
revert: hash pinning on pypi publish workflow

bhearsum:
Missed these ones in #980

(We're already talking about accepting the pinning and loosening the restrictions on the versions of this action that may run; but we may want this to unbreak releases in the meantime.)

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -44,7 +44,8 @@ jobs:
           pip install build
           python -m build
       - name: Publish package distributions to PyPI
-        uses: pypa/gh-action-pypi-publish@cef221092ed1bacb1cc03d23a2d87d1d172e277b  # v1.14.0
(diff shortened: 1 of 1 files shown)


ahal (changes requested):
I went ahead and relaxed the `pypi-publish` restriction in the settings. After looking into it, I think we should keep the hash pinning.

ahal:
Outlined my rationale in channel, copy/pasting here:

I think I understand.. The release/v1 branch is like a latest tag.. it points to whatever the latest release on the 1.x branch line is:
https://github.com/pypa/gh-action-pypi-publish/tree/release/v1

but because we're using renovate now, it'll find the tag refs directly.. so it's likely equivalent to using release/v1 but we get to update explicitly instead of implicitly

bhearsum:
ok, sure

bhearsum:
https://github.com/taskcluster/taskgraph/pull/982 for moving back to hash pinning.

PR #982, opened 2026-06-09
repin pypi-publish action

bhearsum:
Quoting context from @ahal in #981:

> Outlined my rationale in channel, copy/pasting here:
> 
> I think I understand.. The release/v1 branch is like a latest tag.. it points to whatever the latest release on the 1.x branch line is: https://github.com/pypa/gh-action-pypi-publish/tree/release/v1
> 
> but because we're using renovate now, it'll find the tag refs directly.. so it's likely equivalent to using release/v1 but we get to update explicitly instead of implicitly

### DIFF
--- a/.github/workflows/pypi-publish.yml
+++ b/.github/workflows/pypi-publish.yml
@@ -21,9 +21,14 @@ jobs:
         run: |
           pip install build
           python -m build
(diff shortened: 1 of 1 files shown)


ahal (approved):

ahal:
Thanks! Sorry for the churn

PR #983, opened 2026-06-10
feat: add opt-in checkout-bundle option to run-task

hneiva:
Add a per-project `--{project}-checkout-bundle=<dest>` option to the run-task script. When set, run-task writes a `.tar.zst` archive of the checkout's working tree (excluding `.git`/`.hg`) right after checkout, so a decision task can publish its checkout as an artifact that downstream tasks fetch via `fetches:` instead of re-cloning.

The archive uses the same `.tar.zst` format as fetch-content's git-checkout-archive and is created fail-fast: archiving errors exit non-zero rather than leaving a broken artifact behind.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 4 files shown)


jcristau:
I don't think this belongs in run-task if it's meant for the decision task to use?

I'm also curious why you decided to tar up the checkout, excluding the vcs data, instead of the other way around.

hneiva:
I thought that was a good place for it since it wraps the decision call, but looking at it now it makes more sense to add it to decision instead. I'll create a new PR for that.

PR #987, opened 2026-06-11
feat(decision): add opt-in --vcs-bundle option to decision task

hneiva:
Add a `--vcs-bundle` flag to `taskgraph decision` that writes a native git/hg bundle of the checkout (full history, all refs) to `public/vcs.bundle`, letting downstream tasks seed a checkout from it instead of re-cloning. The flag is off by default. Bundle creation lives on the `Repository` abstraction via a new `create_bundle` method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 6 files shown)


jcristau (changes requested):
Before reviewing this we should experiment and find out whether/how much it improves things, and play with different methods to select the most useful.

PR #988, opened 2026-06-12
fix(diff): release graphs from memory as they are completed

ahal:
Because the `as_completed()` call was outside of the `with` scope, we were waiting for all futures to complete before processing any of them. Also we were storing futures (which contain the full output) in a dict.

This meant we were holding the output of all graph generations in memory simultaneously. In Gecko, this will eat through your available RAM in no time.


bhearsum (approved):
oof.

PR #994, opened 2026-06-29
feat: --target-kind with -Y or -J implicitly adds --tasks-regex

bhearsum:
When debugging an individually kind or task it's very common to pass `--target-kind` along with `-Y` or `-J`. In larger repos, doing this without also limiting the tasks that will be printed can cause taskgraph to take multiple minutes to write out all tasks to disk, which are usually ignored (and in fact, in the way of finding the output that does matter). This is easily worked around by passing `--tasks-regex` to limit the output. In most cases, this ends up being largely a repeat of `--target-kind`, and it is also not something that everybody knows about. For these reasons, we should imply it in this specific circumstance, and allow it to be overridden if necessary.

### DIFF
--- a/src/taskgraph/main.py
+++ b/src/taskgraph/main.py
@@ -415,7 +415,10 @@ def show_kind_graph(options):
     "--tasks-regex",
     "--tasks",
     default=None,
-    help="only return tasks with labels matching this regular expression.",
+    help="only return tasks with labels matching this regular expression. "
+    "When --target-kind is given with --json/--yaml and this is omitted, it "
+    "defaults to a regex matching labels that begin with one of the target "
+    "kinds.",
 )
 @argument(
     "--exclude-key",
@@ -499,8 +502,25 @@ def show_taskgraph(options):
         )
         print(f"Generating {options['graph_attr']} @ {cur_rev}", file=sys.stderr)
 
+    target_kinds = options.get("target_kinds")
+    if (
+        target_kinds
+        and options.get("tasks_regex") is None
+        and options.get("format") in ("yaml", "json")
+    ):
+        options["tasks_regex"] = "^({})".format(
+            "|".join(re.escape(kind) for kind in target_kinds)
+        )
+        print(
+            "Filtering tasks with implied --tasks-regex "
+            f"'{options['tasks_regex']}' (derived from --target-kind). "
+            "Pass --tasks-regex explicitly to override, or '--tasks-regex .' "
+            "to disable filtering.",
+            file=sys.stderr,
+        )
+
     overrides = {
-        "target-kinds": options.get("target_kinds"),
+        "target-kinds": target_kinds,
     }
     parameters: list[Any[str, Parameters]] = options.pop("parameters")
     if not parameters:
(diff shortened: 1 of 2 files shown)


ahal (changes requested):
A wrinkle here is that there's no guarantee that a label starts with its kind, that's just a convention. The `mochitest` / `reftest` kinds in Gecko are some examples of this, and IIRC there are a bunch of release labels that also don't start with their kind.

A secondary wrinkle is that sometimes I'm working on parent/child tasks at the same time, so I'll specify `--target-kind <child kind>` with the expectation that the output will also include its parent. I could explicitly pass `--tasks` to workaround this, and I'll agree that the single task use case is more common.. so this one is more of a minor one. But just to say that this comes at the expense of making a different use case less convenient.

But both these wrinkles combined make me lean towards not taking the patch. I could be convinced otherwise though if you have a rebuttal!

bhearsum:
That's all fair; I don't typically have use cases like that, but I'm not surprised if someone else does. I'm inclined not to take this as-is as well, in that case. (I imagine there's others that do what you do, and similar use cases that this would harm.)

I wonder about maybe having a sentinel value for `--tasks-regex` that makes it act this way...or maybe a new flag that adds this behaviour...although I'm not entirely convinced of either of those ideas. I'll do some more thinking about this.

ahal:
We could do a sentinel or something, but would that be any less typing than `--tasks-regex "^<kind>"`?

To be clear I don't think my use case is very important, and we shouldn't prioritize it. To me the first point is the bigger problem. E.g, with this patch running:
```
./mach taskgraph full -k mochitest
```

Wouldn't display any tasks, even though they exist. I think that's the deal breaker with this implementation. It's fundamental w.r.t to `--tasks-regex` too, there's no way for us to know what the label will be ahead of time. If we want to fix this I think we'll need some kind of brand new flag that filters based on `Task.kind` rather than `Task.label`.

ahal:
> kind of brand new flag that filters based on Task.kind rather than Task.label.

Or I guess in the spirit of this patch we make filtering by kind the default behaviour and then add a flag to opt out of that filtering.

bhearsum:
> If we want to fix this I think we'll need some kind of brand new flag that filters based on Task.kind rather than Task.label.

This could work, although it introduces yet another flag that interacts with others. There's already some non-obvious overlap between `-k`, `-J`/`-Y`, and `--tasks-regex`...

Another idea might be to use `.*{target_kind}.*` as the default value for `--tasks-regex`...but I'm not sure I like that any more than the current implementation.

Stepping back, it seems like `--tasks-regex` is maybe not fit for purpose here, and maybe there's a more general or flexibility way to do filtering that would make this use case nicer.

ahal:
> Another idea might be to use .*{target_kind}.* as the default value for --tasks-regex...but I'm not sure I like that any more than the current implementation.
>
> Stepping back, it seems like --tasks-regex is maybe not fit for purpose here, and maybe there's a more general or flexibility way to do filtering that would make this use case nicer.

Right, like the kind could be `build` and the label could be `blueberry-pie`, so yeah you got it.. I don't think there's any regex that would solve this properly.

bhearsum:
> Right, like the kind could be build and the label could be blueberry-pie, so yeah you got it.. I don't think there's any regex that would solve this properly.

Yeah, although the convention is generally that the task will have the kind name _somewhere_ in the label (this was my logic in seeking this fairly lazy improvement...I agree that it's not worth pursuing though).

PR #995, opened 2026-06-29
fix: don't run perf tests with unit tests

bhearsum:
test_graph_perf.py is the dominant cost of running tests (locally, I can run everything else in 22s single threaded; test_graph_perf.py takes almost 90s). We should avoid running it as a unit test.


bhearsum:
Note: the coverage drop is probably expected here seeing as we're running one less test...I'm not concerned.

jcristau (approved):
wizardry...

PR #1005, opened 2026-07-10
feat(run-task): Add perfherder profile markers for Git VCS operations

ahal:
While implementing this patch, I learned that Perfherder actually does
not process `PERFHERDER_DATA` anymore. However the TaskclusterETL is
still processing it (independently of Perfherder). So that's how we're
still able to get clone metrics out of robustcheckout.

Given that TaskclusterETL is slated to be removed, it's not great to be
adding a new feature that depends on it. However this will allow us to
get detailed clone metrics for Git in the short term, which will help
out with the migration.

Longer term, we'll need to figure out an alternative way to profile
`run-task` that doesn't rely on TaskclusterETL.

### DIFF
--- a/src/taskgraph/run-task/run-task
+++ b/src/taskgraph/run-task/run-task
@@ -14,6 +14,7 @@ current time to improve log usefulness.
 """
 
 import argparse
+import contextlib
 import datetime
 import errno
 import io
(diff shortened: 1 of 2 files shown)


bhearsum (approved):

jcristau (commented):
How easy will it be to distinguish hg vs git once we get this in bigquery?  Might be useful to have some marker...

ahal:
> How easy will it be to distinguish hg vs git once we get this in bigquery? Might be useful to have some marker...

There's a `serverUrl` field that is set for Hg but not set for Git.. It's not ideal, but the underlying BigQuery table that TaskclusterETL pushes to has a well defined schema. So unless we want to:

A) Update the schema in BQ
B) Update TaskclusterETL to forward our new marker (I really don't want to do this)

We can't make any changes to the existing format :/

PR #1014, opened 2026-07-29
Improve performance of Graph.transitive_closure()

ahal:


ahal:
I don't think there's a way to not regress the `fan` test. This is a tradeoff, but should be worth it as long as the real graphs have linear / diamond segments in them (which Gecko does).

Ultimately none of this is going to have any measurable impact on real world graphs.. but this does at least make the codspeed task fast(ish) again.

I think we should take it.

bhearsum:
If you could, maybe a quick before/after test with at least one gecko graph as a sanity check would be in order before this lands?

ahal:
> If you could, maybe a quick before/after test with at least one gecko graph as a sanity check would be in order before this lands?

There's too much noise to tell, especially due to that one 70s outlier. Seems like there's not a regression at least (as expected)

Before:
```
$ hyperfine "./mach taskgraph morphed" -w 1 -r 5
Benchmark 1: ./mach taskgraph morphed
  Time (mean ± σ):     51.263 s ± 11.194 s    [User: 114.538 s, System: 37.939 s]
  Range (min … max):   38.233 s … 69.037 s    5 runs
```

After:
```
$ hyperfine "./mach taskgraph morphed" -w 1 -r 5
Benchmark 1: ./mach taskgraph morphed
  Time (mean ± σ):     44.317 s ±  6.868 s    [User: 116.071 s, System: 34.583 s]
  Range (min … max):   34.272 s … 51.743 s    5 runs
```

bhearsum (approved):

PR #1024, opened 2026-08-14
fix: don't assume treeherder.group-names exists in graph config

ahal:
Some projects don't use Treeherder at all, we shouldn't assume this config exists.

Also move it outside the for loop so we only check once.

### DIFF
--- a/src/taskgraph/transforms/task.py
+++ b/src/taskgraph/transforms/task.py
@@ -870,6 +870,8 @@ def add_generic_index_routes(config, task):
 
 @transforms.add
 def process_treeherder_metadata(config, tasks):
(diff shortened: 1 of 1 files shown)


src/taskgraph/transforms/task.py
Eijebong, line 880: They were already not taking that path without a treeherder section?

ahal: There was a transform outside of the repo's control that was adding a treeherder section no matter what.

I think we should fix that transform to not do that, but I also think we shouldn't assume that transforms are well behaved.

Eijebong (approved):
I guess at least the error will be better than a key error if it adds a config containing a th group 🤷

PR #1038, opened 2026-09-10
fix(run-task): put fetched python and uv toolchains on PATH

nahadji:
run-task normalizes `MOZ_PYTHON_HOME` and `MOZ_UV_HOME` but never uses them, so tasks that fetch a python toolchain still resolve `python3` (and pip) to the system interpreter, which has no pip in minimal images:

```
+ python3 -m pip install -r ci-scripts/requirements.txt
/usr/bin/python3: No module named pip
```

(e.g. https://firefox-ci-tc.services.mozilla.com/tasks/BEzQdzYjTqGtuw7ZMfBzYA on a git-backed firefox-dev push)

This ports the post-fetch environment setup from gecko run-task: prepend the fetched python bindir to `PATH`, set `MOZPYTHONHOME` to relocate the binaries (`PYTHONHOME` would conflict with system python, e.g. the one hg uses), use certifi certificates on macOS, and add `MOZ_UV_HOME` to `PATH` when present.

A step toward #381 (stop using system python). Fixes [Bug 2066119](https://bugzilla.mozilla.org/show_bug.cgi?id=2066119) for git-backed pushes — currently blocking the `run-on-repo-type` guard removal for webrender, source-test python and wpt-unittest (Bug 1986768 migration work).

Tested: `test_scripts_run_task.py` 28/28 and `test_transforms_run_run_task.py` 9/9 pass; ruff clean on the added block.


jcristau (commented):
This is extremely gecko-specific.  We might want to reconsider using gecko's in-tree run-task there vs adding gecko shenanigans here IMO.

nahadji:
Local end-to-end check of the new path (macOS, `MOZ_PYTHON_HOME` pointed at a venv standing in for the python toolchain fetch, empty `MOZ_FETCHES`):

**Unpatched run-task** — variable is normalized but unused; the task gets system python:
```
[task] which python3: /usr/bin/python3
[task] MOZPYTHONHOME=unset
```

**With this PR:**
```
[setup] Setting up local python environment
[setup] patching ssl certificate
[setup] updated PATH with python artifact: .../pyhome/bin:/usr/bin:/bin
[task] which python3: .../pyhome/bin/python3
[task] pip 24.0 from .../pyhome/lib/python3.12/site-packages/pip (python 3.12)
[task] MOZPYTHONHOME=.../pyhome
[task] SSL_CERT_FILE=.../pyhome/lib/python3.12/site-packages/certifi/cacert.pem
```

`python3 -m pip` is the exact invocation that fails in production today (see PR description). Existing suites still pass: `test_scripts_run_task.py` 28/28, `test_transforms_run_run_task.py` 9/9.

One note for review: the macOS branch assumes the fetched python ships `certifi` (true for the gecko python toolchains; a fetch without it would raise there).

PR #1039, opened 2026-09-10
feat(run-task): add git sparse checkout support

### LINKED ISSUE #1025: Add Git partial clone support to run-task
In Gecko when cloning from Mercurial, we define a set of [sparse profiles](https://searchfox.org/firefox-main/source/build/sparse-profiles) that then get passed on to `robustcheckout.py` which then uses them with Mercurial's sparse checkout extension. There's currently no equivalent supported for Git checkouts.

We'll need to add support for partial cloning in Git checkouts if we want to avoid regressing clone performance after switching those tasks.

### Delivery Mechanism

We need a way to obtain the sparse profile in the task prior to the proper clone. There are a few options:

1. Fetch just the sparse-profile blob prior to the proper clone. This is essentially what `robustcheckout` does, but it's made efficient by the `remotefilelog` extension on HGMO. I'm unsure if we'd be able to do something similar with Github.
2. Bake it into the task definition (e.g as an env var)
3. Upload it as an artifact on the decision task and fetch it

I lean towards option 2 as it seems the simplest and fastest.

### Some other implementation notes:

1. We should only support native Git sparse profile formats (not the bespoke format in Gecko). For compatibility, we'll need a translation layer in `gecko_taskgraph` to convert the `build/sparse-profiles` format to the Git one.
2. We'll need to make sure checkout caches contain the sparse profile name in them so they don't get mixed.
3. We should only use `no-cone` mode for now. `cone` mode should be much faster, but it only works with directories which means won't be compatible with Gecko's existing profiles. In the future we should probably re-write these profiles so we can use `cone` mode, but that can be a follow-up. This can be achieved with something like:
    ```
    git clone --filter=blob:none --no-checkout <url> <dir>
    git -C <dir> sparse-checkout init --no-cone
    git -C <dir> sparse-checkout set --stdin < patterns
    git -C <dir> checkout <rev>
    ```

ahochheiden:
Fixes #1025 

run-task gains sparse checkouts for git clones. The task supplies native git patterns in `<REPO>_SPARSE_PATTERNS` as a JSON list, and names the profile with `--<repo>-sparse-profile`. A sparse clone is blobless and applies the patterns before its first checkout, so only the profile's files are written or fetched.

Checkout caches are shared between sparse and full tasks and only ever widen:
- A full task widens a sparse cache.
- A sparse task on a full cache runs as-is.
- A sparse task on a differently sparse cache appends its patterns. 

A profile made only of paths uses cone mode with the sparse index.

On the Firefox side this is supported by [Bug 2071081](https://bugzilla.mozilla.org/show_bug.cgi?id=2071081), which translates the Mercurial sparse profiles into these patterns, and [Bug 2071092](https://bugzilla.mozilla.org/show_bug.cgi?id=2071092), which keys checkout caches on the run-task a task actually runs. Delivering the patterns to git tasks lands after a release with this change is pinned.


src/taskgraph/run-task/run-task
```diff
@@ -725,20 +760,32 @@ def git_checkout(
...
+    # coming from a cache and therefore cloned by a different user.
+    args = [
+        "git",
+        "config",
+        "--global",
+        "--add",
+        "safe.directory",
+        Path(destination_path).as_posix(),
+    ]
+    retry_required_command(b"vcs", args, extra_env=env)
```
bhearsum, line 905: This is moving from being measured, to not being measured. It's arguably more correct this way (`git config` is not part of the clone), but it does change this metric. Is that on purpose?

src/taskgraph/run-task/run-task
```diff
@@ -747,7 +794,13 @@ def git_checkout(
...
 
             if shallow:
-                args.extend(["--depth=1", "--no-checkout"])
+                args.append("--depth=1")
+
+            if sparse_patterns:
+                args.append("--filter=blob:none")
+
+            if shallow or sparse_patterns:
+                args.append("--no-checkout")
```
bhearsum, line 930: Why does `sparse_patterns` imply `--no-checkout`?

bhearsum (commented):
I think I need to take a second pass on this with fresh brains; here's a few thoughts to start.

> A full task widens a sparse cache.

Does this mean that tasks that use sparse profiles that run _after_ a full checkout is cached are less performant than those that run with nothing cached, or after a different task that used a sparse profile?

> Patterns go through stdin so a name starting with a dash is not taken for an
option.

This also avoids the possibility of hitting any limits on the length of arguments, presumably?

ahochheiden:
> > A full task widens a sparse cache.
> 
> Does this mean that tasks that use sparse profiles that run _after_ a full checkout is cached are less performant than those that run with nothing cached, or after a different task that used a sparse profile?

You're correct that with a split cache, a sparse task landing on a sparse cache is faster than it is with this shared cache, but overall, on average, this is still a net win.

Workers get tasks without regard to whether they are sparse or full, so on any pool that has both kinds, most workers end up running both, and the two will always mix on their cache(s). Each approach pays for the mix somewhere. The split caches pay for it with a cold clone per kind. One cache pays for it with a sparse task after a full task having to do its fetch, checkout, and clean over the whole tree rather than just its profile.

I tested this on try for both Linux and Windows, and here's some numbers:
- sparse task on a full cache: 4s to 6s.
- sparse task on a sparse cache (same profile): 1.5s to 8s depending on the profile.
- sparse task on an empty cache (cold clone): 9s to 110s depending on the profile.

Which means that the full to sparse case costs a couple seconds at most, but every cold clone we avoid is savings of up to an order of magnitude. So unless there's a ~10:1 ratio of sparse to full jobs in some pool, it's very likely that the one cache solution is better in aggregate.

> > Patterns go through stdin so a name starting with a dash is not taken for an
> > option.
> 
> This also avoids the possibility of hitting any limits on the length of arguments, presumably?

It does!

src/taskgraph/run-task/run-task
ahochheiden: It prevents the clone from doing a checkout without the `sparse_patterns` applied on line 877. We do an explicit checkout later on line 939 with the patterns applied, and only checkout what we want. If we didn't do this, we'd checkout everything with the clone itself, and it'd never be a sparse checkout.

src/taskgraph/run-task/run-task
ahochheiden: We need to be able to read the directory to determine the config, so we can set the proper args for `PerfRecorder`.

Looks like a [2ms gap on Linux from this log](https://firefox-ci-tc.services.mozilla.com/tasks/NdLTtNMwTfSQ_233b5FEsA/runs/0/logs/public/logs/live.log#L52-53), so it's probably not worth the effort to maintain consistency here.

src/taskgraph/run-task/run-task
```diff
@@ -673,6 +678,85 @@ def _clean_git_checkout(destination_path):
...
+            for entry in out.split("\0"):
+                if not entry:
+                    continue
+                meta, path = entry.split("\t", 1)
+                if meta.split(" ")[1] == "blob" and "\n" not in path and "\r" not in path:
+                    patterns.append(f"/{_escape_pattern(path)}")
+    return patterns
+
+
+def _apply_sparse_entries(perf, op, sparse_profile, destination_path, args, entries):
```
bhearsum: Does this function have any value? AFAICT it:
* prints a message, most of which comes from the inputs
* runs a command which is almost entirely from the inputs

It seems like unnecessary indirection, unless there's some way to have more logic inserted into it (eg: choosing the correct `op`)

src/taskgraph/run-task/run-task
```diff
@@ -725,20 +811,45 @@ def git_checkout(
...
+        "git",
+        "config",
+        "--global",
+        "--add",
+        "safe.directory",
+        Path(destination_path).as_posix(),
+    ]
+    retry_required_command(b"vcs", args, extra_env=env)
+
+    is_sparse = (
```
bhearsum, line 907: These changes increase the complexity of this function a great deal, especially because of all of the new boolean decision points that are interleaved throughout. Is there a way to simplify or centralize this? (Maybe Claude would have some ideas?)

bhearsum (commented):
I don't think there's anything broken about this, but I do think it would benefit from a simplification pass, if possible. I'd really prefer not to take the increased complexity if it can be avoided; can you have a go at that before we merge?

(If you already have and this is as good as we can get, that's okay too.)

src/taskgraph/run-task/run-task
ahochheiden: I've reworked this and it's gone now. The decision logic moved into `_plan_sparse_checkout` and the command runs in `git_checkout`.

src/taskgraph/run-task/run-task
ahochheiden: This is much simpler after the rework to `_plan_sparse_checkout`.

ahochheiden:
> I don't think there's anything broken about this, but I do think it would benefit from a simplification pass, if possible. I'd really prefer not to take the increased complexity if it can be avoided; can you have a go at that before we merge?

Done!

src/taskgraph/run-task/run-task
```diff
@@ -282,11 +282,15 @@ def run_command(prefix, args, *, extra_env=None, cwd=None):
...
         stderr=subprocess.STDOUT,
-        stdin=sys.stdin.fileno(),
+        stdin=subprocess.PIPE if stdin_data is not None else sys.stdin.fileno(),
         cwd=cwd,
         env=env,
     )
 
+    if stdin_data is not None:
+        p.stdin.write(stdin_data)
+        p.stdin.close()
```
jcristau, line 292: This is prone to deadlock if stdin_data is larger than the pipe buffer size?  (Unlikely in this case, but...)

src/taskgraph/run-task/run-task
```diff
@@ -673,6 +678,52 @@ def _clean_git_checkout(destination_path):
     print_line(b"vcs", b"successfully cleaned git checkout!\n")
 
 
+def _git_config(destination_path: str, key: str) -> Optional[str]:
```
jcristau, line 681: nit: `repo_path` maybe, it's not really a destination here?

src/taskgraph/run-task/run-task
```diff
@@ -673,6 +678,52 @@ def _clean_git_checkout(destination_path):
...
+        return None
+    if p.returncode != 0:
+        raise RuntimeError(f"git config --get {key} failed: {p.stderr.strip()}")
+    return p.stdout.strip()
+
+
+class SparsePlan(NamedTuple):
+    """The one `git sparse-checkout` command a checkout needs, if any."""
+
+    op: Optional[str]
```
jcristau, line 700: nit: Literal[...] with the possible values instead of bare str?

src/taskgraph/run-task/run-task
```diff
@@ -786,16 +858,38 @@ def git_checkout(
             # shallow clone, head_rev needs to be fetched independently regardless.
             targets.append(head_rev)
 
+        # A partial clone keeps its blob filter on the origin remote, and a
+        # fetch only inherits the filter when it names that remote. Fetching
+        # the same repository by URL would download every blob of the commit.
```
jcristau, line 863: :(

src/taskgraph/run-task/run-task
```diff
@@ -700,9 +700,57 @@ def _git_lines(destination_path: str, *args: str) -> List[str]:
...
+                if meta.split(" ")[1] == "blob" and "\n" not in path and "\r" not in path:
+                    patterns.append(f"/{_escape_pattern(path)}")
+    return patterns
+
+
+def _cone_directories(patterns: List[str]) -> Optional[List[str]]:
+    directories = []
+    for pattern in patterns:
+        body = pattern[1:].rstrip("/")
+        if not pattern.startswith("/") or not body or any(c in body for c in "*?[\\"):
```
jcristau, line 744: Match body against _PATTERN_METACHARS_RE instead of doing this manually?

jcristau (commented):
Thanks for doing this Alex!

> > This also avoids the possibility of hitting any limits on the length of arguments, presumably?
> 
> It does!

They still have to go through the environment though so not really :)

ahochheiden:
> > > This also avoids the possibility of hitting any limits on the length of arguments, presumably?
> > 
> > 
> > It does!
> 
> They still have to go through the environment though so not really :)

Fair.

src/taskgraph/run-task/run-task
ahochheiden: Done.

src/taskgraph/run-task/run-task
ahochheiden: Yeah :/

src/taskgraph/run-task/run-task
ahochheiden: Done.

src/taskgraph/run-task/run-task
ahochheiden: Renamed, and the other read only helpers along with it.

src/taskgraph/run-task/run-task
ahochheiden: Switched to a temporary file for stdin, so no pipe to fill.

bhearsum (approved):
The latest update is a huge readability improvement - thanks for doing that.

jcristau (approved):

PR #1043, opened 2026-09-16
fix: add some comments around non-obvious sparse checkout branches of run-task

bhearsum:
@ahochheiden - I found this a bit confusing to read when reviewing #1039 (but not enough to block it). Can you sanity check that these comments are accurate? I think they'll be helpful to fresh eyes.


src/taskgraph/run-task/run-task
```diff
@@ -809,10 +809,17 @@ def _plan_sparse_checkout(
...
+        # and the existing sparse checkout is `cone`: simply add the
+        # new files to it.
         if cone_directories is not None and is_cone:
             args = ["git", "sparse-checkout", "add", "--stdin"]
             cone, op = True, "sparse_add"
             entries = [d for d in cone_directories if d not in set(current)]
+        # Newly checked files are not compatible with a `cone` sparse checkout
+        # and the existing sparse checkout is `cone`: perform a new sparse
+        # checkout with `cone` disabled with both the existing and newly checked
+        # out files included.
```
ahochheiden:
``` suggestion
        # Newly checked out patterns are not compatible with a `cone` sparse
        # checkout and the existing sparse checkout is `cone`: perform a new
        # sparse checkout with `cone` disabled, including the existing
        # directories, the parent directory files that `cone` mode included
        # implicitly, and the new patterns.
```
This just needs a slight correction. I covered it in one of [the commit descriptions](https://github.com/taskcluster/taskgraph/pull/1039/changes/31838d3aa71f35dc5b7d1ce6eca2b1e310afef71), but it's easy to miss so it's better here:

```
A path to a file works as a cone directory because cone mode includes the
files of every ancestor directory. The same rule means a cone checkout has a
few hundred more files than the profile names, the files directly inside every
ancestor up to the root. Never fewer, so tasks are unaffected.

Mixing follows the widen only rule. A cone cache takes further cone profiles
with `add`. A glob profile on a cone cache converts it to no cone mode and
keeps every directory it had plus the files of their ancestors, listed with
`git ls-tree` at the target revision, so nothing that was checked out
disappears.
```

So, if we didn't do the `_cone_ancestor_files` call here, switching from cone mode to a pattern list built from the old and new profiles would remove the ancestor files already there, and we wouldn't hold true to our 'widen only' claim.

src/taskgraph/run-task/run-task
```diff
@@ -809,10 +809,17 @@ def _plan_sparse_checkout(
     else:
         is_cone = _git_config(repo_path, "core.sparseCheckoutCone") == "true"
         current = _git_lines(repo_path, "sparse-checkout", "list")
+        # Newly checked out files are compatible with a `cone` sparse checkout
+        # and the existing sparse checkout is `cone`: simply add the
+        # new files to it.
```
ahochheiden:
``` suggestion
        # Newly checked out patterns are compatible with a `cone` sparse
        # checkout and the existing sparse checkout is `cone`: simply add the
        # new directories to it.
```
Nit: Strictly speaking, they're patterns and directories.

src/taskgraph/run-task/run-task
```diff
@@ -830,6 +837,9 @@ def _plan_sparse_checkout(
                     + sparse_patterns
                 )
             )
+        # Newly checked out files are not compatible with a `cone` sparse checkout
+        # and the existing sparse checkout is not `cone`: simply add the
+        # new files and patterns to it.
```
ahochheiden:
``` suggestion
        # Newly checked out patterns are not compatible with a `cone` sparse
        # checkout and the existing sparse checkout is not `cone`: simply add the
        # new patterns to it.
```
Nit: Same as above.

ahochheiden (changes requested):

jcristau (approved):

ahochheiden (approved):

bhearsum:
Good, good, github added some non-breaking space characters to those comments when I accepted the suggestions...that's helpful.

PR #1045, opened 2026-09-17
feat(run-task): implement hook point before task payload

ahal:
Sometimes consumers might want to do some additional shared setup before executing the task payload. This implements an environment variable which points to a Python script which will be executed in-process.

Bug: 2066119


ahal:
This was tested over in:
https://treeherder.mozilla.org/jobs?repo=try&revision=f837a428c489247e8c7cdded60bd84ab601c1c59&selectedTaskRun=MbIoDGe0Tru5ZWII-k07ww.0

(note the green mpu task)

The other failures in that push are mostly due to me disabling strict hashing to point at this PR. Still investigating the `ci` failure, but that doesn't need to block this PR.

src/taskgraph/run-task/run-task
```diff
@@ -319,6 +321,23 @@ def run_command(prefix, args, *, extra_env=None, cwd=None, stdin_data=None):
...
 
 
+def run_python_script(path):
+    """Execute a Python script in-process."""
+    print_line(
+        b"setup",
+        b"running python script: %s\n" % path.encode("utf-8"),
+    )
+    try:
+        runpy.run_path(path)
```
jcristau: I don't suppose we can arrange for the output from this script to be prefixed with `[setup $timestamp] `?

jcristau (approved):
Seems like a reasonable way to let firefox / other consumers customize what run-task does.

jcristau (approved):

